PR #36, opened 2015-01-03
Update UAF.md

peta909:
Added a graphicsmagick as a requirement for movie.sh to work.


volpino:
Now the script only requires ffmpeg

https://github.com/moyix/panda/blob/master/qemu/panda_plugins/replaymovie/movie.sh

PR #40, opened 2015-01-21
PANDA_LLVM_BUILD environment variable

m000:
PANDA_LLVM_BUILD variable can be used to choose which LLVM build to use without modifying build.sh.


phulin:
I think my recent commits fixing build.sh should autodetect presence of llvm.

PR #46, opened 2015-02-09
Patching PANDA to support runtime enable and disable of callbacks.

volpino:
At the moment is possible to dynamically load or unload plugins. The problem is that it's possible only to completely load/unload plugins and not to switching them off temporarily. This pull request addresses this issue.

To start getting familiar with the PANDA codebase I wanted to develop a plugin that acts as a filter for other plugins.
Example: Plugin A is the filter while Plugin B logs the CPU state. I want to dump the CPU state only when some conditions are satisfied (e.g.: PC=something or ESP=something). So Plugin B just dumps the state for every instruction but Plugin A switches it on and off based on some conditions.

To do so I patched the callback list structure to add a boolean flag `enabled`. Then when navigating the callbacks linked list the function `panda_cb_list_next` returns a pointer to the next enabled callback (skipping the disabled ones).

It is thus possible to use `panda_enable_plugin(void* plugin)` or `panda_disable_plugin(void* plugin)` to temporarily enable or disable a plugin at runtime.

As I started looking into PANDA a few days ago maybe there is a better way to achieve my goal, so feel free to comment or give me any kind of feedback.

### DIFF
--- a/docs/PANDA.md
+++ b/docs/PANDA.md
@@ -104,6 +104,10 @@ This can be used to allow one plugin to call functions another, since the handle
 
 Load a PANDA plugin. The `filename` parameter is currently interpreted as a simple filename; no searching is done (this may change in the future). This can be used to allow one plugin to load another.
 
+	void   panda_unload_plugin(void *plugin);
+
+Unload a PANDA plugin. This can be used to allow one plugin to unload another one.
+
 	void   panda_disable_plugin(void *plugin);
 
 Disables callbacks registered by a PANDA plugin. This can be used to allow one plugin to temporarily disable another one.
--- a/docs/PANDA.md
+++ b/docs/PANDA.md
@@ -104,6 +104,15 @@ This can be used to allow one plugin to call functions another, since the handle
 
 Load a PANDA plugin. The `filename` parameter is currently interpreted as a simple filename; no searching is done (this may change in the future). This can be used to allow one plugin to load another.
 
+	void   panda_disable_plugin(void *plugin);
+
+Disables callbacks registered by a PANDA plugin. This can be used to allow one plugin to temporarily disable another one.
+
+	void   panda_enable_plugin(void *plugin);
+
+Enables callbacks registered by a PANDA plugin. This can be used to re-enable callbacks of a plugin that was disabled.
+
+
 ### Argument handling
 
 PANDA allows plugins to receive options on the command line. Each option should look like `-panda-arg <plugin_name>:<key>=<value>`.
--- a/docs/ppp.md
+++ b/docs/ppp.md
@@ -24,8 +24,9 @@ they might permit A to obtain intermediate results from B.  So, the
 idea here is that A wants to control or communicate with B.
 
 Note that Way 1 is already well supported by the functions
-`panda_load_plugin` and `panda_unload_plugin`.  See `panda_plugin.h`
-for more details.
+`panda_load_plugin` / `panda_unload_plugin` and `panda_enable_plugin` / `panda_disable_plugin`
+(the former completely load or unload plugins while the latter allow to temporarily enable or
+disable callbacks registered by a given plugin).  See `panda_plugin.h` for more details.
 
 The interactions described in Ways 2 and 3, however, are tricky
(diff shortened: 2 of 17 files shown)


moyix:
Another way of accomplishing this would be to have plugin B expose enable/disable functions using [PPP](https://github.com/moyix/panda/blob/master/docs/ppp.md). However, I think this is a general enough need that it's worth adding it.

Could you do two things though before we merge, though? 

* Take out the whitespace changes (maybe your editor is doing this automatically?) so that the changes are more visible
* Add a little bit of documentation for the two new APIs to [docs/PANDA.md](https://github.com/moyix/panda/blob/master/docs/PANDA.md)

Thanks for the patch!

volpino:
I know the PPP way but that would not be generic, I would have to adapt every plugin to interact with mine. This is a more generic approach.

However I still have some doubts about my solution:

* it's everything but thread-safe
* the "filter" plugin needs to be loaded after the "target" plugin as PANDA appends callbacks at the head of the linked list. Only in this way callbacks of the filter plugin are called before the ones of the target plugin
* I tried to build the syscall plugin as explained in docs/PANDA.md and a filter plugin that in the translate_callback() checks if EAX is, let's say, zero: if the condition is true it enables the syscall plugin otherwise it disables it. In the logs of the syscall however there are also some lines that were logged with EAX different from zero. Is there a problem of concurrency? state not being consistent? If you want to do some tests I can upload the code on a gist or something

Would you have some feedback on that?

I fixed the whitespace issue (sorry, my bad!) and I will write some doc now :)

Thank you very much!

moyix:
Right now we are not too worried about thread safety: QEMU does run two threads, but it has a lock to ensure that only one thread runs at a time -- so it's basically single-threaded.

As for the trouble with the logging -- it may be due to the fact that once a block is translated it will be cached, so it may execute multiple times without calling the translate_callback again. But even so I *think* it should still be working as intended, since `helper_panda_insn_exec` will be called, which does check if the plugin is disabled. You could try issuing `panda_do_flush_tb()` after disabling or enabling a plugin to see if that helps.

If not, please do post a gist and I can take a look.

volpino:
`panda_do_flush_tb()` doesn't help unfortunately :(

Here's the gist
https://gist.github.com/volpino/2c5af80ff07978f7b9c4

the plugin `detect_instr` is just a test to try capstone and having something to detect instructions without having to hardcode raw bytes for specific instructions (you can specify the mnemonic kind of instruction you want to log, e.g.: 'sysenter'). You can also use the filter plugin with the syscall plugin on the PANDA docs

moyix:
Ok, sorry to take so long to review this! There are a couple issues with the code that you posted:

1. The value of `r0` is checked inside the `translate_callback`. This happens before any code in the block is executed, which means that its value may not be the same as it will be when the code actually runs. For example, in the `squeeze-arm-filetest` replay, one of the detect_instr.txt log entries is:

        PC=c00db820, R0=fffffdfe

    But the block containing `c00db820` disassembles to (using QEMU's `-d in_asm`):

        IN:
        0xc00db81c:  e1a00005      mov  r0, r5
        0xc00db820:  e28ddf47      add  sp, sp, #284    ; 0x11c
        0xc00db824:  e8bd85f0      pop  {r4, r5, r6, r7, r8, sl, pc}

    So as you can see the value of `r0` at runtime will actually be whatever was in `r5`, not its value at the beginning of that basic block.

2. The other weird thing is that from what I understand, it should only print when the disassembled opcode is `svc`; however, this seems to be triggering whenever it sees an `add  sp, sp` instruction. Possibly this is some endian issue?

In any case, I don't think the problem here is the enable/disable mechanism, so I will merge the pull request.

volpino:
1. Ok, yes that is what I suspected, as I wrote up there
```EDIT: Answer for this might be that the translate_callback() is called before the translation of the basic block while the exec_callback() has the actual environment context. So now I am checking that EAX is zero before the basic block but inside it it could be changed, before the syscall instruction. Can this be the case?```

2. Yeah I don't know if it actually works as expected or not, it was just a PoC to try capstone and learn the PANDA plugin API.


Ok! Happy that the pull request will be merged :) and thanks for the info!

PR #47, opened 2015-02-10
Changing panda_unload_plugin to receive a plugin as argument

volpino:
I find not really convenient the fact of having the function `panda_unload_plugin` to get an index as a parameter. It would be nicer to just get a pointer to a plugin, so we can use `panda_get_plugin_by_name` in combination with it and have a more consistent API. What do you think?

If you agree I can add documentation as well.

### DIFF
--- a/qemu/panda_plugin.c
+++ b/qemu/panda_plugin.c
@@ -163,6 +163,9 @@ void panda_unload_plugin(void* plugin) {
 }
 
 void panda_unload_plugin_idx(int plugin_idx) {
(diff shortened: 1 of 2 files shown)


volpino:
Also I just noticed now that the current implementation does not check for boundaries of idx when getting commands from the monitor

https://github.com/moyix/panda/blob/master/qemu/panda_plugin.c#L491

Added a fix for that as well

moyix:
Looks good, thank you!

PR #70, opened 2015-06-26
Fixed osi_winxpsp3x86 plugin

evandowning:
Fixed bug where free() would produce an error because of my conditional statement in on_get_current_process(). If the condition was false, then "name" in fill_osiproc() would never get allocated any memory, resulting in a free-before-malloc error.

Also made sure each of my branches was up-to-date with the master branch.


moyix:
Could you rebase this so that the pull request contains only the bug fix? It's hard to review otherwise.

evandowning:
Absolutely.

Sorry about that.

PR #84, opened 2016-01-07
Thread listing support for OSI Linux.

m000:
Added support for listing threads along with procs.
Feature must by enabled at compile time via the OSI_LINUX_LIST_THREADS flag (disabled by default).


moyix:
This looks great Manolis! Perhaps we should incorporate it into the main OSI API as well (along with the fd tracking).

m000:
For the thread listing, could we add an `uint8_t isthread` flag to `osi_proc_struct`? Would that fit with the Windows implementation of threads?

We would also need to modify/add some API call. If we don't worry a lot about backwards API compatibility, adding a flag argument to `get_processes()` would be enough. I believe adding an argument is ok, since fixing up older code will be straightforward.

Otherwise we'll have to add a new call.

PR #101, opened 2016-05-17
Minor Taint Tracking Fixes in taint2

fry:
The first two should be obvious (let me know if not). Please take a look at the alloc fix if it makes sense - I had issues where taint was lost when storing in alloc'd LLVM variables, but it has been a while ago.

Thanks for your great work on this project!


phulin:
Thanks!

PR #110, opened 2016-06-22
Fixes for introspection initialization code.

m000:
* `-os` argument is no longer checked against the compiled-in list for linux. This was preventing loading configurations from kernelinfo.conf without adding them to the compiled-in list. In general, it feels that this type of testing should better take place in the plugins rather than `panda_set_os_name()`.
* Cleanup of introspection initialization code. Eliminated use of fixed-size static buffers.
* Updated kernelinfo.conf with configurations for the latest Debian 7 (wheezy) kernel.
* Improved kernelinfo module and output parser to include more information about the kernel version.


m000:
I can also share the image I used to get the new kernel configurations. This will take a few days though, because I'll need to sanitize it before sharing.

PR #112, opened 2016-07-03
Now installing and running on Ubuntu 16.04

Svieg:
Hey guys, I was trying to make PANDA work on Ubuntu 16.04 and I had to change `CLang` and `LLVM` from `3.3` to `3.8` and `libprotobuf-c0-dev` to `libprotobuf-c-dev` and the `panda_install.bash` script now works fine on my Ubuntu 16.04 machine and everyhtings works well (creating a machine and using the plugins). Would those changes break the script on 14.04 ?


m000:
Is there any point in this?

- `libprotobuf-c-dev` breaks all other supported Linux except Ubuntu 16.04. See the [Travis CI log](https://travis-ci.org/moyix/panda/builds/141943112).
- IIRC LLVM 3.3 is a hard requirement because of specific features not present in later versions. Changing the version to 3.8 in `panda_install.bash` only allows the script to complete. [qemu/build.sh](https://github.com/moyix/panda/blob/master/qemu/build.sh) will not enable LLVM support unless v3.3 is detected. <br/> If you need PANDA with LLVM support on your Linux, you will need to follow the instructions in [docs/compile.md](https://github.com/moyix/panda/blob/master/docs/compile.md) to install v3.3 manually.

Svieg:
Okay, thanks. As you stated I think It might be bad if that breaks the script on the other distributions and if LLVM 3.3 is a hard requirement, there is no point in changing it 3.8.

m000:
It is still possible to parse the output of `lsb_release -a` to make `panda_install.bash` to work for Ubuntu 16.04 without breaking the rest of the supported platforms.

Svieg:
So that looks better. TL;DR: LLVM is now back to 3.3 and there is a check to see if it runs Ubuntu 16 and, if so, installs ```libprotobuf-c-dev``` and installs the ```c0``` version if not.

Svieg:
Should I close this ?

PR #116, opened 2016-07-15
Corner case fix for OSI Process Events

m000:
* OSI Process Events would crash PANDA if enabled at the same time with Linux OSI testing mode. This is no longer the case.
* Slightly better output format for Linux OSI testing mode.
* Removed stray .d file (unrelated with the first two fixes).


moyix:
Looks good; not sure why I didn't see this earlier. Sorry for the delay!

PR #119, opened 2016-08-03
Dump packets to pcapng instead of pcap format.

giomasce:
This allows to add additional information in frames' comments.
For instance, the guest instruction number is dumped in each frame,
which allows very precise scissoring to interesting execution
snippets.


moyix:
Nice idea! Does this add a dependency on wireshark? If so, maybe it should be added to `panda_install.bash` as well?

giomasce:
Last commit should fix that as well. Required packages should be `libwiretap-dev` and `libwsutil-dev` in Debian.

PR #120, opened 2016-08-06
Fix a double subtraction bug.

giomasce:
The removed line was already executed six lines before and led to
a malformed scissored log. This in turn caused abnormal termination
when replayed, breaking plugins whose functionalities depend on
the uninit code (i.e., tapindex).


moyix:
Good catch. Thanks!

PR #121, opened 2016-08-07
[scissors] Fix incorrect replay replication.

giomasce:
The scissors plugin used to have a bug that led to incorrect buffer
copying for some RR_CALL_CPU_MEM_RW items being in the log queue in
the moment the plugin itself began working. This produced log not
conforming to the original one and causing the replay to fail.

Although the exact error source is not known to me, this commit
introduce a small architectural change in the scissors plugin which
makes it design cleaner, removes a lot of nearly duplicate code
and solves the bug.

Specifically, the scissors plugin used to write the first log items
from the active replay queue (and here the bug occurred), then
began trivially copying the content of the log file, just shifting
the instruction counts. Now the first step has been removed. In
order to allow the scissors plugin to know where to begin copying
the old log file, the RR_log_entry struct has been extended with a
field containing the offset at which that specific entry was read
from the log file.

### DIFF
--- a/qemu/rr_log.c
+++ b/qemu/rr_log.c
@@ -675,7 +675,7 @@ static inline RR_log_entry *alloc_new_entry(void)
 static RR_log_entry *rr_read_item(void) {
     RR_log_entry *item = alloc_new_entry();
 
-    item->file_pos = ftell(rr_nondet_log->fp);
+    item->file_pos = rr_nondet_log->bytes_read;
 
     //mz read header
     rr_assert (rr_in_replay());
--- a/qemu/rr_log.c
+++ b/qemu/rr_log.c
@@ -675,6 +675,8 @@ static inline RR_log_entry *alloc_new_entry(void)
 static RR_log_entry *rr_read_item(void) {
     RR_log_entry *item = alloc_new_entry();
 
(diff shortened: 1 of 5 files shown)


qemu/rr_log.c
```diff
@@ -675,6 +675,8 @@ static inline RR_log_entry *alloc_new_entry(void)
 static RR_log_entry *rr_read_item(void) {
     RR_log_entry *item = alloc_new_entry();
 
+    item->file_pos = ftell(rr_nondet_log->fp);
```
phulin: Thanks for this PR. I'm pretty sure there is a reason ( @moyix ) that we got rid of this line - it was slow, right?

phulin: Oh. I am confused. This is your new field in the log entry struct. Ftell still is not great from a performance point of you. Brendan had a solution for this but I don't remember what it was.

moyix: I think you can replace ftell with `rr_nondet_log->bytes_read` and it will achieve the same thing without calling `ftell`. Removing `ftell` calls gave us ~10% speedup in replay, so we'd prefer to keep them out unless they're really necessary.

giomasce: You are right, I had not noticed that field. I pushed a commit that fixes this thing.

PR #124, opened 2016-11-15
Fix bug in callstack_instr that caused callstacks of size 16 to be all 0

Soly:
Also general cleanup of pandalog_callstack_create().

The `n < 16` in the loop condition on line 372 was causing the loop to be skipped whenever `n >= 16` leaving an uninitialized malloc block in the pandalog callstack entry.


moyix:
Wow, good catch. @tleek could you port this over to panda2?

PR #134, opened 2017-01-18
Changed all scripts to explicitly use python2.7 instead of default python

Komasa:
All scripts were using !#/usr/bin/env python instead of !#/usr/bin/env python2.7 so I just replaced all. On systems using python2.7 as system default (like Ubuntu and Debian) this should not change anything, on systems that default to python3 (like Arch) this ensures the correct python version is used.
Only issue that might possibly occur is in https://github.com/panda-re/panda/blob/master/tests/qemu-iotests/check#L294-L298 in case the binary run when using "python2.7" is different from the one specific in $PYTHON. Can't see a case where this would be an issue on Debian/Ubuntu, might break unusual setups though.


Komasa:
Reverted changes to non-panda scripts as per request. The setup.sh was a script I created on my own because the install instructions on the main page were lacking some necessary steps (like adding the source repositories for apt-get build-dep), will add changes to the install_ubuntu.sh in the future

PR #148, opened 2017-03-15
record absolute address

wintao:
should record the absolute address, currently the recorded address is offset in the mr, but the mr may have a none zero addr.


moyix:
Thanks! These look good. In the future, please group related changes into a single pull request though.

PR #156, opened 2017-04-08
Start recording via commandline options

vigliag:
Hi! I needed a way to launch Panda in record-mode from a python script without having to send commands via QMP. I'm currently launching Panda with the undocumented `-record-from` commandline option, and I've added an `if rr_in_record() rr_do_end_record();` statement after the main loop to be able to cleanly stop recording when closing panda with SIGTERM.

Unfortunately, I don't have enough knowledge of qemu's internals to tell if that change is correct, but it seems to be working fine on my machine. I also don't know if that `record-from` option was undocumented for some good reason. Feel free to ask for changes or close the PR.

In this pull:
- the aforementioned `rr_end_record()` after `main_loop()`
- documentation for the `-record-from` option

### DIFF
--- a/panda/docs/manual.md
+++ b/panda/docs/manual.md
@@ -382,7 +382,10 @@ end_record:
     be resumed and another recording can be made once the guest is
     resumed.
 
-Start replays from the command line using the `-replay <name>` option.
+A commandline flag `-record-from <snapshot>:<record-name>` to restores a
+qcow2 snapshot and immediately start recording is also provided for convenience.
+
+Start replays from the command line using the `-replay <name>` option. 
 
 Of course, just running a replay isn't very useful by itself, so you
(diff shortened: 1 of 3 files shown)


RickyUlrich:
Haven't looked at the pull request which I am sure is good. But just wanted to pipe in that @vigliag, you may want to take a look at https://github.com/panda-re/panda/blob/master/panda/scripts/run_guest.py which exports a function `create_recording()` that may be of use to you.  https://github.com/panda-re/panda/blob/master/panda/scripts/run_on_32bitlinux.py is an example usage of `create_recording()`

vigliag:
Thanks for the script, it may indeed be useful :) It is also way simpler and less ugly than I thought (I had to handle QMP with C++ and plain sockets for a previous project)

PR #163, opened 2017-04-26
Several Minor Fixes

mariusmue:
Hi again,

This PR contains three distinct fixes:
a) I realized that I introduced a subtle bug with my PR yesterday, as I used panda_cb_list_next() to traverse the cb_list when registering a callback. As this function skips cbs for disabled plugins, this can result into pruning cbs for disabled plugins when they are at the end of the list. This is fixed by using plist->next when traversing the list.

b) callstack_instr always uses CS_MODE_ARM for generating disassembly for ARM-Targets. I added a check whether execution is currently in THUMB-Mode or not, and dynamically change the capstone mode to CS_MODE_THUMB or CS_MODE_ARM. 

c) the board we are working with returns ARMMMUIdx_S12NSE1 as ARMMMUIdx, which was not handled inside the regime_el-function.

For the future, what is your general contributing-policy? Would you prefer to have the fix for one feature in a self-containing PR, or is it fine to send you bundled fixes as done in this PR?

Best,
Marius

### DIFF
--- a/panda/plugins/callstack_instr/callstack_instr.cpp
+++ b/panda/plugins/callstack_instr/callstack_instr.cpp
@@ -180,7 +180,17 @@ instr_type disas_block(CPUArchState* env, target_ulong pc, int size) {
 
 #if defined(TARGET_I386)
     csh handle = (env->hflags & HF_LMA_MASK) ? cs_handle_64 : cs_handle_32;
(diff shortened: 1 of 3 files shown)


moyix:
Looks good to me. I think as long as it's a small number of small changes bundling them together is fine. Thanks!

PR #178, opened 2017-05-23
makefile changes to make Panda build on gcc-6.2

vigliag:
Hi. I've started building panda with gcc-6.2 on ubuntu 16.10, patching clang and compiling llvm3.3 from sources, as from documentation. It seems to work well enough (I had been using an ubuntu 14.04 container, which was cumbersome).
This commit disables some newer compiler flags that cause problems with llvm3.3 headers or clang3.3. It should still build on ubuntu 14.04.


moyix:
Thanks!

PR #183, opened 2017-06-22
Use qemu's icount feature instead of rr_guest_instr_count.

phulin:


phulin:
Not merging due to issues with icount and guest clock.

PR #193, opened 2017-07-25
Implemented Taint2 Get LabelSet Functions

antverdovsky:
Implemented API functions for the taint2 plugin which allow other PANDA
plugins to get the taint label set of some memory address or register.
This allows plugins to see exactly what labels have been applied to
tainted data rather than only seeing the number of taint labels.

### DIFF
--- a/cpu-exec.c
+++ b/cpu-exec.c
@@ -58,6 +58,7 @@ const int has_llvm_engine = 1;
 
 int generate_llvm = 0;
 int execute_llvm = 0;
+extern bool panda_tb_chaining;
 
 /* -icount align implementation. */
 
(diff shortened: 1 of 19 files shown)


moyix:
Could you check to see if it's still possible to use the taint2 plugin from a C plugin after this? If I remember correctly the reason the labelset_iter API is designed this way is because LabelSetPs are typedefed to `std::set<uint32_t>` which means that C plugins can't include taint2.h without errors.

antverdovsky:
@moyix, Unfortunately, it doesn't work in C afterwards, though I don't believe it was possible to use it in a C plugin before this commit because of the functions:
`void taint2_labelset_spit(LabelSetP ls);`
`void taint2_labelset_iter(LabelSetP ls,  int (*app)(uint32_t el, void *stuff1), void *stuff2);`
Since they also take in a LabelSet object, which is undefined in C.

antverdovsky:
@moyix, I changed the `taint2_query_set` functions to write to a user specified array so that they may be used from C. I also went ahead and removed the incompatible `labelset_spit` and `labelset_iter` functions since there is no way for a plugin to get a `LabelSetP` anyway and so these functions can never be called.

PR #200, opened 2017-07-30
Taint2: Implemented Support For Multiple Labels

antverdovsky:
Implemented support for applying multiple labels to the same address in
the taint2 plugin. Previously when the label function was called, any
current labels applied to the address would be lost and overridden by the
new label. Now, the new label would simply be appended to the current
labels of the address, so previous taint labels are not lost.


moyix:
Under what circumstances would you want to apply multiple labels to the same address? Are you worried about labeling something that is already tainted with other labels?

m000:
Not sure about the potential application, but this changes the semantics of the existing API. So my 2¢ is that this should be made into a new taint2 API call.

antverdovsky:
@moyix, Yes this was an issue for me. I'm working on a dependency plugin for PANDA which checks if the data read from some file A is later written to another file B in some form (ergo, file B is dependent on A). When the plugin encounters a read of file A, it taints the buffer, and when it encounters a write to file B, it queries the buffer for taint. Without this change to taint2, I have an issue which occurs if I have file A write to some file B, and then I read from file B and write to some other file C. When I read from file B, I re-taint the data, so it loses the taint label applied to it when I read from file A. This results in the contents being written to file C only being tainted with the B label, so my plugin will not be able to easily detect a dependency between file A and file C. I suppose I could simply query the data to check it is not already tainted before tainting it but in the future I may want to identify a dependency between file B and file C, as well as between file A and file C as well, and I would definitely need multiple labels then.

antverdovsky:
@m000, Okay I can do that to make sure there are no conflicts for any existing utilizations of the taint2 plugin. Please see my previous comment for an application where this might be useful.

moyix:
Ok, that scenario makes sense. Once it's split into a separate API as @m000 suggested I'd be happy to merge.

antverdovsky:
@moyix, Please see commit bf92431c04e8e399a1d41e01e6a186a4e1b5ee23.

Issue #194, opened 2017-07-25
kernelinfo doesn't compile for linux_osi


bridgeythegeek:
Trying to compile `kernelinfo` as per [the instructions](https://github.com/panda-re/panda/blob/master/panda/plugins/osi_linux/USAGE.md), but get errors:

```
ake -C /lib/modules/4.4.0-81-generic/build M=/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo modules
make[1]: Entering directory '/usr/src/linux-headers-4.4.0-81-generic'
  CC [M]  /home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.o
/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.c: In function ‘init_module’:
/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.c:56:132: error: ‘struct vfsmount’ has no member named ‘mnt_parent’
 RINT_OFFSET(structp, memb, cfgname) printk(KERN_INFO "%s.%s_offset = %d", cfgname, cp_memb(#memb), (int)((void *)&(structp->memb) - (
                                                                                                                           ^
/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.c:138:2: note: in expansion of macro ‘PRINT_OFFSET’
  PRINT_OFFSET(vfsmnt_p, mnt_parent,  "fs");
  ^
/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.c:56:132: error: ‘struct vfsmount’ has no member named ‘mnt_mountpoint’
 RINT_OFFSET(structp, memb, cfgname) printk(KERN_INFO "%s.%s_offset = %d", cfgname, cp_memb(#memb), (int)((void *)&(structp->memb) - (
                                                                                                                           ^
/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.c:139:2: note: in expansion of macro ‘PRINT_OFFSET’
  PRINT_OFFSET(vfsmnt_p, mnt_mountpoint, "fs");
  ^
scripts/Makefile.build:264: recipe for target '/home/btg/build-panda/panda/panda/plugins/osi_linux/utils/kernelinfo/kernelinfo.o' failed
[...]

PR #216, opened 2017-09-06
Add support for kernel >= 3.3 to kernelinfo module.

bridgeythegeek:
As discussed in [issue#194](https://github.com/panda-re/panda/issues/194), from kernel 3.3-rc1 the `mnt_mountpoint` and `mnt_parent` members were moved from `struct vfsmount` to `struct mount`. This broke `kernelinfo`.

These commits add support for the newer kernels whilst maintaining backwards compatibility with the older kernels.

Documentation also updated accordingly.

### DIFF
--- a/panda/plugins/osi_linux/utils/kernelinfo/README.md
+++ b/panda/plugins/osi_linux/utils/kernelinfo/README.md
@@ -26,3 +26,10 @@ svn export https://github.com/panda-re/panda/trunk/panda/plugins/osi_linux/utils
 
 To compile the module, you will need to have installed the appropriate
 linux-headers package.
+
+## Kernels from v3.3-rc1 onwards
+
+As of [v3.3-rc1](https://github.com/torvalds/linux/releases/tag/v3.3-rc1), the structures `mnt_parent` and `mnt_mountpoint` were moved from `struct vfsmount` to `struct mount`. `struct mount` is defined in `/fs/mount.h` of the kernel source which is not included in the `linux-headers` package. Consequently, [`mount.h` from kernel 4.12](https://github.com/torvalds/linux/blob/6f7da290413ba713f0cdd9ff1a2a9bb129ef4f6c/fs/mount.h#L33) was added to this folder.
+
+### Here be dragons! `__randomize_layout`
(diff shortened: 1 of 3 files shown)


moyix:
Nice! Do you happen to have a replay & corresponding kernel config we can use to make sure `osi` is still working?

bridgeythegeek:
I'm sure I can rustle something up. Will need a couple of days though as am kind of maxed out right now - probably this weekend.

m000:
Good stuff. Thanks for taking the time to look into this.

bridgeythegeek:
Ok, so using this branch I did a clean `git clone` and conducted the following tests:

1. Run `make` on `kernelinfo` using the Debian image Manolis provided to me to ensure the generated conf showed the same values as are already available in `kernelinfo.conf` - it was.

1. Make a recording named `xwindows` which is of X starting up using the Debian image from Manolis and run `asidstory` against it. The output seems entirely sane. A 7z archive containing the log file, the snp file and the asidstory output can be found [here](https://drive.google.com/open?id=0B8K8oi5QFUzRdlJ5ZEo2SVlGTGc).

1. Use an install of Linux Mint 18.1 32-bit (Kernel 4.4.0) and take a recording of some arbitrary programs being launched (`less`, `vim`). `make` the new version of `kernelinfo` in this guest saving the resulting conf to `kernelinfo.conf` and check `asidstory` produces sane output - it did. A 7z archive containing the log file, the snp file, the asidstory output and the `kernelinfo.conf` used can be found [here](https://drive.google.com/open?id=0B8K8oi5QFUzRNGF1c0g4TzM1WXM).

StablelJay:
@bridgeythegeek 
The kernel of my VM is 4.10 of Ubuntu16.04, and I use make command, but some errors have happened,

moyix:
@StablelJay Yes, it looks like they've changed the data structures yet again. Going back to kernel 4.4 should work. If you feel like going adding support for newer kernels, please do create a pull request!

StablelJay:
which kernel has proved to work well in Ubuntu VM? kernel 3.16 and kernel 4.4 ?

moyix:
Yes. Also, if you just want to run Linux command line programs under PANDA, have a look at the `run_debian.py` script, which will download a VM that has a compatible profile in OSI (`linux-32-lava32`).

PR #228, opened 2017-10-31
syscalls2: Fix syscalls being tracked on a single process on windows

vigliag:
Problem addressed:
on windows the same TB can be shared by several processes. This caused the previous version of syscalls2 to miss syscalls, as it saved the couple (pc,asid) on the instruction translate callback, and checked it on the instruction exec callback.

This PR: 
This PR removes the need for the asid+pc check by repeating the "isThisInstructionASyscall" test both on instruction translation and execution callbacks


moyix:
Thanks! I think this approach will work well until we get around to revamping the translate/exec callback.

vigliag:
Perfect. Thank you!

PR #236, opened 2017-11-14
Fixed several bugs in scissors

wangray:
* Reads 1 byte for item.header.kind, item.args.kind, and item.header.callsite_loc instead of using sizeof(), which returns a wrong result for these fields (see rr_log.h)
* Doesn't touch `rr_nondet_log->bytes_read`, which it shouldn't need to modify
* Actually checks whether we've reached the end of the log using `ftell`, not `feof`
* Adds an `RR_pending_interrupts` entry at beginning of snipped log


wangray:
Merging when CI passes

panda/plugins/scissors/scissors.c
```diff
@@ -66,49 +66,74 @@ static INLINEIT void rr_fcopy(void *ptr, size_t size, size_t nmemb, FILE *oldlog
...
+        return false;
+    }
+}
+
 // Returns guest instr count (in old replay counting mode)
 static RR_prog_point copy_entry(void) {
     // Code copied from rr_log.c.
     // Copy entry.
-    RR_log_entry item;
+    RR_log_entry* item = alloc_new_entry();
```
moyix: I don't think this gets freed anywhere?

wangray: I think it's fine because it is declared `static` in alloc_new_entry? Or is the better way to not declare it static and free it at the end of this function each time.

panda/plugins/scissors/scissors.c
```diff
@@ -66,51 +66,77 @@ static INLINEIT void rr_fcopy(void *ptr, size_t size, size_t nmemb, FILE *oldlog
...
         case RR_EXIT_REQUEST:
-            RR_COPY_ITEM(item.variant.exit_request);
+            RR_COPY_ITEM(item->variant.exit_request);
             break;
         case RR_SKIPPED_CALL: {
-            RR_skipped_call_args *args = &item.variant.call_args;
+            RR_skipped_call_args *args = &item->variant.call_args;
             //mz read kind first!
-            RR_COPY_ITEM(args->kind);
+            rr_fcopy(&args->kind, 1, 1, oldlog, newlog);
```
wangray: Ughhh this optimization is soo annoying

wangray: And it will last forever due to backwards compatibility >.<

moyix: We really should have added a version number to the rr log header when we moved to PANDA2 :(

PR #243, opened 2017-12-02
Fix for SSL tutorial

Manouchehri:


Manouchehri:
Hmm, there's a lot more things broken with the tutorial as I found out, but at least this fixes one problem. 😄

moyix:
Lots of bit rot on that tutorial unfortunately :( Haven't had time to redo it since panda2...

PR #247, opened 2017-12-14
fix callstack_instr 64bit init + build fixes 

piec:
Hi,

Here's a pull request with a few build fixes + a fix on callstack_instr.

The patch in callstack_instr fixes the bug that capstone's 64bit cs_open is only done if the 32bit init fails.

```
if (cs_open(CS_ARCH_X86, CS_MODE_32, &cs_handle_32) != CS_ERR_OK)
    if (cs_open(CS_ARCH_X86, CS_MODE_64, &cs_handle_64) != CS_ERR_OK)
```

Cheers,
Pierre

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -640,7 +640,7 @@ TAGS:
 
 cscope:
 	rm -f "$(SRC_PATH)"/cscope.*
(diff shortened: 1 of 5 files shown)


moyix:
Good catch, and the other changes look good as well. Thanks!

PR #251, opened 2018-01-16
Simplified processing of -os option parsing.

m000:
Instead of splitting the option and then checking the individual components, now the options is first as a whole to be valid. Then the individual components are parsed.

Also fixed some minor documentation errors.


moyix:
Looks great! Thanks Manolis :)

PR #253, opened 2018-01-22
Fixed argument for on_unknown_sys_return.

m000:
Syscall number provided to on_unknown_sys_return - same as on_all_sys_return.


moyix:
Thanks for putting this together. Good to have this consistent now! :)

m000:
I've just pushed a much bigger delta.

My ultimate goal is to implement support for semi-automated system call
processing based on the system call signature.

To do this, I need to make signature information available at runtime. This
means that more files will be generated automatically.
With the old implementation of the system call parser, that was pretty
messy.



2018-01-23 17:53 GMT+01:00 Brendan Dolan-Gavitt <notifications@github.com>:

moyix:
Nice! Having access to signature info at runtime will allow for some really nice stuff, like auto-generating strace-like plugins... this is something I've been hoping to get around to for years (particularly with rich [SAL annotation info available on Windows system calls](https://gist.github.com/moyix/75492c74073ab450d08420489fca45b0)).

PR #254, opened 2018-01-24
syscalls2 improved prototype parsing

m000:
A fresh reimplementation of syscall_parser.py.
The new script uses Jinja2 templates for rendering. This allows for more readable code, and also properly formatted generated files.
It should now be much easier to add support for generating files for additional operating systems.


moyix:
I like where this is going :)

PR #255, opened 2018-01-29
get_syscall_info() API

m000:
Added a new API call for syscalls2. The API call provides information about the specified system call, such as its name, no of arguments, argument types and argument sizes.
This can be used in conjunction with the `on_all_sys_enter` and `on_all_sys_return` callbacks to enable processing of many different systemcalls, without implementing one callback for each.


m000:
Currently the loaded system call info are hard-coded to linux, x86. Will update the PR very soon.

moyix:
Let me know when non-x86/Linux is added and I'd be happy to merge.

m000:
This should do it. I had to rename the panda_os_* variables and the generated filenames. This swells the pushed changes, but helps to avoid excessive ugly string parsing code now and in the future.

moyix:
Great! Thanks again for putting this together!

Issue #259, opened 2018-02-12
Commandline arguments are truncated to 255 characters


vigliag:
Hi! I've ran into the 255 characters limit for the panda argument (unfortunately I hadn't read the warning as I was piping and grepping the output). This limit seems due to the way `panda_argv` is defined (as a globally allocated char array) in `panda/src/plugin.h`. Those strings should probably be heap-allocated, or the check changed so that panda exits with an error altogether.
The "argument too long" warning is being printed in `vl.c`. Checks against the hardcoded limit are currently also done in `panda_get_args` and `panda_add_arg`.
Have a nice day!

PR #260, opened 2018-02-12
Fix for #259.

m000:
panda_add_arg() now does not have any string length limitations.
panda_add_arg() signature changed so that panda_add_plugin_args() can be eliminated.


moyix:
Wonderful – I was going to try to get to this but you beat me to it! :) Thanks!

Issue #225, opened 2017-10-22
i386 replay instruction discrepancy detected


senjitsu:
Hello! I am currently experiencing an issue with replaying recordings. 

Running i386 running windows 2000, -cpu pentium3 -m 512 -vga cirrus and three isa-serial com port devices.

I get different disagreement errors for any given recording replay. The replay command I am using is identical to the normal run. The errors are usually a failed assertion somewhere in rr_log as some form of disagreement. I did originally have networking devices, and multiple HD's, but have tried to reduce to a minimally complete error reproduction.

The disagreement is usually pretty early in the replay, maybe 3-8% of the instructions. I have only recently tried debugging it a bit, but its a slow process as I need to refresh my gdb knowledge.

I know I saw something about Mozilla RR, but cant seem to find the link on it. I may be able to use that to debug this a bit more.

moyix:
Sure. RR is [Mozilla rr](http://rr-project.org/). `diverge.py` can be found in `panda/scripts/diverge.py`; it takes two `rr` recordings (one of the PANDA recording and one of the PANDA replay) and then replays them under `rr`. It searches for the first point where the state of the system diverges (i.e., memory or CPU state is different from what it should be), and then tries to backtrack to find the cause of that divergence.

senjitsu:
Interesting concept, Hopefully I can find something of interest from this.

senjitsu:
Just out of curiosity, does anybody have Mozilla RR built and all tests passing on Ubuntu 16.04? I have about 25% of tests failing, I dont want that to skew anything for looking at panda...

senjitsu:
I could really use a i386 replay file that is known good. It looks like the replay files from PandaShare are primarily using Panda 1. I want to test that my build function with a control.

moyix:
You can use run_debian to get a replay file that works (or if it doesn't, then that's a much more obvious bug!).

```
$ cd build
$ ../panda/scripts/run_debian.py ls
```

That will record "ls" running on an i386 Debian VM.

senjitsu:
Alright I will give that a shot. Here is some info from what I have gathered. 

So I am not able to get recording / replays that will even fully process every attempt, sometimes I get the infinite loop assertion, sometimes, it hangs when loading the .log file and cpu goes idle. 

But when I do get the mismatch to trigger in a replay, using mozilla RR processing with diverge.py, I have two data points. 

[...]

PR #272, opened 2018-03-26
Record VAPIC Memory Writes

nathanjackson:
We have a fix for #225. It appears that PANDA was not recording changes to the guest memory where the VAPIC was mapped.

We also have a fix for the diverge.py script that we used to discover this bug.


moyix:
Awesome, this looks great. Thanks so much for tracking this issue down! I wonder if this will also fix some issues we've heard reported with Windows XP guests...

I'll merge once CI finishes.

Issue #143, opened 2017-02-27
Net transfer callbacks missing


moyix:
The `PANDA_CB_REPLAY_NET_TRANSFER` and `PANDA_CB_REPLAY_HANDLE_PACKET` callbacks exist, but aren't currently hooked up to anything. This should be fairly easy to port over (the e1000 device doesn't seem to have changed much); someone just needs to do it.

scottcarr:
I'm going to attempt this.  Please see #275.  I think I may know how to do part of it, but guidance is welcome.

PR #275, opened 2018-04-02
[Ready for review] Net transfer callbacks missing

scottcarr:
I'm going to attempt to port the callbacks from PANDA 1.0

See also #143


scottcarr:
My current issue is this assert fails: https://github.com/scottcarr/panda/blob/master/panda/src/rr/rr_log.c#L263

According to my printf debugging size is 0.

This loop wasn't there in PANDA 1.0: https://github.com/scottcarr/panda/blob/master/hw/net/e1000.c#L1049-L1075

So, my porting in that area may be the problem.

scottcarr:
I moved the calls to `rr_record_handle_packet_call` and `rr_record_net_transfer` in `e1000_receive_iov` into the do-while loop.  This may cause one `e1000_receive` call to generate multiple entries in the recording file, but that's kind of how the code works.  It doesn't crash during recording now.  Instead, my replay crashes in `rr_replay_skipped_calls_internal` because there is no `case RR_CALL_NET_TRANSFER` or `case RR_CALL_HANDLE_PACKET`.   That's to be expected.  I'll need to implement those cases.

scottcarr:
I implemented `panda_callbacks_handle_packet` and panda_callbacks_net_transfer` and invoke them inside of `rr_replay_skipped_calls_internal`.  I think the replay/record piece may be complete.  I'll attempt to create a test plugin.

scottcarr:
I think this is working reasonably well.  My plugin prints plausible looking data.  Can someone with write access take a look please?  Maybe @moyix?  Thanks.

scottcarr:
ping @tleek @moyix Any thoughts on this PR?  Thanks.

moyix:
I think this looks good, thanks for putting it together! :)

PR #278, opened 2018-04-14
Ubuntu 18.04LTS support

m000:
Fixes for compiling on Ubuntu 18.04LTS (currently on beta2).
A new annoyance for the build process is that Ubuntu now ships with protocol buffers v3, so you need to rebuild packages for v2.
If you happen to use other software that requires protocol buffers v3 on the host, then you have a real problem.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,51 +1,96 @@
 #!/bin/bash
 
+# Prefer greadlink over readlink if present. Important for OSX (incompatible readlink).
+if type greadlink >/dev/null 2>&1; then
+    READLINK=greadlink
+else
+    READLINK=readlink
+fi
+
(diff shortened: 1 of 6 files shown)


moyix:
Hmm, we should probably look into upgrading to Protobuf v3 then. I have noticed some other issues (related to `-Wundef`) with protobuf as well, so our current version is showing its age.

m000:
Upgrading to v3 should be done eventually, but it may be too early at this point.
This would force the majority of the users who still run 16.04LTS (or even 14.04LTS) to install v3 manually. Maybe @phulin could add the debs for protobufs v3 to his ppa to smooth the transition?

btw, I can share my deb packages to anyone interested (although building them is pretty straightforward - see compile.md).

PR #279, opened 2018-04-18
Add windows 2000 support to syscalls2 plugin

MarkMankins:


moyix:
Thanks for the PR! I had no idea anyone was using PANDA to analyze Win2k :)

PR #280, opened 2018-04-19
assorted improvements

m000:
* Cleanup of stale callbacks for user mode.
* Calls to enable/disable plugins.
* Some refactoring in taint2 to make hypercalls optional to compile. Enabled by default.
* Better handling of boolean arguments.
* More compact/readable code in some files.
* Added taint2_query_set_a() - automatic space allocation.


moyix:
I think the for loop initializers are causing the build to fail right now since C99 is not the default for the version of gcc that Travis is using. Could you either move those outside the loop or change the build flags to include C99?

m000:
Switched to C99 mode for gcc/g++<5.
I've also gcc/g++ v7 on my box. This fails due to more aggressive checks (e.g. compiler determines that in some cases buffers are not sufficient for the data written). But I don't recall if the failed checks were in PANDA code or QEMU code.

m000:
@moyix I was wonder, is the implementation of `panda_disable_plugin()` sufficient? If the plugin has registered p2p callbacks with other plugins, they will still be executed, right?

moyix:
@m000 Is this ready to be merged?

m000:
Yes, it is ready. I added a final note to compile.md.

moyix:
Great, thank you!

MarkMankins:
Just noting that standalone implicit boolean arguments no longer work - perhaps because of the changes to boolean option processing.

For example, -panda file_taint:filename=hello.txt,notaint 

must now be written as -panda file_taint:filename=hello.txt,notaint=true

This probably impacts plugin documentation.

m000:
I've made some changes to bool options parsing on ea74579e6.

Personally, I prefer to have the value of the argument explicitly defined
even if we have to update a few documentation files.
The implicit default saves you exactly 2 characters worth of typing (i.e.
notaint=y).

But if there's a consensus that the the implicit default is a good idea,
adding "" to the panda_bool_true_strings array  should restore this
behaviour.

M.


Στις Τρί, 1 Μαΐ 2018 στις 3:14 μ.μ., ο/η Mark Mankins <
notifications@github.com> έγραψε:

PR #284, opened 2018-04-26
Fix race condition and add recording of VAPIC interrupt number.

nathanjackson:
I have a branch with significant changes to PANDA that I would like to offer up to merge. The changes are significant, so I'm hoping to get some feedback from the PANDA developers and to allow the CI to build to see if I missed anything build system related.

During our (NGC's) work with PANDA we've discovered that the base version of QEMU that PANDA 2 was developed from had significant problems with a race-condition mentioned both in the QEMU 2.8 changelog: https://wiki.qemu.org/ChangeLog/2.8 and in a mailing list thread: https://lists.gnu.org/archive/html/qemu-devel/2016-12/msg02643.html. We have observed this problem in Windows 2000 and Windows XP.

To fix the race condition, we bisected the QEMU code base to determine the first good QEMU commit without the race condition we had observed. Once we had the commit, we pulled those changes into PANDA. While many changes just worked, significant changes to PANDA had to happen, especially concerning the QEMU CPU loop. We also had to back out a single QEMU commit from those that were merged that allows the APIC to run concurrently with the CPU thread because of the way PANDA records I/O operations.

Secondly, we found that PANDA was not recording the interrupt number that was set on the CPU by the QEMU VAPIC. This PR also includes a fix to record APIC generated interrupt numbers. We noticed this problem after bringing in changes from QEMU to fix the race condition because the record and replay would get out of sync. Windows Server 2003 and later don't have this problem (or at the very least it is harder to reproduce on Windows Server 2003 and later) due to a change in how the NT kernel obtains the real interrupt number from the APIC TPR. Some information here: https://helpforsure.wordpress.com/tag/apic-tpr/.

If the PANDA developers are interested in bringing in these changes, but don't feel this PR is ready to merge, I can keep this branch in sync with the upstream PANDA and fix other issues as required. Either way, I'd really appreciate feedback and hopefully we can get these changes merged as I think it will solve a lot of Windows 2000 and XP (and potentially earlier NT) specific issues.


moyix:
Wow, thanks for putting this together. I think this helps resolve the mystery of why Windows XP record/replay stopped working in PANDA 2.0. I'll try to find some time to do some testing but I'm fairly busy for the next week – please poke me if I haven't followed up after that.

nathanjackson:
Will do, I will keep this branch up to date in the mean time.

nathanjackson:
Closing and reopening... looks like a problem with Travis.

nathanjackson:
I know this is a big change and I don't mean to prod, but @moyix do you have a timeline for when you can look at this?

moyix:
Hey! Thanks for the reminder. I had a chance to look at this today. I think it looks great and the changes it seems to have resolved the issues with XP.

The only change I would ask for is to remove the extraneous printf of `exiting loop in cpus.c.`. Then I'd be happy to merge it.

moyix:
Thanks! Merging :)

PR #287, opened 2018-05-01
Windows introspection rework, Enable file-taint for Windows 2000

MarkMankins:
I reworked the windows introspection plugins a bit.  Code that works on all supported versions of windows, with perhaps slight configuration variations due to kernel structure offsets, lives in the wintrospection plugin.  When necessary for OS-specific support, the wintrospection plugin will consult os-specific functions in the win7x86intro or win2000x86intro plugins.

Made a slight tweak to the file-taint plugin to account for some windows 2000 behavior in regards to system calls.  With this change, file-taint appears to be working with windows 2000 guests.


moyix:
Very cool! I like the refactor; this will help pave the way for things like Windows 10 support as well. Thanks for the contribution.

PR #288, opened 2018-05-01
Turned implicit boolean plugin arguments to explicit in documentation and scripts.

m000:
Addresses comment on #280.


moyix:
I overlooked this – it's going to break some scripts we've built on top of PANDA for LAVA but I think the changes are pretty minor, and I prefer the explicit form too. Let me double check with the other LAVA folks and make sure.

PR #289, opened 2018-05-01
File taint updates.

MarkMankins:
Previous versions of file taint cached process details prior to each basic block.
This was problematic (at least for Windows 2000) because the ASID returned by
panda_current_asid() and the process details returned by get_current_process()
are not guaranteed to be in sync prior to a basic block.  The ASID refers to
the basic block about to be executed, while the process details refer to the
basic block that was previously executed.

It looks like a workaround for this very issue was in place in the previous
version of the code.  At the beginning of each basic block, the process
details were added to the running_procs map, overwriting any previously
cached entry.

Updated the code to, by default, cache process details at the beginning of
each read system call.  I haven't encountered a case yet where caching the
details at this point results in a mismatch between the ASID and process details.

To revert to the previous behavior, one can enable the new file taint options
"cache_process_details_on_basic_block" and "always_overwrite_process_details_cache".

I'm hopeful that neither of these two options are actually necessary.  Perhaps
they can go away someday.

Another option was added to validate that the process details and ASID correspondence
remain one to one.  By enabling "process_details_cache_sanity_check", extra
assertions are added to the code to ensure that each ASID encountered maps to
the same kernel process structure.  By running a replay with
"process_details_cache_sanity_check" enabled along with "cache_process_details_on_basic_block"
enabled, it is likely that the ASID/process mismatch documented above will be
observed (resulting in a program abort.)

Note that "process_details_cache_sanity_check" option only makes sense when
"always_overwrite_process_details_cache" is not enabled.

Fixed a few memory leaks in cache_process_details() (previously osi_foo()).

Added an option to enable debug mode without recompiling the plugin.

### DIFF
--- a/panda/plugins/file_taint/file_taint.cpp
+++ b/panda/plugins/file_taint/file_taint.cpp
@@ -291,8 +291,8 @@ void read_enter(CPUState *cpu, target_ulong pc, std::string filename, uint64_t p
 
 // 3 long sys_read(unsigned int fd, char __user *buf, size_t count);
 // typedef void (*on_sys_read_return_t)(CPUState *cpu,target_ulong pc,uint32_t fd,target_ulong buf,uint32_t count);
-void read_return(CPUState *cpu, target_ulong pc, uint32_t buf, uint32_t actual_count) {
-    ThreadInfo thread{ panda_current_asid(cpu), panda_current_sp(cpu) - get_ntreadfile_esp_off() };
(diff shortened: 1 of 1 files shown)


MarkMankins:
Ug, that second commit message should be "Fix file taint..." not "File file taint..."  :(

m000:
Do a rebase to edit the message and then force push to your branch. This
should update the PR as well.

MarkMankins:
Fixed - thanks @m000.

PR #294, opened 2018-05-17
Port PANDA 1 'network' plugin to PANDA 2.

nathanjackson:
This PR brings in the PANDA 1 network plugin to PANDA 2. Tested with various protocols and operating systems.

Little changes were needed. I did have to add a call to `wtap_init(...)` and change the Makefile a little bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,9 @@ dist: trusty
 before_install:
     - sudo pip install pycparser
     - git submodule update --init dtc
-script: sudo panda/scripts/install_ubuntu.sh && cd ./build && sudo make check
+script:
+    - travis_wait 35 sleep infinity &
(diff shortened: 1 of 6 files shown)


nathanjackson:
Looks like there might be some issues with the version of the wireshark\wiretap libraries on Travis. There should be a workaround in place now.

nathanjackson:
Another build timeout. Closing\reopening to retry.

nathanjackson:
I added `travis_wait 35 sleep infinity &` which should prevent future build time outs.

moyix:
Oof, sorry Travis is being so flaky. If you have any ideas on how to make it more reliable we would be very receptive to them.

moyix:
Whew! Sorry that was such an ordeal :(

PR #297, opened 2018-05-21
use current x86 processor mode when disassemble in i386

LauraLMann:
Always using 32-bit mode in capstone causes problems when disassembling x86 16-bit binaries (running in 32-bit OS, such as FreeDOS), as some instructions are misinterpreted, causing call statements to be missed occasionally as their bytes (or some of them) are absorbed into the previous, erroneously disassembled, instruction.  Changing the mode in capstone as the processor mode in PANDA changes deals with this issue.

### DIFF
--- a/panda/plugins/callstack_instr/callstack_instr.cpp
+++ b/panda/plugins/callstack_instr/callstack_instr.cpp
@@ -11,6 +11,8 @@
  * See the COPYING file in the top-level directory.
  *
 PANDAENDCOMMENT */
+/* Change Log: */
+/* 2018-APR-13   watch for x86 processor mode changing in i386 build */
 #define __STDC_FORMAT_MACROS
 
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


moyix:
Thanks for the contribution! We currently check whether the processor is running in 64 or 32 bit mode, but we missed the case where it may be running in 16-bit mode.

Do you think you could move your changes here:

https://github.com/LauraLMann/panda/blob/eb1b5b680fdd4757ac1946cefb72ac260786d9c6/panda/plugins/callstack_instr/callstack_instr.cpp#L181-L184

So that they're all in a single location? That's also where we decide on ARM vs Thumb mode.

LauraLMann:
The disas_block function doesn't have access to the TranslationBlock (passed in to after_block_translate) that is needed to determine what the current mode is.
What is your preferred way to deal with this?  Change the signature of disas_block to pass in the new modeBit (and ignore it unless in the 32-bit i386 build), or turn modeBit into a file-level variable (like cs_handle_32), or something else?  Or is it 'programmer's choice'?

moyix:
I believe that the same information is also accessible by checking `env->hflags & HF_CS32_MASK`, which is already passed into `disas_block`.

LauraLMann:
Closing to force a rebuild in Travis.

LauraLMann:
Closing and reopening to force a rebuild - hopefully Travis is working now.

moyix:
Looks great, thank you!

PR #298, opened 2018-05-22
PPP API fix

m000:
Function pointers for PPP API calls are now properly initialized once for all compilation units. This allows splitting plugins in multiple files without having to worry about initializing PPP APIs once per compilation unit.

Implementation overview:
  - apigen.py still generates a single _ext.h header for plugins exporting PPP API calls.
  - The function pointers to PPP API calls in the _ext.h file are declared as extern instead of static.
  - The function pointers are defined in the same file, but only when PLUGIN_MAIN macro is defined.
  - panda.mak now defines PLUGIN_MAIN macro for the main source file of each plugin.


moyix:
Nice, this is a very clean way to do it. Thanks!

Issue #267, opened 2018-02-27
Cannot build with `plog_reader`


bridgeythegeek:
Using the latest docker image.

As per [the instructions in plog_reader.cpp](https://github.com/panda-re/panda/blob/master/panda/src/plog_reader.cpp), I uncommented the two lines in [Makefile.panda.target](https://github.com/panda-re/panda/blob/master/panda/Makefile.panda.target) to enable the build:

```
# Uncomment next two lines to enable compiled plog_reader.
# if PLOG_READER is defined, plog writing functions in plog-cc.cpp will be disabled.
#PROGS+=$(PLOG_READER_PROG) 
#QEMU_CFLAGS += -DPLOG_READER
```

Config with:
```
./configure --target-list=i386-softmmu --extra-plugins-path=/my_plugins
```

And then `make` gave:
```
  GEN     config-host.h
  GEN     trace/generated-tcg-tracers.h
  GEN     trace/generated-helpers-wrappers.h
  GEN     trace/generated-helpers.h
  GEN     module_block.h
  CC      block.o
  GEN     i386-softmmu/config-target.h
  GEN     trace/generated-helpers.c
  LINK    i386-softmmu/qemu-system-i386
  CXX     i386-softmmu/panda/src/plog_reader.o
  LINK    i386-softmmu/plog_reader
panda/src/plog-cc.o: In function `rr_get_guest_instr_count':
/panda/panda/include/panda/rr/rr_log.h:138: undefined reference to `cpus'
panda/src/plog-cc.o: In function `PandaLog::write_entry(std::unique_ptr<panda::LogEntry, std::default_delete<panda::LogEntry> >)':
/panda/panda/src/plog-cc.cpp:303: undefined reference to `panda_in_main_loop'
/panda/panda/src/plog-cc.cpp:304: undefined reference to `cpus'
/panda/panda/src/plog-cc.cpp:304: undefined reference to `panda_current_pc'
collect2: error: ld returned 1 exit status
make[1]: *** [plog_reader] Error 1
make: *** [subdir-i386-softmmu] Error 2
```

m000:
Do you really need to hardwire your plugin path?

I've found that setting EXTRA_PLUGINS_PATH environment variable to work
just fine.

The path pointed by the variable is expected to reflect the PANDA directory
structure. I.e. store your plugins under
$(EXTRA_PLUGINS_PATH)/panda/plugins. To enable compilation of a plugin, add
[...]

PR #303, opened 2018-06-18
Compile plog_reader during build process.

MarkMankins:
Removed PLOG_READER preprocessor symbol.  Updated plog_reader.cpp
so it can be linked with the same version of plog-cc.o that is linked
with the main PANDA binary.

Proposed fix for issue #267.


moyix:
Looks good to me. Note that we also have [panda/scripts/plog_reader.py](https://github.com/panda-re/panda/blob/master/panda/scripts/plog_reader.py) which does much the same thing.

PR #306, opened 2018-06-22
Port PANDA 1 HD transfer record\replay and callbacks.

nathanjackson:
This PR ports the PANDA 1 ability to record\replay HD transfers and run callbacks when an HD transfer occurs.


moyix:
Excellent, thank you! This should make it possible to integrate the HDD into the taint system as well...

PR #311, opened 2018-07-05
Do not mix taint on LLVM SHL 0 - just copy it

LauraLMann:
Mixing the taint on any byte involved in the SHL 0 is overkill, as if the shift is 0 bits, as there are no side effects to SHL, there is no more taint on the output than was on the input.  So, just copy the taint from the input to the output.
This change results in fewer taint reports being made, which is helpful as the console output is voluminous.


LauraLMann:
Travis couldn't install something - really doubt that's my problem.  Close and reopen to force Travis to start again.

LauraLMann:
Travis - behave yourself!

LauraLMann:
Travis couldn't install ubuntu in one of the subchecks - not my problem.  Close and reopen to force it to try again.

LauraLMann:
One more time...

LauraLMann:
Now, the last subcheck is timing out - close,...

LauraLMann:
...reopen, repeat

moyix:
Thanks and sorry as always for the pain with Travis! I'm going to see if we can disable some of the build configs so that the chance of failure goes down a bit at least...

PR #312, opened 2018-07-06
HD Taint Tracking

nathanjackson:
This PR brings the HD taint tracking functionality to Taint 2. This is somewhat of a big change so a fairly detailed explanation is below. Also, I was hoping that maybe another set of eyes could point out any problems. The testing I've done seems to indicate that the HD taint tracking works, but I'm just now becoming familiar with the taint system.

I tried to fully understand how taint tracking worked in the Taint 1 plugin and how that would fit into Taint 2. My first inclination was to use `FastShad` to store the HD shadow, but it appears that `FastShad` would not be suitable for shadowing a hard drive because even a FastShad for a small hard drive is likely to exceed a host's memory. Then I noticed that the old shadow memory data structure `SdDir` had been copied from Taint 1 into Taint 2 but it just wasn't being used, so I tried to understand the differences between `SdDir` and `FastShad`. From what I can gather, the biggest difference is that `FastShad` allocates memory up front in its constructor and `SdDir` allocates memory on demand. Therefore, `SdDir` would be the obvious choice to shadow the hard drive. Implementing the HD taint tracking using `SdDir` in Taint 2 would require some awkward mapping between `FastShad` and `SdDir`. I think this code would be ugly but I suspect also slow. So I came up with a solution that required some refactoring to the shadow memory classes and usages, but I think makes for a clean implementation of HD taint tracking.

A new abstract class called `Shad` is created that defines the functions that a shadow memory class must implement. `FastShad` now inherits from `Shad` but is otherwise the same as it was before. A new shadow memory type called `LazyShad` also inherits from `Shad` but provides an implementation that allocates memory on demand and is backed by a `std::map`. `std::map` I think is a good choice because the map elements can be accessed sequentially. The `copy` method was moved into the `Shad` class and instead of calling `memcpy` to deal with the copy, it now relies on the `query_full` and `set_full` methods. I was a little worried that this might slow things down, but in practice I found no noticeable performance regression.

There are also some stylistic changes in this PR. I don't know if PANDA has a style guide, but I tried to configure clang-format to match QEMU's coding style and the style changes are a result of this. If need be I can easily remove the formatting changes, but I had a feeling that there may not be an official style guide for PANDA code and I was hoping to improve the overall quality of the code base. If there is a style guide somewhere that I just couldn't find, I'd love to provide a clang-format settings file for it.


nathanjackson:
Try again.

nathanjackson:
Yet again.

moyix:
I had a look through and I like this refactoring – and of course it will be great to have hard drive taint! The coding style looks good too. Thanks for cleaning things up.

One question I had was related to the use of a std::map for taint lookups – in the past I believe we used a page table-like structure for this. But given the infrequency of hard drive access, probably the extra implementation complexity isn't worth the performance gain (if any?).

At the moment @LauraLMann has a couple PRs with more minor changes to the taint system pending – do you mind if I merge those first?

As always, sorry for the pain with Travis.

nathanjackson:
So I saw the `SdDir` structs using the page table style lookup. The great part about the refactor is that we could just drop in the page table style lookup later if it the `std::map` proves to be too slow and not have to change anything else. But I had the same thought as you did, the performance gain I suspect would be minimal if any. My suggestion would be to fix it later if it really becomes a problem, but if you think that it would be good to port the page table lookup I can give it a shot.

Absolutely, go ahead and merge @LauraLMann's PRs. I'll keep this branch up to date in the mean time.

moyix:
Looks like (with a bit of nudging) Travis is happy. I'm going to merge, thanks for the contribution! This has been a long time coming :)

PR #317, opened 2018-07-18
fix Shad issues and adjust copy to not double-report taint changes

LauraLMann:
Tweak include of qemu/log.h so taint2 plugin can be loaded when TAINT2_DEBUG compile switch is on.
Also check last byte in buffer to see if it is tainted in LazyShad::range_tainted.
Don't double-report changes to taint from Shad::copy (the number of taint reports went up about 50% compared to before the shadow refactoring because they were being reported once per byte in set_full, plus a one overall report from copy itself).

### DIFF
--- a/panda/plugins/taint2/shad.h
+++ b/panda/plugins/taint2/shad.h
@@ -24,7 +24,11 @@ PANDAENDCOMMENT */
 
 #ifdef TAINT2_DEBUG
 #include "qemu/osdep.h"
+
+// force qemu_log to be undecorated so it can be found
+extern "C" {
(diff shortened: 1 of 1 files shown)


LauraLMann:
Actually, I think a few more fixes need to go into this pull request.  While working another issue, I noticed that:
1) LazyShad::label is using the wrong TaintData constructor, so the new label is applied with 0 cb_mask even if the label isn't null
2) there's a misplaced } in LazyShad::remove, so it will never remove the taint as requested if you're not also tracking taint changes

moyix:
Looks good, thank you!

PR #319, opened 2018-07-20
ensure LazyShad::label creates TaintData object with cb mask set; ens…

LauraLMann:
…ure LazyShad::remove deletes labels even if taint state not being tracked
Dunno why I didn't get an E-Mail when you accepted the previous changes on this branch.  I used to get notified just fine, and I haven't changed my notification preferences.
Anyhow, hopefully this is the last of the issues that came out of the taint2 shadow refactoring.


moyix:
Whoops, sorry if I was too hasty accepting the previous PR. This looks good.

PR #323, opened 2018-07-30
improve taint precision for a few ops

Sisyph:
added dynamic checks for mul
  mul tainted 0(untainted) -> no taint propagation
  mul tainted 1(untainted) -> parallel taint propagation
lshr 0 , ashr 0 -> parallel propagation just like shl 0
sub, sdiv, udiv, fsub, fdiv (x,x) -> a repeated operand makes the output range exactly 1 value, no need to taint


moyix:
Thanks for this! These changes look great.

PR #326, opened 2018-08-07
Fix uninitialized name and size in LazyShad.

nathanjackson:
Fixes an issue in `LazyShad` where the name and size are uninitialized. Initialization of these variables is moved into the abstract class.


nathanjackson:
Close\reopen due to timeout.

Issue #322, opened 2018-07-26
Infinited loop detected when debugging tainting with llvm debug build


Sisyph:
principal failure: abort, infinite loop detected 

*changing llvm build config to release+asserts from debug+asserts removes failure*

only clue, in main loop of cpu-exec.c, num_instr_before_next_interrupt = 0, but on next iteration
cpu_handle_interrupt does nothing. Looks like interrupt number is 2

loop threshold of 20 instruction repetitions not really relevant, it goes at least 100K 

llvm commit a1e338368f5df797f3a1da81adac794340e30d11
debug with asserts build

panda master HEAD as of 7/24/18
debug build

taint2 makefile -DTAINT2_DEBUG

Recordings and logs available as 73MB zip on request
[analysisscript.txt](https://github.com/panda-re/panda/files/2231938/analysisscript.txt)
[machconf.txt](https://github.com/panda-re/panda/files/2231939/machconf.txt)

m000:
I'm also getting an infinite loop message. In my case the message appears even when no plugins are run. It looks like this:

```
$ ./build-panda/i386-softmmu/qemu-system-i386 -monitor stdio -m 512 -os linux-32-ubuntu:4.4.0-130-generic -replay ubuntu16-test

PANDA[core]: os_familyno=2 bits=32 os_details=ubuntu:4.4.0-130-generic
QEMU 2.8.50 monitor - type 'help' for more information
(qemu) loading snapshot
... done.
opening nondet log for read :	./ubuntu16-test-rr-nondet.log
rr_guest_instr_count = 0
qemu-system-i386: /home/mstamat/spammer/panda/cpu-exec.c:695: detect_infinite_loops: Assertion `false' failed.
```

Are you getting a similar message, or did we hit different bugs?

The replay I am using can be found here: https://surfdrive.surf.nl/files/index.php/s/Nk204ShtxQ3Q5FX

The replay completed fine in another branch of my private repo. I did a `git log panda/src/rr` in the working and non-working branches. I've included the output in  `rr-log-working.txt` and `rr-log-notworking.txt`.

@moyix , were the changes introduced by the delta of the commits meant to break older replays?

m000:
@moyix Indeed, ea8afa841f4228c91cbb2bdcd00bf3c36046259e is part of #284. But I'm not sure what would be an appropriate fix.

@Sisyph, if I understand correctly, for you the infinite loop manifests halfway through the replay?

[...]

PR #328, opened 2018-08-15
osi_linux fixes (wip)

m000:
First step towards addressing the infinite loop issue I have described in the [list](http://mailman.mit.edu/pipermail/panda-users/2018-August/000724.html).

Unfortunately, I can't continue with the actual fix because of #322.

However, the current batch of changes is safe to merge. It introduces some minor fixes and additional logging so that the infinite loop can be demonstrated. Additional logging is turned off by default.


m000:
Travis bump.

m000:
Travis bump.

PR #331, opened 2018-08-22
Taint prop fixes

LauraLMann:
Fix bit twiddling bugs in taint2's update_cb (mostly due to implicit casts not being accounted for).  Also added test code in tests subdirectory to test these changes.
Also fixed detaint_cb0 flag to not delete 'taint' where no labels exist in the taint object (as if there are no labels, it's not really tainted so there's nothing to delete).


moyix:
Quick question – what 's the reasoning behind pulling out the switch statement into a header file?

LauraLMann:
Sorry - I forgot to explain that.  The idea is to be able to use the exact same switch statement in both the taint2 plugin and in the test code (also submitted in this pull request).  This reduces the chances of cut-n-paste errors invalidating the testing.  It is also less complicated than adding some kind of "testing" option to the taint2 plugin, especially as trying to come up with a recording that has examples of all the bit twiddling scenarios I need to test would be quite onerous.

moyix:
Gotcha, I like it. Thanks!

PR #332, opened 2018-08-22
osi_linux fixes

m000:
Fix for the infinite loop issue during linux process enumeration. This is described in more detail in the [list](http://mailman.mit.edu/pipermail/panda-users/2018-August/000724.html).

The issue can reproduced using the trace here, after applying the temporary fix for #322:
https://surfdrive.surf.nl/files/index.php/s/Nk204ShtxQ3Q5FX

This PR changes the default order of the enumerated process. Previously enumeration started with the current task. Now enumeration starts with the init task. The old behaviour can be enabled with the `-DOSI_LINUX_LIST_FROM_CURRENT` flag.

### DIFF
--- a/cpu-exec.c
+++ b/cpu-exec.c
@@ -746,21 +746,18 @@ int cpu_exec(CPUState *cpu)
 
     /* if an exception is pending, we execute it here */
     while (!cpu_handle_exception(cpu, &ret)) {
-
-        if (panda_exit_loop) break;
-
         TranslationBlock *last_tb = NULL;
         int tb_exit = 0;
 
-        while (true) {
-
-            if (panda_exit_loop) break;
-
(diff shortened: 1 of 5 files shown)


cpu-exec.c
```diff
@@ -746,21 +746,18 @@ int cpu_exec(CPUState *cpu)
...
     while (!cpu_handle_exception(cpu, &ret)) {
-
-        if (panda_exit_loop) break;
-
         TranslationBlock *last_tb = NULL;
         int tb_exit = 0;
 
-        while (true) {
-
-            if (panda_exit_loop) break;
```
nathanjackson: @m000 Can I ask why this changed? I just ran into the issue where PANDA doesn't exit at the end, instead it deadlocks. Reverting this commit fixed that issue.

PR #334, opened 2018-08-24
Build instructions & script for Arch Linux

willish:
I have spent a bit of time making PANDA compile on Arch Linux.

This could potentially be useful for other people, so I created this PR with 3 files changed.

- `README.md` - Separate the different build instructions a little with headings
- `install_arch.sh` - An installation script to setup the dependencies and build PANDA
- `host-libusb.c` - A very small change to correct the use of a deprecated function call

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,9 @@ analysis development.
 It is currently being developed in collaboration with MIT Lincoln
 Laboratory, NYU, and Northeastern University.
 
-## Building on Debian 7/8, Ubuntu 14.04/16.04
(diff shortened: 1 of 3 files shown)


moyix:
It looks like the `libusb` change breaks with the version of `libusb` we're using in Travis CI (and I suspect under Ubuntu 16.04 which is our officially supported distribution). Could you have a look and see if there's a way to make it build under both?

willish:
Oops I forgot to include a version check

PR #336, opened 2018-08-27
add callback after the instruction executed

HighW4y2H3ll:


m000:
@HighW4y2H3ll - if everything works on your machine, keep closing and reopening the PR until Travis manages to finish all tests without timing out.

moyix:
@HighW4y2H3ll Do you think you could also implement the callback in the ARM and PPC versions of `translate.c` ?

HighW4y2H3ll:
sure, i'll look into it.

onionyst:
@HighW4y2H3ll Great job for adding the afterwards callback. I wonder, however, why there is a condition for `!dc->is_jmp` in `translate.c`. The code is as follows:

```c++
if (unlikely(panda_callbacks_after_insn_translate(ENV_GET_CPU(env), pc_ptr))
        && !dc->is_jmp) {
    gen_helper_panda_after_insn_exec(tcg_const_tl(pc_ptr));
}
```

In my understanding, the condition is marked for terminating the translation workflow. I cannot see why the `PANDA_CB_AFTER_INSN_EXEC` callback should require a `!dc->is_jmp`. With this condition enabled, all last instructions of basic blocks will not trigger the `PANDA_CB_AFTER_INSN_EXEC` callback.

HighW4y2H3ll:
Whoa that was quite a while back, if I remember correct that was to make llvm happy. At least back when I tested while in llvm 3.3, there seems to be some verifier in LLVM that checks if there’re instructions after the end of control flow. It failed miserably otherwise. I haven't tested on the new panda with LLVM 10 yet. I could roll some test on the new panda to see if it's safe to remove the `dc->is_jmp` check.

But again, I think it also sorta make sense not to instrument at the end of translation block, since you would already jump/call/ret away, and never got chance to comeback to execute the instrumented instructions. I felt a better way is to be expected to see a new translation block, and instead instrument at the beginning of the new translation block?

onionyst:
Very much appreciated for the explanation. I agree with the idea not to keep on the old block (also because `pc` has already been updated). I later realized that what I had expected can actually be worked around by constructing other callbacks. 

Thanks again.

PR #340, opened 2018-08-28
Don't exit QEMU when end of replay is requested by the user

mariusmue:
A trivial fix to not shutdown QEMU when the end of a replay is requested by the user via hmp or qmp.
Although this typically rolls back the state of QEMU to the begin of replay, it is still useful,
as other replays and or plugins can be loaded programmatically without the need to restart QEMU/PANDA.

### DIFF
--- a/panda/src/rr/rr_log.c
+++ b/panda/src/rr/rr_log.c
@@ -1508,7 +1508,7 @@ void rr_do_end_replay(int is_error)
     if (is_error) {
         panda_cleanup();
         abort();
-    } else {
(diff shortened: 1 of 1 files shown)


mariusmue:
Update: I just figured that this PR is immature - it would also keep PANDA open when the Replay finished under normal conditions. I'll write a fix for this and can either update this PR, or create a new one, up to your preferences.

moyix:
I'm going to close this for now and we can redo it when you have time!

PR #343, opened 2018-08-30
Change mask in detecting mode

tathanhdinh:
The mask used to detect mode should be `HF_CS64_MASK` (instead of `HF_LMA_MASK`), as [used](https://github.com/qemu/qemu/blob/master/target/i386/cpu.c#L5459) in qemu upstream

### DIFF
--- a/panda/plugins/callstack_instr/callstack_instr.cpp
+++ b/panda/plugins/callstack_instr/callstack_instr.cpp
@@ -180,7 +180,7 @@ instr_type disas_block(CPUArchState* env, target_ulong pc, int size) {
     instr_type res = INSTR_UNKNOWN;
 
 #if defined(TARGET_I386)
-    csh handle = (env->hflags & HF_LMA_MASK) ? cs_handle_64 : cs_handle_32;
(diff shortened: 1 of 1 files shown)


tathanhdinh:
Some builds are failed with "E: There were unauthenticated packages and -y was used without --allow-unauthenticated", I still do not understand why.

moyix:
Could you redo this by pulling in the relevant commit from upstream QEMU rather than redoing the change manually? I think this will make it easier to re-sync with upstream in the future...

tathanhdinh:
@moyix I would check out the original commit but it may not help, IMHO. Because the commits from Qemu upstream may not relevant to ones in this Panda plugin?
I've used the mask `HF_LMA_MASK` (as in Panda) but it did not give the correct mode, then I've looked into Qemu to know how it does, then discovered that it uses `HF_CS64_MASK` instead.

moyix:
Sorry, I misunderstood – yes, for the change in callstack_instr we can just use CS64. Let me kick Travis and I'll merge when it passes.

PR #348, opened 2018-09-21
Stringsearch Fixes

nathanjackson:
The computation of the string match address was incorrect in cases where the read or write size was greater than one such as in the case of an optimized `strelen(...)`. The match callback in `tstringsearch` would also not compute the address correctly because it did not know the last read size. To address these issues, I did the following:

* This PR moves all of the logic to locate a string into `stringsearch` and simplifies `tstringsearch` so that its relatively "dumb".
* A flag was added to tell match callbacks whether or not a string is actually in memory or if the string merely passed through the same program point. `tstringsearch` only taints the case where the entire string is in memory.


nathanjackson:
I removed the extra build configurations, this should reduce the chance of having problems with timeouts.

moyix:
Nice! Thank you for doing that.

m000:
A few points on the changes:
- It's not obvious why we need the `in_memory` argument in the callback. I'd surmise that this is for failed writes, but I'm not sure.
- Documentation for the plugins also needs to be brought up to date.
- There's a discrepancy between the initialization of the `pcb` struct and the registration of the callback. Although it shouldn't matter, since `pcb` is a union. https://github.com/nathanjackson/panda/blob/12bdae91d4cf911d8c0942ceb41f9f68e46e393f/panda/plugins/stringsearch/stringsearch.cpp#L228-L229

PR #350, opened 2018-09-21
Rewrite phys mem rw function to bypass DMA cbs.

nathanjackson:
Using the `cpu_physical_memory_rw_ex` in the panda_physical_memory_rw call can cause a segmentation fault when using taint and OSI plugins simultaneously. This is because of the newly introduced network taint capability which relies on DMA callbacks. Using `panda_physical_memory_rw` would trigger a DMA callback to run even though no device is actually accessing memory.

Rewriting this function to allows us to bypass the callbacks.

### DIFF
--- a/panda/src/common.c
+++ b/panda/src/common.c
@@ -218,8 +218,21 @@ void panda_set_os_name(char *os_name) {
     fprintf(stderr, PANDA_MSG_FMT "os_familyno=%d bits=%d os_details=%s\n", PANDA_CORE_NAME, panda_os_familyno, panda_os_bits, panda_os_variant);
 }
 
+extern AddressSpace address_space_memory;
+
 int panda_physical_memory_rw(hwaddr addr, uint8_t *buf, int len, int is_write) {
-    return cpu_physical_memory_rw_ex(addr, buf, len, is_write, true);
+    hwaddr l = len;
+    hwaddr addr1;
+    MemoryRegion *mr = address_space_translate(&address_space_memory, addr,
+                                               &addr1, &l, is_write);
+    void *ram_ptr = qemu_map_ram_ptr(mr->ram_block, addr1);
+
+    if (is_write) {
+        memcpy(ram_ptr, buf, len);
+    } else {
+        memcpy(buf, ram_ptr, len);
+    }
+    return MEMTX_OK;
 }
 
 


moyix:
Does this version have issues if the address passed to `panda_physical_memory_rw` is not in RAM? It seems like in that case `mr` will be null. 

More generally the reason we added the "safe" parameter to `cpu_physical_memory_rw_ex` is so that it would abort gracefully (without triggering MMIO) when passed an address outside of RAM. This allows things like dumping memory to be a lot simpler since you can just iterate page by page from 0 to `ram_size` and try to read every page without fear that this will trigger MMIO and abort the replay.

nathanjackson:
My initial thought is that it shouldn't be an issue, but maybe you know something I don't. Wouldn't you expect the address to always be an address relative to the start of RAM for `phyiscal_memory_rw`?

For `virtual_memory_rw` I could see what you're saying, but it looks like that has its own logic for handling if the virtual address is not in RAM.

nathanjackson:
Maybe a better question: does "physical" imply physical RAM or physical address space? I was under the impression physical implied physical RAM.

moyix:
I think the issue is that MMIO areas can be overlaid on top of RAM – in such cases we want to avoid accessing any MMIO region and just return -1.

To answer your second question, the "physical" in `physical_memory_rw` is meant to be an address in the physical memory space, not relative to the start of RAM (particularly since this is what the virtual to physical translation functions give you – it's not easy to figure out where the start of RAM is).

moyix:
Possibly a simple test to see if your current code handles this – take an x86 replay and run the `memsavep` plugin halfway through the replay (i.e. give it 50% as an argument). I think the default x86 memory map includes "holes" of MMIO in the same range as `[0, ram_size)`.

nathanjackson:
I see what you're saying, that makes sense. If you pass an address in the PCI hole for example, then this will fail (probably a segfault in the current implementation).

Should be easy enough to rework.

moyix:
Thanks. The other thing I just noticed is that in `panda_virtual_memory_rw` we assume that failure is `ret < 0`, but the `MemTxResult` errors are actually `ret > 0`. (This is a bug in the existing implementation too).

If you'd rather I fix this up after your pull request I'd be happy to do that too.

nathanjackson:
I don't mind. Give me just a few moments.

PR #351, opened 2018-09-21
Serial Taint

nathanjackson:
This PR introduces serial taint for both the `isa-serial` and `pci-serial` devices. This has been tested in Linux, Windows, and FreeDOS (so x86 only at this time). If other architectures are desired, it should be fairly straightforward to add, but you'll have to deal with MMIO instead of Port IO.

Note that I removed the port shadow as it shouldn't be needed. When doing Port IO on x86, the only register effected is EAX. Also this adds four new callsite IDs and event types, and this time I made sure they were all added to the very bottom of the list so as not to break older replays.

### DIFF
--- a/hw/char/serial.c
+++ b/hw/char/serial.c
@@ -481,6 +481,8 @@ static uint64_t serial_ioport_read(void *opaque, hwaddr addr, unsigned size)
                     timer_mod(s->fifo_timeout_timer, qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL) + s->char_transmit_time * 4);
                 }
                 s->timeout_ipending = 0;
+                fprintf(stderr, "serial ioport read (val = \'%c\')\n",
+                        (char)ret);
             } else {
                 ret = s->rbr;
                 s->lsr &= ~(UART_LSR_DR | UART_LSR_BI);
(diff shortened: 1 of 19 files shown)


moyix:
This looks good to me. One quick question: I noticed we're only instrumenting `helper_outb`. I guess that's because the serial port I/O only uses single-byte I/O (so no `helper_outw` or `helper_outl`)?

nathanjackson:
That is correct sir.

I was reviewing this again on Monday, this doesn't break hard drive taint as we weren't doing anything with IO ports before in taint but I removed the hard drive port IO code.

I was thinking that hard drive taint ought to be refactored so that its not in `taint2` directly. Thoughts?

moyix:
Ah, I see. Sure, I think it makes sense to pull the HDD stuff out into its own plugin then.

By the way I wanted to mention an alternative way to getting the things you want tainted that we've used in the past (but is not currently implemented). You can write a small in-guest utility that:

1. Reads in a file one block at a time
2. Makes a hypercall to tell PANDA to taint the block
3. Writes the block back to the hard drive

Then once the file is tainted and on the hard drive you just run the program you want to analyze, and the hard drive taint should automatically pick it up. You can see an example of how we use hypercalls in the LAVA code: https://github.com/panda-re/lava/blob/master/include/pirate_mark_lava.h

Anyway, all this is not actually related to the serial taint PR, which I'll merge now :)

PR #354, opened 2018-09-26
improvements in filename resolution for osi_linux

m000:
**The good:** Mounted filename resolution now *mostly* works for recent kernels. Moreover, we have partial resolution of dynamic filenames.

**The bad:** Mounts deeper than 1st level may still not work. Needs testing. Also needs testing with older kernels.

**The ugly:** This required the extraction of additional offset. Most of the configurations in `kernelinfo.conf` are now obsolete. Time for a cleanup?


moyix:
Nice. As for the kernelinfo question - maybe we can simply generate these
from the Volatility Linux profiles? They have a lot of kernels and
distributions covered and should contain all of the data structures and
offsets (and, I believe, global variables) for the kernel.

https://github.com/volatilityfoundation/profiles

m000:
Ok. Will check to see if we can reuse volatility's data.

Please hold merging until tomorrow. I want to make another pass on the code and see if I've missed any glitches.

moyix:
Cool. Just to check, this still works with the older kernels as well? (In particular we should try not to break the kernel version that the `run_debian.py` script uses)

m000:
It should be running after having the offsets regenerated.
I can see that the url for the used VM is in the script, so I'll download
it, make sure it runs, and get back to you.

M.

moyix:
Cool. If you want a pretty easy test, you should be able to just do

`../panda/scripts/run_debian.py ls`

And then replay with `asidstory` to make sure OSI is working.

m000:
Updated kernelinfo.conf to work with wheezy_panda2.qcow2, tested with a custom recording. So this should be ready to merge.

Couldn't find where run_debian.py script specifies the kernel profile to use. The script may need updating.

m000:
I removed the `lava32:32` section to keep kernelinfo.conf uncluttered—it was identical to `debian:3.2.0-4-686-pae:32`. I have added a comment to show this.

https://github.com/panda-re/panda/blob/6cda540512f7d175e2ed56a7e8925b2ca89d753f/panda/plugins/osi_linux/kernelinfo.conf#L55-L58

m000:
Yes, that's right.
If you have to changes multiple lava scripts, maybe it's easier to comment
the verbose section label and uncomment the lava section label.

M.

PR #356, opened 2018-10-01
ida_taint2

nathanjackson:
This brings over the ida_taint2 functionality from PANDA 1, but I've enhanced it to work for any OS that is supported by PANDA's OSI. The script was also changed to be run directly inside IDA and to highlight basic blocks instead of just instructions at the start of the basic block.


moyix:
Looks great. Thanks!

Issue #360, opened 2018-10-10
osi_linux: Process ID or Thread ID?


nathanjackson:
If I call `get_current_process(...)` when using `osi_linux`, do I get the PID of the current kernel task or do I get the PID of the current process? In other words, when I call `get_current_process(...)` am I actually getting the current thread?

moyix:
`get_current_process` retrieves the current `thread_info`, which lives at the bottom of the kernel stack, and then pulls out the pointer to the `task_struct`; my understanding is that each userland process gets a single `task_struct`, and the PID from that should correspond to the PID of the process.

nathanjackson:
A `task_struct` can belong to either a thread or process.

According to this file in `osi_linux` (https://raw.githubusercontent.com/panda-re/panda/master/panda/plugins/osi_linux/kernel_structs.html): get_task_struct can return either a task struct for a process or thread. What is not clear to me is whether or not these lines get the task struct for the process or thread.

```
target_ptr_t kernel_esp = panda_current_sp(env);
target_ptr_t ts = get_task_struct(env, (kernel_esp & THREADINFO_MASK));
```

My assumption was that threads while they share an address space within a process, they have their own stacks. Retrieving the current stack pointer should then return the current thread's stack?

What would be ideal is to be able to get both. I'm hoping that I can get both the process ID and the thread ID.

nathanjackson:
It looks like the task struct used in `get_current_process(...)` can be a thread.

Here's some sample code to demonstrate:

```
#include <pthread.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <unistd.h>

#include <stdio.h>

void t2(void *ptr) {
  pid_t tid = syscall(SYS_gettid);
  printf("thread id (spawn) = %d\n", tid);
}

int main()
{
  pthread_t spawn;
  if (pthread_create(&spawn, NULL, (void *)&t2, NULL)) {
    fprintf(stderr, "error creating thread\n");
    return 1;
  }
  pid_t tid = syscall(SYS_gettid);
  printf("thread id (main) = %d\n", tid);
  if (pthread_join(spawn, NULL)) {
    fprintf(stderr, "join thread failed\n");
    return 2;
  }
  return 0;
}
```

A plugin to demonstrate:

```
void gettid_return(CPUState *cpu, target_ulong pc)
{
#ifdef TARGET_I386
    CPUArchState *env = (CPUArchState *)cpu->env_ptr;
[...]

PR #361, opened 2018-10-10
OSI Thread ID Support

nathanjackson:
This PR adds support for thread id retrieval in OSI. Implementations for Windows and Linux are also provided.

Re-running my test case as mentioned in #360 reports the PID and TID as I would expect.

### DIFF
--- a/panda/plugins/osi/USAGE.md
+++ b/panda/plugins/osi/USAGE.md
@@ -53,6 +53,16 @@ typedef void (*on_get_current_process_t)(CPUState *, OsiProc **)
 
 Description: Called to get the currently running process in the guest OS. The implementation should allocate memory and fill in the pointer to an `OsiProc` struct. The returned `OsiProc` can be freed with `on_free_osiproc`.
 
+Name: **on_get_current_thread_id**
+
+Signature:
+
+```C
+typedef void (*on_get_current_thread_id_t)(CPUState *, OsiThreadId *)
+```
+
+Description: Called to retrieve the current thread id from the guest OS. OsiThreadId's memory is presumed to be managed by the caller.
+
 Name: **on_get_modules**
 
 Signature:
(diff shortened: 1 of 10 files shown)


m000:
Since we seem to be giving threads an equal treatment now, maybe we should also enable the `OSI_LINUX_LIST_THREADS` by default? This just needs to uncomment a line in the makefile.

Also, could you switch indentation to tabs for your changes in `osi_linux` files?

nathanjackson:
Tabs!? :rage1: Just kidding :laughing: Done.

Retested both Linux and Windows implementations again. Should be ready for review. @m000 @moyix

m000:
I don't have any hard preference on tabs or spaces. I just hate it when they're mixed in a single file.

Current changes look ok for now.

I believe we'll have to add one item to the todo list (e.g. add an XXX mark): Currently, there is no way to get stack information for threads. This is because `on_get_libraries()`  has an `OsiProc *p` argument, which will now be filled with the `tgid`.
One solution to this is to make `on_get_libraries()` to mark stacks/heaps for all threads in a thread group. If I understand correctly, stacks of other threads in the group will now be marked as '???'.

nathanjackson:
Couldn't we just change `on_get_libraries` to check the `tgid` instead of the task struct `pid`? Different threads should still have the same address space for a given process, so even if you aren't actually looking at the group leader the associated libraries would be the same. Right?

m000:
Exactly. Libraries don't have any issues because the address space is shared between threads. What may not work as expected is marking the stack memory region. Based on information from the current task_struct, only one area is marked as "[stack]". In a multithreaded program, the stacks of the other threads should be marked "[???]".

What I propose, is to modify on_get_libraries, so that it traverses the thread group and marks stacks for different threads as [stack:&lt;pid&gt;].
This is not a blocker for this PR, but we should probably open a new issue for it so we know what bit us when we notice that on_get_libraries doesn't work as we expect.

moyix:
Ok! Adding stack base and limit to the OsiThread data structure may be a nice future improvement...

PR #362, opened 2018-10-12
cleanning file_taint plugin

rawsample:
Clean and reorganize the code of the plugin.

* move function and struct declarations to the header file
* gather functions around 3 themes: regular, linux syscalls registered functions & windows syscalls registered functions
* replace duplicated code by function calls:
`linux_openat_enter` now calls `linux_open_enter`
`linux_read_return` now calls `linux_pread_return`

### DIFF
--- a/panda/plugins/file_taint/file_taint.cpp
+++ b/panda/plugins/file_taint/file_taint.cpp
@@ -29,65 +29,86 @@ extern "C" {
 #include "syscalls2/gen_syscalls_ext_typedefs.h"
 #include "taint2/taint2_ext.h"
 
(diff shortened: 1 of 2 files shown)


moyix:
Unfortunately I think you and @nathanjackson both did a rewrite of `file_taint` at the same time – could you have a quick look through through his work, which I've merged (since it added some new features that we wanted) and see if there are still changes that you'd like to incorporate?

Sorry for the duplication of effort – I'll try to coordinate things better in the future.

moyix:
Closing this. Sorry again for duplicated effort!

PR #364, opened 2018-10-16
Update network plugin to work with newer wireshark

nathanjackson:
Ubuntu 16.04 apparently updated wireshark. This PR should enable both the new and old versions.


moyix:
Cool. Thank you!

PR #369, opened 2018-10-30
File Taint Improvements

nathanjackson:
I've reworked a lot of the file taint code in this PR. I had lots of issues using `file_taint` in Linux and Windows. These changes fix all of the issues I could find.

The overall design of file taint is mostly the same, but there are a couple of notable changes.

1. `file_taint` doesn't need to see a file open, just that it gets read. This eliminated the need to have the notaint and enable taint at instruction count options. Now taint gets enabled when file_taint sees a read for the first time.

2. Instead of tracking reads by ASID, stack pointer, we track by process id, thread id, and file id (handle or descriptor in windows\linux). This is roughly equivalent to what was going on before, but in practice it seems much more reliable, especially in Windows.

My focus has been to make `file_taint` just work, no longer do you have to set extra command-line options. I have tested many different test binaries under Windows 2000, Windows 7, and Linux (Lubuntu 16.04).

The callbacks for `file_taint` did not look like they were in use so they aren't present but I could add them back if needed.

Note that I found wintrospection's resolution of filenames was broken. There needed to be some conversion from UTF-16 to UTF-8.


moyix:
Quick update on this: we're holding off on merging until we get the next [rode0day](http://rode0day.mit.edu) competition out, since it depends on `file_taint`. That should happen in the next few days, at which point I'll do a bit of testing and get it in. Sorry for the delay!

nathanjackson:
No problem! Thanks for the update.

nathanjackson:
@moyix I don't mean to prod, but I was wondering if there was any update on this? I've got some test cases for other work I'm doing where this would be useful to have on the master branch.

moyix:
Sorry for the delay! Got very busy the past few days. I looked through the changes and ran a few quick tests (Linux) and things look good. It's really nice that the `filename=foo` now just works by default without anything else, and the code is a lot easier to read. Merging! :)

PR #370, opened 2018-10-31
ida_taint2: Some instructions for a block may not be highlighted

nathanjackson:
This fixes an issue where the whole block won't be highlighted in IDA in some cases.


moyix:
This one is easy, merging. Thanks! :)

PR #372, opened 2018-11-01
[WIP] OSI Process Events plugin

m000:
Ongoing work on spinning off the OSI Process Events functionality as `osi_pse` plugin. The functionality was part of the `osi` plugin and enabled via a compile-time flag `-DOSI_PROC_EVENTS`. This was because it incurred some costs on every context switch in order to update its state and implement its callbacks. A separate plugin will provide more flexibility and perhaps additional PPP callbacks without overcomplicating `osi`.
Work is in progress, but a PR is sent so changes can be discussed as they evolve.

**Updated list of changes:**
- Filling `OsiProc` structs to implement `osi_pse` is pretty heavyweight (at least when fully implemented). Only a fraction of the data in `OsiProc` are needed for Process Events-like functionality. Towards this end, `OsiProcHandle` is introduced. This allows keeping track of processes without reading excessive data.
- `OsiProcHandle` related callbacks have been added and implemented for `osi_linux`.
- Dropped the `OsiProcs` structure from the codebase. A more compact and robust approach is to reuse the `GArray` container which is available through glib.
- Dropped the `OsiModules` structure from the codebase. A more compact and robust approach is to reuse the `GArray` container which is available through glib.
- Eliminated the OSI-Provider specific `on_free_*` callbacks. Generic implementations are provided instead of them in the `osi_types.h` header.
- Added `PANDA_LOG_LEVEL` macro for setting the desired log-level when using the `LOG_*` macros.


m000:
@moyix - while working on this, I've come to wonder if the `on_free_*` callbacks are really needed. In principle these seem to have been added to support custom allocation of struct members by the os-specific plugins. But this doesn't seem to be happening so far.

Maybe these callbacks can be dropped altogether to simplify the whole OSI ecosystem? We already have generic deallocators in `osi_types.h` which make the reasonable* assumption that the glib allocators are used.

*glib allocator/deallocator requires less error-checking in our code compared to stock malloc.

moyix:
Hmm, do the other OSI plugins (e.g. on Windows) use the glib allocator? I thought there was plenty of plain malloc/strdup floating around?

m000:
`osi_linux` uses only the glib allocator. Windows introspection is a mixed bag, but nothing super-complicated to switch. I can do the conversion and update the documentation.

Looking closer, we may be able to use the generic osi_free functions automatically as fallbacks when a custom free function is not provided. Something like:
```
if (PPP_CHECK_CB(on_free_osiproc)) {
    PPP_RUN_CB(on_free_osiproc, p);
}
else {
    free_osiproc_generic(p);
}
```

But if we don't foresee any osi plugins that will use some fancy allocation scheme, maybe it's better to slash the free callbacks altogether.

moyix:
Yeah, I think we can get rid of them. Right now @nathanjackson has a PR open to rewrite `file_taint` that seems to add a lot of `malloc`s and replaces some glib mallocs with plain mallocs – so perhaps you two can coordinate to get these two in sync as far as the allocator used?

nathanjackson:
I don't think I added any plain malloc to file taint. There appear to be a couple of plain frees. I did end up using `malloc` for a couple of bits in `wintrospection`, but I honestly didn't know what `g_malloc` did, so I didn't use it. `wintrospection` though is a mixed bag in general.

It sounds like I should use only `g_malloc`\`g_free`?

m000:
`g_malloc` currently is a wrapper around `malloc`, so intermixing is not a cardinal sin and we don't need to track down and eliminate every single `malloc`/`free`.

https://github.com/GNOME/glib/blob/dbf9959d15146f7b20a133a12371f075d0813576/glib/gmem.c#L92-L111

But it does help cut down boilerplate code (e.g. aborting when `malloc` failed). Glib in general is a great convenience to avoid "bad" patterns (e.g. static-sized buffer + `sprintf` vs. `g_strdup_printf`), or reimplementing common containers (e.g. `OsiProcs` vs `GArray`). It seems reasonable to promote its use in plugins, since the documentation is pretty good, we already link against it and we don't want to go full-c++.

Just my 2¢.

m000:
I can make a clean rebase on #469. When the PR is merged, I will rebase and update this one.

moyix:
I see you're still updating this; now that @nathanjackson's file_taint rewrite is in, let me know when this one is ready! Were you planning on doing separate PRs for the osi_free removal and osi_pse?

m000:
Yes, still WIP. I was hoping I could make the plugin os-agnostic, but it seems that this may not be possible, at least for linux. E.g. special handling is required for `execve` and `vfork`, because they do not trigger a PGD update. But if we add special cases for these, maybe it's better/simpler to just handle the remaining process-creation related syscalls and have one implementation per supported OS.

Shall I make a separate PR for changes to OSI? These haven't been updated for a while. I'm only working on the `osi_pse` plugin for the past couple of weeks.

Btw, while experimenting with this plugin, I realized that he current implementation of `syscalls2` does not handle properly system calls that return in two locations (e.g. `vfork`).

moyix:
Sure, let's do the changes to OSI separately so that we can get those in.

As for `syscalls2`, I think I've noticed this before but have no idea how to fix it in a reasonable way (i.e., without a bunch of OS-specific hacks). Any thoughts?

m000:
I've made a new PR containing all the changes not directly related to `osi_pse`.

At the core of the `syscalls2` issue is that syscall returns are tracked via `return_pc:asid` pair. To handle a syscall that returns in two locations, we would need to create two such pairs. But that would mean special handling of specific system calls.
We probably need a different way to identify the running system calls to handle this without implementing special cases. Not sure what that could be.

PR #373, opened 2018-11-05
Time-travel debugging support

wangray:
Test by adding `-S -s -panda checkpoint` flags to replay. Can specify RAM space for checkpoints with panda arg `:space=2G`

Attach with `gdb -x ~/panda/panda/scripts/gdbinit -ex 'set architecture {your_arch}' -ex 'target remote localhost:1234'`

* Reverse-step and reverse-continue for breakpoints on x86/x64 (TODO: watchpoints)
* Added gdbinit with `when` command to get current rr instr (more commands coming)


wangray:
yea i just pushed so i could work locally, complete docs incoming

PR #379, opened 2018-11-20
add check for zero length strings

nathanjackson:
Not sure why I didn't notice this before with my file taint testing, but this fixes an issue where the Windows unicode strings can be empty.

The Windows docs seem to indicate this would be expected: https://docs.microsoft.com/en-us/windows/desktop/api/subauth/ns-subauth-_unicode_string


moyix:
Thanks!

PR #380, opened 2018-11-21
Assorted OSI improvements

m000:
- Discontinued the PPP callbacks for freeing osi structs. Generic inlines are now provided by `osi_types.h`.
- Discontinued custom array-like structs defined by `osi` in favour of generic GArray.
- Introduced `OsiProcHandle`. This is a minimal version of `OsiProc`, which may required a fair amount of processing to be filled properly (including `pages` member). `OsiProcHandle` may be used to keep track of processes and retrieved detailed information on-demand. Currently only implementation for `osi_linux` is provided. Windows implementation should be straightforward.
- Renamed `offset` members of `osi` structs. The names were unintuitive as they were not really offsets. Opted for `taskd` and `modd` for task/module descriptor. Any alternative names are welcome.


m000:
If there's no other feedback on this, then it's ready to merge. I will then be able to rebase #372.

nathanjackson:
@m000 Did something significant change in the syscalls2 plugin-plugin API in this PR? I've pulled in the latest master and my tests for `file_taint` all segfault. Do I need to change something in `file_taint`?

moyix:
I think this is related to the extra info argument that is now available in syscalls2. I had thought this argument was optional, but someone just reported an issue in LAVA related to the same thing:

https://github.com/panda-re/lava/pull/11

@m000 is there a way to make the new syscalls2 work without the extra details (so we can avoid having to modify existing plugins and scripts)?

m000:
@moyix It's only a trivial check that is missing. That was an oversight. I had already used a new name for the added callbacks (`on_all_sys_return2`/`on_all_sys_enter2`) so that existing plugins won't break.

@nathanjackson If the crashes are consistent, it's probably the same issue. Check if #383 fixes it for you. If it doesn't, let me know.

PR #383, opened 2018-11-28
fix for panda-re/lava#11

m000:
A simple check to avoid invalid dereference when system call info have not been loaded. Fixes panda-re/lava#11.


moyix:
Looks good. Thanks!

PR #384, opened 2018-11-29
Removing reference to removed plugin dead_data

vwvw:
According to [this commit](https://github.com/panda-re/panda/commit/8d72cbecc7a93b56cd8357730b5b8618356d0da4), the dead_data plugin was removed from panda. There were still reference to it in the manual.


moyix:
Cool, thanks for the docfix!

Issue #382, opened 2018-11-27
LLVM broken for 64-bit guests


nathanjackson:
Record\replay for 64-bit guests seem to work just fine. However, when I try to use the `-llvm` switch or the taint system, I get errors.

Guests that I've tested:
- Windows Server 2008 R2 (amd64)
- Debian 9 (amd64)

The error I get changes from recording to recording. I've seen assertion errors when trying to read something from the nondet log, infinite loop assertions, segfaults, and guest instruction count disagreements.

In my debian recording, I just do an echo 'hello world' at a console.

When using `-llvm` switch, I get the following:

```
./x86_64-softmmu/qemu-system-x86_64 -net none -m 4G -replay debian-9.6.0-amd64-hello-world -llvm
loading snapshot
... done.
opening nondet log for read :   ./debian-9.6.0-amd64-hello-world-rr-nondet.log
Ahead of log while looking for log entry of type RR_SKIPPED_CALL
{guest_instr_count=1336}
        RR_INPUT_8 from RR_CALLSITE_RDTSC
FOUND DISAGREEMENT!
Replay program point:
{guest_instr_count=1339}

Record program point:
{guest_instr_count=1336}

>>> guest instruction counts disagree
debian-9.6.0-amd64-hello-world:        1339 (  0.14%) instrs.    0.50 sec.  0.28 GB ram.
ERROR: replay failed!
Time taken was: 0 seconds.
Stats:
RR_INPUT_1 number = 0, size = 0 bytes
RR_INPUT_2 number = 0, size = 0 bytes
RR_INPUT_4 number = 1, size = 14 bytes
RR_INPUT_8 number = 21, size = 378 bytes
RR_INTERRUPT_REQUEST number = 3, size = 42 bytes
RR_EXIT_REQUEST number = 0, size = 0 bytes
RR_SKIPPED_CALL number = 0, size = 0 bytes
RR_END_OF_LOG number = 0, size = 0 bytes
RR_PENDING_INTERRUPTS number = 0, size = 0 bytes
RR_EXCEPTION number = 0, size = 0 bytes
max_queue_len = 22
Checksum of guest memory: 0x1cdb357f
ERROR: replay failed!
Aborted (core dumped)
```

When using stringsearch\tstringsearch to taint the hello world string, I run into a segfault once taint is enabled.

```
[...]

PR #388, opened 2018-12-03
Fix LLVM Codegen for 64-bit Guests

nathanjackson:
This PR fixes issue #382. The load\store TCG ops were broken when loading\storing between 32-bit and 64-bit values. By adding an explicit cast, the issue has been corrected.

Tested with:

* Debian 9 (amd64)
* Windows Server 2008 R2 (amd64)

There is an issue with Taint however, with 64-bit guests there can be mul instructions that operate on 128-bit values. Since the control bits and other parts of taint2 assumed 64-bit, this isn't an easy fix. So for now I've added a warning when a mul that involves a 128-bit value that taint is not propagated. I'm not sure how much under-tainting this can introduce, but a long term solution will be needed. With some simple test binaries, I was able to see the taint reports I expected within my binary. However, I don't know what I'm missing. The alternative (ignoring the issue) results in an assertion error when compiled against a debug build of LLVM.


moyix:
This looks great. Thanks for tracking it down and fixing!

PR #389, opened 2018-12-03
Remove Reference to Binary Taint Option

nathanjackson:
This option does not appear to be implemented. Removing from `USAGE.md`.


nathanjackson:
If at first you don't succeed, try, try again.

moyix:
By the way, I noticed that when it fails it seems to get stuck on a test named something like "sleep infinitely". It may be worth just disabling that test.

Issue #378, opened 2018-11-15
taint2 generates taint change reports for untainted data


LauraLMann:
The taint2 plugin always calculates new bit masks (controlled bits, zeros, ones) when a taint operation comes in, even if the byte in question isn't really tainted (has no labels).  As some LLVM instructions put a constant into some part of one or more masks, this can cause a taint change report to be generated.  (For example, LLVM SHL will put some ones in the lower bits of the zero mask.)
This erroneous taint change report messes up the ability of tainted_instr to quit early (using the num argument), as it considers consecutive changes for the same PC to be one instruction, only increments the counter when the byte has taint labels, but updates the last-seen-PC regardless if any labels were found.  So, if the first taint 'change' is for unlabeled data, the last-seen-PC is updated, and then if the next taint change is for taint-with-labels on the same PC, the counter will not be incremented.  Thus, tainted_instr won't quit at the limit requested, but some time later.
Although it would be easy to work around this issue in tainted_instr, wouldn't it be better to fix taint2 to not calculate bit masks on untainted data?

moyix:
Yes, I think that makes sense. Do you want to make a PR for it?

LauraLMann:
Yes, I would like to make a PR for this problem.  But, until #377 is resolved (or time-travel debug support is backed out of master) it won't be very useful.  It won't even be testable, which is why I hesitate.

moyix:
That shouldn't block testing (there's a workaround listed in the comments of that issue; we're just not applying it until we're sure we understand what's going on). I'll have another look at it today.

moyix:
#377 is fixed now! :)

LauraLMann:
The problem with update_cb updating masks on untainted data is only part of the problem causing tainted_instr to miscount the tainted 'instructions' it has seen.  (It's really counting tainted blocks, but that's another matter.)  Tainted_instr doesn't account for the fact that when taint_delete is called on a tainted item that a taint change report will also be generated.  So, tainted_instr shouldn't note the PC and ASID for the most recently seen tainted 'instruction' unless the current change report is actually for a tainted item.  It's easy enough to fix this too, so I will do so.  (I will fix it in tainted_instr of course, in case some other plugin wants to know about taint that is deleted.)

wangray:
Fixed by #390 , closing unless indicated otherwise.

Issue #226, opened 2017-10-28
taint2 segfaults on taint_host_memcpy when debugging is on


vigliag:
Hi! I've encountered this segfault while trying to debug a problem I was having with taint2 (I'll talk about that issue in details in a separate issue/mail, basically some data which should have been untainted had several labels attached). Can you please have a look at it?

I was running on the i386 version of panda (`qemu-system-i386`), with a windows7SP1 x86 guest, with these arguments to taint2: `taint2:no_tp=true,debug=true`

Also, I don't know if it this is relevant, or if it may affect the symbols the backtrace refers to, but I've rebuilt  the underlying panda executables a couple of time while this analysis was running (it took quite a bit of time). I have however run this test twice, obtaining the same error and backtrace.

This is the backtrace I got from the core file:
```
#0  FastShad::query (addr=<optimized out>, this=0x7fbe94a3f6e8) at /home/vigliag/thesis/panda/panda/plugins/taint2/fast_shad.h:172
#1  taint_host_memcpy (env_ptr=<optimized out>, dest=<optimized out>, src=94199508767888, greg=<optimized out>, gspec=<optimized out>, size=16, labels_per_reg=4)
    at /home/vigliag/thesis/panda/panda/plugins/taint2/taint_ops.cpp:397
#2  0x00007fbe9accb9f6 in ?? ()
#3  0x0000000000000004 in ?? ()
#4  0x0000000000000004 in ?? ()
#5  0x0000003000000000 in ?? ()
#6  0x00007fbedd9ca650 in ?? () at /home/vigliag/thesis/panda/panda/plugins/taint2/fast_shad.h:119 from /home/vigliag/panda/panda/build/i386-softmmu/panda/plugins/panda_taint2.so
#7  0x00007fbedd9cb160 in ?? () at /home/vigliag/thesis/panda/panda/plugins/taint2/taint_ops.cpp:642 from /home/vigliag/panda/panda/build/i386-softmmu/panda/plugins/panda_taint2.so
#8  0x00007fbedd9cb160 in ?? () at /home/vigliag/thesis/panda/panda/plugins/taint2/taint_ops.cpp:642 from /home/vigliag/panda/panda/build/i386-softmmu/panda/plugins/panda_taint2.so
[...]
```

This is what I had in dmesg. Not sure why it is pointing to those .so files though.
```
[...]

PR #390, opened 2018-12-04
Only update tainted data in taint2

LauraLMann:
Do not calculate taint masks on data that is not tainted in update_cb.  Also modify tainted_instr plugin so that a taint change due to a delete does not update the PC and ASID of the most recently seen tainted instruction.  (See issue #378.)
Also, send offset of address to taint log 'function' in taint2's taint_host_memcpy so don't seg fault when producing taint2 debug output.  (See issue #226.  Yeah, I know it's a totally separate issue, but it was getting in the way of debugging other problems.)


moyix:
@wangray could you take a quick look? I think you've been looking at the update_cb code recently.

wangray:
These changes make sense to me. I don't have direct experience with the issues #378 and #226, but based on the descriptions these are the correct fixes.

PR #393, opened 2018-12-12
Allow recording of register writes via GDB

mariusmue:
This PR introduces a new SKIPPED_CALL variant `RR_CPU_REG_WRITE` and various code additions to record and handle it. The idea is, that when the analyst changes register values during a record, that those are reflected in the log as well. As a result, interactive "fiddling" from outside the guest can be recorded and replayed as well.

Memory writes from a gdb-session are already handled by `rr_device_mem_rw_call_record`.
However, while digging into this, I discovered that this function would set the SKIPPED_CALL kind always to RR_CPU_MEM_UNMAP. While this has the same signature as RR_CPU_MEM_RW, it's handled different during replay; I included a patch which cleans this up and provides a new `rr_device_mem_unmap_call_record` function, which is called during `address_space_unmap` in `exec.c`. All other calls to `rr_device_mem_rw_call_record` seemed to actually just record a memory write.

Please note that I couldn't test this thoroughly yet and are not sure, whether it breaks any other parts of the code. However, small tests with a uboot-binary on the versatilepb board didn't result in any failure and provided the desired functionality.

### DIFF
--- a/exec.c
+++ b/exec.c
@@ -3322,7 +3322,7 @@ void address_space_unmap(AddressSpace *as, void *buffer, hwaddr len,
         if (is_write) {
             //bdg Save addr1,access_len,buffer contents
             if (rr_in_record()) {
-                rr_device_mem_rw_call_record(addr1, buffer, access_len, is_write);
+                rr_device_mem_unmap_call_record(addr1, buffer, access_len, is_write);
             }
             invalidate_and_set_dirty(mr, addr1, access_len);
         }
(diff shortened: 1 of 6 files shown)


moyix:
This appears to work fine with 32 and 64 bit x86 replays I've tried (or rather it doesn't break the existing record/replay functionality – I didn't test register writes). And I like this because it takes us a bit closer to being able to represent arbitrary changes to memory and CPU states during a replay.

Thanks!

Issue #392, opened 2018-12-11
handling of clone from syscalls2


m000:
It seems that clone (and possibly other?) system calls are not handled properly by `syscalls2`. Specifically, no return callback is invoked. This is what I see when analyzing my trace:

1. We have the `agetty` process.
```
taskd    asid     pid/ppid name
d7766900 1793c8a0 1872:1   agetty
```

2. The process runs several syscalls. There are one `on_all_sys_enter2` and one `on_all_sys_return2` hooked. Additionally, there is one `on_sys_clone_return` return callback hooked. Only `on_all_sys_enter2` is invoked for clone.

```
PANDA[osi_pse]:D> SYS0  sys_ioctl               d7766900 1793c8a0
PANDA[osi_pse]:D> SYSR  sys_ioctl               d7766900 1793c8a0
PANDA[osi_pse]:D> SYS0  sys_rt_sigprocmask      d7766900 1793c8a0
PANDA[osi_pse]:D> SYSR  sys_rt_sigprocmask      d7766900 1793c8a0
PANDA[osi_pse]:D> SYS0  rt_sigaction            d7766900 1793c8a0
PANDA[osi_pse]:D> SYSR  rt_sigaction            d7766900 1793c8a0
PANDA[osi_pse]:D> SYS0  sys_ioctl               d7766900 1793c8a0
PANDA[osi_pse]:D> SYSR  sys_ioctl               d7766900 1793c8a0
PANDA[osi_pse]:D> SYS0  sys_clone               d7766900 1793c8a0
PANDA[osi_pse]:D>
```

3. A couple of context switched later, a new `agetty` process appears. The new process has the first one as parent. Which implies that clone succeeded but somehow the callbacks were missed.
```
taskd    asid      pid/ppid    name
d7762300 18231d80  1878:1872   agetty
```

Note that the callbacks are invoked normally in other cases. So this is probably some corner case that is missing (?). Not sure where to start looking, so any hints are appreciated. I can also upload the trace for anyone interested.

m000:
A demonstration for this can be made with #395. I ran it for a short trace and at the end of the analysis, `running_syscalls` contained the following entries: 
```
      1 sys_clone
      1 sys_futex
      1 sys_poll
      1 sys_rt_sigreturn
      1 sys_rt_sigtimedwait
      1 sys_select
      2 sys_nanosleep
      2 sys_write
      5 sys_epoll_wait
    208 sys_sigreturn
    596 sys_execve
[...]

PR #395, opened 2018-12-14
syscalls2 debug printout

m000:
Dumps debug information every time a syscall return is detected. Related to #392.


moyix:
Is this something you were hoping to get merged into master? If so is it disabled by default?

m000:
It is enabled by default, as currently it is meant to demonstrate help solve the issues described in #392.
Let's keep the PR open for now. I will try to at least add some white-listing to prevent accumulation of execve syscalls. Eventually, it could be merged into master after having it disabled by default.

m000:
This can now be merged. #392 is largely addressed.

Still, some system call returns may not be handled properly. But it's better to start a fresh issue on any remaining issues to keep the discussion focused.

Issue #396, opened 2018-12-17
The recording of login/logout failed to replay, on the latest version of Win7 64-bit.


Dazdin9o:
I found a case that PANDA cannot replay the process of login/logout on the lated version of Win7 64-bit.

It may caused by rr_memory_region_find_parent(), in panda/src/rr/rr_log.c:1013, returns NULL. Stack backtrace:

```
(gdb) bt
#0  0x0000555555a5052f in memory_region_del_subregion (mr=0x0, subregion=0x555557788170) at /home/jungu/timeplayer/timeplayer-src/memory.c:2091
#1  0x00005555559c47cb in rr_replay_skipped_calls_internal (call_site=<optimized out>) at /home/jungu/timeplayer/timeplayer-src/panda/src/rr/rr_log.c:1017
#2  0x00005555559fc4d4 in rr_replay_skipped_calls () at /home/jungu/timeplayer/timeplayer-src/panda/include/panda/rr/rr_log_all.h:331
#3  address_space_stw_internal (endian=DEVICE_NATIVE_ENDIAN, result=0x0, attrs=..., val=256, addr=<optimized out>, as=<optimized out>) at /home/jungu/timeplayer/timeplayer-src/memory_ldst.inc.c:585
#4  address_space_stw (as=<optimized out>, addr=<optimized out>, val=256, attrs=..., result=result@entry=0x0) at /home/jungu/timeplayer/timeplayer-src/memory_ldst.inc.c:620
#5  0x0000555555b1d521 in helper_outw (env=<optimized out>, port=<optimized out>, data=<optimized out>) at /home/jungu/timeplayer/timeplayer-src/target/i386/misc_helper.c:61
#6  0x00007fffd02ac6c9 in ?? ()
#7  0xfffff96000296000 in ?? ()
#8  0x000055555771b070 in ?? ()
#9  0x0000000000000005 in ?? ()
#10 0x0000555557723310 in ?? ()
#11 0x0000555558fe7620 in ?? ()
#12 0x00007ffff313b559 in g_hash_table_lookup () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x0000555555a06c77 in tcg_gen_callN (s=0x5555571f1060 <tcg_ctx>, func=0x555555b19040 <helper_cc_compute_c>, ret=4, nargs=4, args=0x7fffbade95d0) at /home/jungu/timeplayer/timeplayer-src/tcg/tcg.c:739
[...]

Issue #330, opened 2018-08-21
Segfault when deleting a memory subregion


toomanybananas:
`rr_memory_region_find_parent` can return NULL, on line 1013 of [rr_log.c](https://github.com/panda-re/panda/blob/master/panda/src/rr/rr_log.c) that is called, and then passed to memory_region_del_subregion which dereferences the parent, causing a segfault. No NULL check is performed.

I have no idea what it is about my replay that causes it to segfault in this way but there's a bug here.

I can't share the trace but I'm running `panda/build/x86_64-softmmu/qemu-system-x86_64 -fda some_dos_image.img -m 128 -replay test1`, operating system being run is DOS.

moyix:
Any chance you could find a public DOS image that shows this problem?

toomanybananas:
Sorry for the super late reply, but here's the image in question (which is now public):
Password is 'infected' (not malware, CTF challenge).

[suspicious_floppy_v1.0.zip](https://github.com/panda-re/panda/files/2542877/suspicious_floppy_v1.0.zip)

PR #397, opened 2018-12-18
Workaround of issue #396 and #330

Dazdin9o:
I just check the return value of `rr_memory_region_find_parent `, and I'm nor sure if this way solved the root cause of this issue, but it is workable :)


moyix:
Nice – does this allow the replay to complete successfully?

moyix:
In any case, I suppose this fix can't *hurt* anything, so I'll merge it. Thanks!

PR #399, opened 2019-01-21
WIP - PANDA-QEMU 2.9.1

nathanjackson:
This PR brings PANDA up to date with QEMU 2.9.1. The long-term goal here is to make PANDA as stable as possible and potential future upgrades of the QEMU version easier.

There are two notable differences between QEMU 2.9.1 and PANDA at this point:

* Changes related to Multi-threaded TCG (MTTCG) were omitted in favor of the PANDA version (by using a combination of `git merge -s ours` and `git checkout --ours`) since MTTCG is incompatible with PANDA's replay model.
* During development of QEMU 2.9.1, some APIC changes were introduced. These changes were also omitted as they are incompatible with old replays. However, it appears that the only reason they are incompatible is due to a change in the device's QEMU class structure. I think this could be handled if we had versioned non-deterministic logs or a schemaless format.

Last time we had changes like this we had some issues that broke old replays, but care has been taken to ensure this doesn't happen again.

Our regression tests run various plugins against a set of recordings that we have created and check the output of the plugins. These tests cover the following operating systems on x86:

* Ubuntu 16.04 (Lubuntu flavor)
* Windows 2000, 2003, 7
* FreeDOS

Note that our tests don't cover every combination of plugin\OS, especially for FreeDOS. Even so, the regression tests are fairly extensive and cover various combinations of taint, OSI, and devices.

Also note, we don't work directly with any ARM or PPC software, so our testing is limited to checking to ensure replay still works for previously recorded ARM\PPC traces. x64 testing was limited to record\replay has not been a priority.

I know this is a big change and I'm happy to help in anyway for additional testing.


nathanjackson:
Maybe a travis issue? Closing\re-opening.

nathanjackson:
Finished some last minute testing and found some edge cases that we've added. Although I'm not sure if the test failures are travis or changes I made yet.

It appears to be travis.

moyix:
I've given Travis a couple kicks, but it seems like with the new version the build is consistently taking longer than the 35 minute time limit. Maybe we should try altering the Travis configuration to only build the x86_64 and ARM builds? I worry about letting PPC bitrot, but having coverage of two diverse 64-bit and 32-bit architectures should help catch a lot of the most common issues.

m000:
Does Travis support snapshots of some sort? A fair amount of time is wasted on installing pre-requisites. If we could start from a snapshot with the pre-requisites installed, testing should be able to complete in <35'.

moyix:
A solution I've used in other projects is to grab a tarball of all the prebuilt dependencies (compiled in a Travis-compatible environment) and then just untar it. We could try doing that here. 

From what I can see, VM-style snapshots aren't supported by Travis:

https://github.com/travis-ci/travis-ci/issues/1338

nathanjackson:
I'm going to try one more time. Maybe since its the end of the day on the east coast there's less load? The tests don't seem to run any differently locally for me.

nathanjackson:
This is definitely a head scratcher... currently we're hung at some PXE boot test, but the same test passes on my local machine. It looks as if it has not even attempted to start the next test. However, if you look at the raw log it indeed has started.

An earlier build today worked fine at that point but failed in the PPC tests. For the record, I've seen this behavior before any of these changes, but its not clear why its not working at all right now.

I'd hate to disable any architecture specific tests in case a problem in one architecture reveals a bug in the core that silently affects other architectures. Also, this very well could be an issue in my PR, but its hard to debug.

nathanjackson:
One more verbose build.

nathanjackson:
Okay, something may have broken the PXE boot test. Running the test manually seems to work fine, the guest is in the state its supposed to be anyway. The issue happens intermittently, but I was able to produce it a couple of times on this branch locally.

Setting this to WIP in case the tests some how miraculously pass and I'm not ready yet.

nathanjackson:
Closing this, I'm going to trigger travis on my fork for testing. I'll re-open a new pull request once this gets worked out.

nathanjackson:
Note it looks like f66e7ac might be the culprit, but I'm not sure why yet.

nathanjackson:
So this looks to be an issue on master as well actually. I think the changes I tried to introduce make the issue more prevalent, but its a problem on master nonetheless. To reproduce the issue on master, I had to run the pxe-test in a loop of 100 tries. It doesn't hang all the time, which is probably why travis sometimes still fails without these changes.

nathanjackson:
I believe I've worked out the issue, there were some changes in the main loop that should not have been brought in. I'm re-running our regression tests and cleaning up. Once that's complete, I'll re-open or create a new PR.

PR #400, opened 2019-01-25
PANDA-QEMU 2.9.1

nathanjackson:
Take 2 of #399.

This PR brings PANDA up to date with QEMU 2.9.1. The long-term goal here is to make PANDA as stable as possible and potential future upgrades of the QEMU version easier.

There are two notable differences between QEMU 2.9.1 and PANDA at this point:

- Changes related to Multi-threaded TCG (MTTCG) were omitted in favor of the PANDA version (by using a combination of git merge -s ours and git checkout --ours) since MTTCG is incompatible with PANDA's replay model.

- During development of QEMU 2.9.1, some APIC changes were introduced. These changes were also omitted as they are incompatible with old replays. However, it appears that the only reason they are incompatible is due to a change in the device's QEMU class structure. I think this could be handled if we had versioned non-deterministic logs or a schemaless format.

Last time we had changes like this we had some issues that broke old replays, but care has been taken to ensure this doesn't happen again.

Our regression tests run various plugins against a set of recordings that we have created and check the output of the plugins. These tests cover the following operating systems on x86:

- Ubuntu 16.04 (Lubuntu flavor)
- Windows 2000, 2003, 7
- FreeDOS

Note that our tests don't cover every combination of plugin\OS, especially for FreeDOS. Even so, the regression tests are fairly extensive and cover various combinations of taint, OSI, and devices.

Also note, we don't work directly with any ARM or PPC software, so our testing is limited to checking to ensure replay still works for previously recorded ARM\PPC traces. x64 testing was limited to record\replay has not been a priority.

I know this is a big change and I'm happy to help in anyway for additional testing.


moyix:
I've tested this out on some of my own (very large, ~170 billion instructions) x86_64 replays that were made before this PR, and it seems to work well. It also seems to make Travis happy. I'm going to see if any of the other PANDA folks can do some testing but I think this looks good from my end.

Thanks so much for helping us keep up to date!

PR #403, opened 2019-02-11
osi_linux: Kernel Profiles and Linux 2.4.x Support

nathanjackson:
This PR introduces the notion of a kernel profile and Linux 2.4.x kernel support in `osi_linux`.

Older variants of the Linux kernel differ significantly from what `osi_linux` currently supports. To achieve compatibility with older and more recent kernels, I've added what I'm calling a "kernel profile". Right now a profile is just a struct that contains kernel-version specific functions. For the default profile (which supports relatively recent kernels) and the Linux 2.4.x profile, we only care about how to get the current task struct and how to get the file descriptors. Thus we only have two functions in the kernel profiles. I would expect this to change though to support even older and newer kernels.

A kernelinfo module for Linux 2.4.x kernels is also included to generate new configs. There are a few fields that aren't yet generated (like the credentials) because they don't appear to be used in `osi_linux` and also differ significantly from the default profile.

I've added version information to the kernel config generation and reading, this allows the correct profile to be loaded at runtime, so I've seen no difference in usage when it comes to actually using OSI dependent plugins.

Our tests make use of OSI (file_taint, ida_taint2) for Ubuntu 16.04 guests and now fully test CentOS 3 (get libraries, get process list, etc). I also used to the run_debian.py script to verify that version of debian was still working. I also added a `panda_require(...)` call to `osi_test` so that you don't have to fiddle with the order of the plugins when that plugin is used.


nathanjackson:
Changing this to WIP. I discovered that at least for a couple of fields I was depending on backported CentOS patches. Need to make a few additional changes.

nathanjackson:
Updated my tests to test Red Hat 8, we're more interested in this distro because the kernel is a vanilla (or closer to it at least) 2.4.x kernel.

Ready for review.

moyix:
@m000 do you want to have a look since you're most familiar with the OSI Linux code?

nathanjackson:
Just an update, when I went back I had to add two additional functions to the kernel profiles: `get_task_struct_next(...)` and `get_group_leader(...)`. The default profile expects Red Hat's Native POSIX Threading library changes to be in place since that functionality was merged back in Linux 2.6. A vanilla 2.4.x kernel does not have those changes, so we have to get the next task and group leader differently.

nathanjackson:
Bump, m000?

nathanjackson:
I rebased this against the latest master yesterday, any chance @moyix you could review it? We are hoping to get this merged soon.

moyix:
Yep, I'll take a look at it tomorrow.

moyix:
Ok, sorry for the long review time. LGTM, I'm going to merge it.

nathanjackson:
Thanks!

PR #404, opened 2019-02-13
Callfunc plugin features (ARM)

tnballo:
3 improvements to the callfunc plugin:

* Save/restore all ARM CPU state before/after the injected call
* Support arbitrary parameter count
* Support injection of arbitrary file into guest memory

Output from Example 1 in `USAGE.md` (slightly revised):

```
root@debian-armel:~# ls
Called function 0xc00bb0cc returned 0xc7970000
```

Output from Example 2 in `USAGE.md` (new):
```
root@debian-armel:~# ls
[  121.175748] Hello from kernel's printk! 1 2 3 4 5 6 7
```

### DIFF
--- a/panda/plugins/callfunc/USAGE.md
+++ b/panda/plugins/callfunc/USAGE.md
@@ -29,11 +29,11 @@ None
 Example
 -------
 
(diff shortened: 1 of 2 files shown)


moyix:
Looks great, thanks!

Issue #381, opened 2018-11-21
taint2 may generate taint for a constant


LauraLMann:
I'm using the i386 build of PANDA, which uses helper functions to calculate condition codes.  One such helper function, helper_cc_compute_c_llvm, if it is provided an CC operation for LOGICB/W/L, CLR or POPCNT, takes the default case in the switch statement in the helper.  This means when taint_select is run in the last block, that the return value (0) is not copied into LLVM, as the return value is a constant.  However, a taint_copy does copy from the (unused) destination in the LLVM taint shadow to the Ret shadow.  Sometimes I get a taint report from this copy.
Is this just leftover taint from the previous time this frame in the LLVM shadow was used?  (The taint_push and taint_pop don't clear taint from the new (for push) or abandoned (for pop) frame, but just change the pointer into the list of TaintData objects.)  It doesn't seem right - shouldn't the helper clear the taint from the destination in LLVM if a constant is returned?

AndrewFasano:
Sounds like this was fixed by #405- if anyone runs into this issue again in the future we can reopen. Thanks for identifying the bug and submitting the fix!

PR #405, opened 2019-02-21
Clear LLVM frames before reusing them

LauraLMann:
The LLVM code generated by the taint2 plugin wasn't clearing the LLVM frames before using them.  This can result in spurious taint reports (when the old taint just happens to be in a destination that is assumed to be 0 and so isn't cleared), or in lost taint reports (when the old taint just happens to be the same value as the new taint being placed there, and so no change report is generated).
This occurred in the block prologues when the previous block did not execute taint_pop_frame (such as after calling a function that doesn't return), because taint_reset_frame was executed AFTER taint_delete on the frame (so the wrong frame was being cleared).
This occurs in the block bodies because there was no attempt to clear out the LLVM frame before taint_push_frame (except for the slots used for the arguments to the function, if they were untainted).
This fix also seems to fix the problem noted in issue #381 - at least it did for one example.  (I didn't look for them all.)

### DIFF
--- a/panda/plugins/taint2/llvm_taint_lib.cpp
+++ b/panda/plugins/taint2/llvm_taint_lib.cpp
@@ -12,11 +12,12 @@
  * See the COPYING file in the top-level directory.
  *
 PANDAENDCOMMENT */
-/*
- * Change Log:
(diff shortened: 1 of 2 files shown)


moyix:
Very nice, thank you for putting this together!

PR #406, opened 2019-02-26
fix get_callers callstack ordering

HighW4y2H3ll:
Guess **get_callers()** is intended to get callstack in reversed order?
According to the usage here:
https://github.com/panda-re/panda/blob/f2ba5a27880bebce47034c68d2f6d1f5eb1c2c32/panda/plugins/pri_dwarf/pri_dwarf.cpp#L1897

### DIFF
--- a/panda/plugins/callstack_instr/callstack_instr.cpp
+++ b/panda/plugins/callstack_instr/callstack_instr.cpp
@@ -393,7 +393,7 @@ Panda__CallStack *pandalog_callstack_create() {
     cs->n_addr = std::min((uint32_t)v.size(), (uint32_t)CALLSTACK_MAX_SIZE);
     cs->addr = (uint64_t *)malloc(cs->n_addr * sizeof(uint64_t));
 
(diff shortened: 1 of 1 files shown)


LauraLMann:
Shouldn't panda_callstack_create get a similar fix to the one made in get_callers?  Those two functions are doing the same thing, except for the kind of list the output goes in.
And I think get_functions needs a similar fix as well - need to index the vector relative to the size of the vector, not the number of items desired to return.

HighW4y2H3ll:
> Shouldn't panda_callstack_create get a similar fix to the one made in get_callers? Those two functions are doing the same thing, except for the kind of list the output goes in.
> And I think get_functions needs a similar fix as well - need to index the vector relative to the size of the vector, not the number of items desired to return.

ahh thanks! I'll take a look there!

moyix:
Just wanted to check – is this ready for review?

HighW4y2H3ll:
> Just wanted to check – is this ready for review?

Yes, it is. :)

moyix:
Great! It looks good to me; thanks for the fix!

PR #407, opened 2019-02-27
Adjust i386 taint2 plugin to handle taint inconsistencies from cpu_exec functions

LauraLMann:
The cpu_exec function calls cpu_exec_enter before starting executing and cpu_exec_exit as it leaves.  The i386 emulator takes advantage of this to do some calculations using the condition code variables stored in the CPUArchState structure (as it hasn't been updating certain values for every guest instruction that needs to, but just storing the info needed to do the calculations - apparently this is faster than updating as they go along).  This can result in different instances of replaying the same recording under the same conditions and the same PANDA executable generating different taint reports, as the cpu_exec function can be interrupted at unpredictable times (eg. to handle debug requests - see #315 for more details).
These changes create two new callbacks that cpu_exec calls to inform the taint2 plugin, when built for the i386, that the condition code variable information needs to be saved or restored, to prevent these inconsistent taint reports.
These changes take care of the i386 specific issue noted in #315 (the more generic issues causing inconsistent taint reports were taken care of in separate pull requests that have already been accepted).


moyix:
Very nice! Thank you for all your work tracking this one down!

PR #421, opened 2019-04-17
ida_taint2.py tweaks

nathanjackson:
This PR makes a couple of minor usability changes to the `ida_taint2.py` script:

- Input files are now validated by the script and the script will display an error message if the parsing fails.
- Made process selection dialog table items non-editable.


moyix:
Nice, thank you!

PR #423, opened 2019-04-18
fix to the register frame base value at function entry

HighW4y2H3ll:
So here's the issue. At the entry of a function, there are code like 
```
push ebp;
mov ebp, esp;
...
```
When parsing the DWARF info in `pri_dwarf`, if the `DW_AT_frame_base` is the `ebp` register, we could actually access the frame base of the caller function instead of the current function.

This is a temporary fix that return 0 for all the frame base if it's a register. Haven't come up with better ways to deal with it...

### DIFF
--- a/panda/plugins/pri_dwarf/pri_dwarf.cpp
+++ b/panda/plugins/pri_dwarf/pri_dwarf.cpp
@@ -1860,6 +1860,7 @@ target_ulong get_cur_fp(CPUState *cpu, target_ulong pc){
         printf("loc_cnt: Could not properly determine fp\n");
         return -1;
     }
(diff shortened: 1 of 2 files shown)


HighW4y2H3ll:
hmmmm, nop, something's broken... It's probably better export the `get_cur_fp`, so the plugin writers could explicitly fix the stack frame register?

moyix:
Could you let me know when this is ready to look at & merge?

HighW4y2H3ll:
Yes! it's ready. Actually... I didn't fix the issue. I just exported `get_cur_fp`, so there would be a way to fix that externally...

PR #424, opened 2019-04-25
ida taint undo

nathanjackson:
A couple more usability changes for ida taint:

- Add undo script that scans for changes that ida taint may have made and remove them. This won't work if you've changed the comments after the fact but.

- Another solution is to just make use of IDA database snapshots. ida_taint2.py now adds a snapshot before it makes any annotations and undo_ida_taint2.py now takes a snapshot before making any changes.


moyix:
Glad to see IDA taint getting some more TLC :)

PR #428, opened 2019-05-01
added configurable rehosting machine

alomk:
Created a new panda callback called "during_machine_init" and hooked it using pypanda to perform the same functionality that was previously done in `/hw/arm/rehosting.c`. Also wrapped more QEMU functions and structures using CFFI


moyix:
Looks good! Thanks!

PR #429, opened 2019-05-04
OSI Linux working from C library

lacraig2:
osi_linux functional. Additional functionality in sending monitor commands from python. better organization and documentation.


lacraig2:
Includes
- General Documentation updates
- OSI linux implementation
- Generalized interface for `osi` for eventual `wintrospection` work
- Updates to plugin importing

@moyix @tleek It would be nice if someone could merge this. I'm giving a presentation on the work I've been doing and it would be helpful to point to this branch instead of my own repo.

PR #434, opened 2019-05-09
Update plugin building part in PANDA.md

DalvikArt:
The plugin building part is out of date. Updated to the latest version.


moyix:
Good catch. Thanks!

PR #435, opened 2019-05-09
use weak slots in port output propagation rules

nathanjackson:
This fixes an assertion error in the Taint 2 plugin. Note, this only occurs if you're using a debug build of LLVM and in certain guests.

I'm not sure about the difference between a {{constSlot}} and a {{constWeakSlot}}, but the fix does work without breaking any of our existing tests.


moyix:
Sounds good. I'm not totally clear on the difference either (maybe @rjwhelan knows?) but I'll merge since it doesn't seem to break things.

PR #440, opened 2019-05-31
get_syscall_retval API call

m000:
`syscalls2` has implemented a profiles-based solution for retrieving system call return values. However, this was not exposed to other plugins, which means that each plugin had to reimplement something similar.


moyix:
Cool, looks good! Thanks.

Issue #411, opened 2019-03-19
Callstack Plugin Sometimes Returns Ridiculous Results


nathanjackson:
Occasionally, we've observed callstack plugin returning ridiculous results. The callstack almost looks as if its repeating itself over and over again, but then there will be a spurious entry. When the plugin returns a callstack that doesn't make sense, it looks like maybe a mode switch or context switch has occurred, but I don't know this conclusively.

We really need this fixed as it provides context for analysis purposes.

moyix:
Is it possible this is related to thread switches? Currently these aren't tracked, although the callstack plugin has a disabled-by-default mode where it tries to guess when a thread switch happens by looking for large jumps in the value of the stack pointer (now that we have support for getting thread IDs in OSI, maybe that would be a better way to check?).

Also, is this a new issue (i.e., since https://github.com/panda-re/panda/pull/406) or one that you've noticed previously?

nathanjackson:
It seems possible (maybe even likely) that this is due to thread switches. The mode\context switch idea was just a theory.

We have seen this issue before and the workaround is to hook PANDA up to GDB and use its backtrace. Using GDB is fine, but its really nice to be able to dump the callstack in a plugin's output to get a better context. I don't think it has anything to do with #406.

I think using the ASID to track anything is likely to have problems. I'm pretty sure that what PANDA uses for ASID is recycled for other processes in at least Linux. I found a forum post about ASID on ARM, I would think x86 would do something similar. See: https://community.arm.com/developer/ip-products/processors/f/cortex-a-forum/5229/address-space-identifier---asid. @moyix do you know something about ASID that I don't know?

It should be possible to use a composite key of PID\TID to track the stacks, but then you introduce a dependency on OSI which may or may not be a problem depending on your use case. It would be nice to select the callstack strategy at runtime as opposed to compile time. There are a few strategies that have various pros and cons:

1. Use the base or frame pointer + sizeof(target_ulong) to traverse the stack - I think this is what GDB does. Pro: No OSI. Con: Doesn't work for binaries compiled with `-fomit-frame-pointer`, also platform specific.

2. Track call\returns and use the PID\TID combination to look up (looks like the default mode). Pro: Should be accurate even without a base pointer. Con: Requires OSI.

[...]

PR #443, opened 2019-06-04
Enhanced Callstack Segregation Options

LauraLMann:
The main thing was to add an option to use the process ID and thread ID to segregate callstacks.  This does a better job of distinguishing between stacks than the ASID.  This technique does require OSI support, so the ASID and USE_STACK_HEURISTIC techniques are still available in case OSI is not available for one's guest OS.  Note that if an OS is specified (via the -os switch), that this new technique (called "threaded") is used instead of ASID.  The "asid" technique remains the default if no OS is specified.
Also changed the old USE_STACK_HEURISTIC option, which was formerly made available only at build time, to now be a run-time option.
Also adjusted the get_prog_point API in callstack_instr, and the plugins which rely upon it (stringsearch, textprinter, unigrams) to not assume the callstacks were segregated via ASID.
I think this takes care of issue #411.

### DIFF
--- a/panda/plugins/callstack_instr/Makefile
+++ b/panda/plugins/callstack_instr/Makefile
@@ -1,8 +1,6 @@
 # Don't forget to add your plugin to config.panda!
 
 # If you need custom CFLAGS or LIBS, set them up here
-# -DUSE_STACK_HEURISTIC tries to detect thread switches by sudden
-# jumps in the stack pointer
 LIBS+=-lcapstone
 
 # The main rule for your plugin. List all object-file dependencies.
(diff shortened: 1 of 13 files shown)


LauraLMann:
Please put a pause on this pull request until I figure out where an infrequent seg fault is coming from.  I don't KNOW it's from these changes, but it did appear shortly after they were made...

LauraLMann:
OK, I've got the seg fault fixed (yes, it was my problem - mea culpa).  Close and reopen to force Travis to review the changes.

LauraLMann:
Reopening to force Travis to start up again.

LauraLMann:
My changes are done now, and Travis is happy.  Please review to see if these enhancements to the callstack_instr plugin (resolving issue #411) are acceptable.

moyix:
This is great! Thanks in particular for taking the time to update the various plugins that use tap points. Very much appreciated :)

PR #446, opened 2019-06-14
Network Taint Refinements

MarkMankins:
Added support for semantic labels to IDA taint.
Plugins that wish to use this feature should create an additional file that maps integer taint labels to a semantic label.
Replaced positional taint labels in tainted_net with semantic labels.  When semantic labels are used, tainted instructions in IDA Pro will be annotated with "Packet Number-Byte Offset" instead of integers.
Added options to tainted_net to refine which network packets and/or which byte offsets in packets are tainted.


moyix:
Nice, thanks!

PR #448, opened 2019-06-20
Better Windows Thread ID Query and In Kernel Flag for Threaded Stack IDs

nathanjackson:
This PR addresses an issue with the Windows thread id lookup we discovered. Basically, looking up the thread ID by starting with the FS segment is only valid in user-mode. In kernel mode, this is sometimes valid but other times not. This pull will fix wintrospection so that the thread id may be looked up for any threads, not just those that have a thread information block.

After updating the thread id lookup, we found some inconsistencies with the new threaded stack type. It turns out that we weren't accounting for kernel and user mode stacks. In Linux and Windows, each thread has a separate stack for kernel and user space as a security measure. By adding a flag, the callstack plugin can differentiate between the kernel and user space stack. Unigrams, textprinter, and stringsearch have been updated accordingly.


moyix:
Thanks for the contribution! The changes look good, I'll merge.

PR #449, opened 2019-06-26
Add coverage plugin

LauraLMann:
Plugin used to spit out a comma-separated-value file noting the information on the blocks executed.
If process mode is used, it notes the process name, process ID, thread ID, whether in kernel mode or not, block address, and block size.
If asid mode is used, it notes the address space ID, whether in kernel model or not, block address and block size.
An IDAPython script is also provided for importing the data to IDA Pro and colorizing and (optionally) annotating the blocks executed for the selected process (or asid).


moyix:
Very nice! This is really useful.

PR #452, opened 2019-07-04
install_ubuntu.sh compatibility with Ubuntu 18.04

m000:
install_ubuntu.sh should now be compatible with Ubuntu 18.04. The main issue was that the PANDA repository doesn't support the 18.04/bionic release. A fallback to 16.04/xenial packages was implemented for that case.
Also reworded README.md a bit to restate that target platform as "the latest Debian stable/Ubuntu LTS". There may be issues when a new release is made, but these issues should be considered as bugs and reported so they can eventually be ironed out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,28 +21,41 @@ Laboratory, NYU, and Northeastern University.
 
 ## Building
 
-###  Debian 7/8, Ubuntu 14.04 / 16.04
+###  Debian, Ubuntu
 Because PANDA has a few dependencies, we've encoded the build instructions into
 a script,, [panda/scripts/install\_ubuntu.sh](panda/scripts/install\_ubuntu.sh).
-The script should actually work on Debian 7/8 and Ubuntu 14.04, and it
-shouldn't be hard to translate the `apt-get` commands into whatever package
-manager your distribution uses. We currently only vouch for buildability  on
-Debian 7/8 and Ubuntu 14.04, but we welcome pull requests to fix issues with
-other distros.
+The script should actually work on the latest Debian stable/Ubuntu LTS versions.
+It should be straightforward to translate the `apt-get` commands into whatever
+package manager your distribution uses. We currently only vouch for buildability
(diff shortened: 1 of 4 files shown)


moyix:
Nice! So this just pulls down the 16.04 PPA to get LLVM 3.3? At some point we should probably figure out if it's possible to package that for 18.04 (or, even better, try to migrate to the newer LLVM JIT API...).

m000:
Yes, only LLVM/clang 3.3 are pulled from the 16.04 PPA. The rest of the packages have more recent versions in the main 18.04 repositories, so they are ignored.
Maybe the best option for cross-platform building would be to switch from a custom shell script to [Ansible](https://www.ansible.com/) or a similar tool. This should abstract a lot of the details we have to take care in the custom shell script. Ansible seems to be a good choice because it is pure-python, so it won't introduce additional build dependencies. We can merely create a simple wrapper script (like the one we use for [syscalls2](/panda-re/panda/blob/master/panda/plugins/syscalls2/scripts/make_all_prototypes.sh)) which will download it on the spot.

nathanjackson:
@moyix hypothetically speaking, if one were to migrate to a newer LLVM, should they target the latest LLVM or the one that comes with the current LTS version of Ubuntu (like 18.04)? I have no idea if it has the ORC JIT or not.

My thought is that it would be ideal if we could use a distro’s LLVM (and libdwarf if it’s possible) packages - one less custom built dependency would be great from a maintainability/getting started/packaging perspective.

m000:
@nathanjackson LLVM have their own repos for Debian/Ubuntu: https://apt.llvm.org/. These repos should provide more recent versions than Ubuntu, without having to do keep them up to date ourselves.

moyix:
ORC is definitely what we'd like to move to. I believe S2E is currently using MCJIT, and is locked to LLVM 3.9 – which is certainly better than 3.3, but it would be great to be able to go all the way up to a modern LLVM.

PR #455, opened 2019-07-08
Add ability to collect coverage data through monitor commands

LauraLMann:
Enhanced coverage plugin so that monitor commands can be used to enable and disable the collection of coverage data.
This required fixing the monitor callback framework (thank you @nathanjackson ) so it was hooked up properly.


moyix:
Looks good! Thanks for fixing up the monitor command infrastructure too!

PR #456, opened 2019-07-25
callback support for enable/disable callback and support for multiple architectures in pypanda

lacraig2:
I thought this was already a thing, but apparently not.


AndrewFasano:
Merging without any testing because it's just the pypanda branch. @moyix could you give @lacraig2 push access for the branch?

PR #460, opened 2019-08-08
Make PANDA Installable

nathanjackson:
I've been working on packaging up PANDA. The goal here is to make it so that PANDA can be installed just like QEMU. A few longer-term goals that we're working toward along this line: package with deb\rpm, a PANDA SDK (building plugins outside the build tree).

There a few notable changes:

- The PANDA binaries are now named `panda-system-<arch>`. This lets you install QEMU and PANDA side by side. Plus, I think its actually a bit more user friendly - since now we're running a panda binary instead of a qemu binary.
- Plugins are installed in `<prefix>/lib/panda/<architecture>/`. PANDA will try to resolve plugins in the following order: relative to PANDA_PLUGIN_DIR -> relative to the PANDA binary (current functionality) -> relative to the install prefix (new functionality).

I'm marking this PR as a work in progress to work out any Travis CI issues and hopefully to get some input from the other users\developers as this will at the very least require you to use `panda-system-<arch>` instead of `qemu-system-<arch>`.

That being said... thoughts?


m000:
Sounds good. This is probably something long due. Getting to the point where deb/rpm packages can be produced would be great for expanding the community. I'd like to design a PANDA student assignment for our binary analysis course here at @VUSec. But with the current state of things, we'd have to spend too much time just to get PANDA running – not a good investment of the limited course time.

nathanjackson:
Our internal tests are passing. This PR is ready.

moyix:
Thanks! Do you think you could fix up the `run_debian.py` script? I think it's a pretty small fix, but since that script calls out to `qemu-system-<arch>` I think it's broken by this PR.

nathanjackson:
Updated. I also updated the `diverge.py` script.

Although, I did a grep over the files to see where there are still references to "qemu-system-*". There are still some references to it, but they seem to be mostly in files that I don't have a way to test or are in QEMU documentation.

I'll leave that up to @moyix to whether or not they should be changed, but if I change them its going to be a find and replace without much testing other than our test suite (`make check` and our internal tests).

m000:
Question: After installing PANDA, can you now drop the source tree and continue development using only the installed headers and libraries?

nathanjackson:
The installation does not install any headers or any libraries. We definitely want the ability to do this.

m000:
Cool! Automating the creation of docker images from builds would also be interesting. Is this possible now? Any pointers for adding this feature? (In case you're not a docker n00b like myself :).)

moyix:
Thanks for taking a look at it @AndrewFasano! And sorry it took us so long to get to it. I'll merge now.

PR #461, opened 2019-08-13
Add 64-bit and linux kernel 4.9 support to osi_linux plugin

LauraLMann:
These changes adapt the changes made by @eShard on an older version of PANDA to work with the newer PANDA, by using the per cpu variables to get the current task information (except when working with the really old 2.4 kernel).
Note that I wasn't quite sure how to test the changes to on_get_current_process_handle added by @m000 because, at least for old kernels, the pre-changes version didn't produce reasonable looking results.  (With kernel 2.4 guest, the handle returned could never be turned into a process.)  So, @m000  and @moyix might want to take a closer look at that function in case I missed something.


LauraLMann:
Does the lack of comment from @m000 or @moyix mean there are no changes needed related to on_get_current_process_handle?

m000:
@LauraLMann Sorry, somehow I missed the initial tag. I'll check if this PR works with existing code in the following days.
I'm still puzzled that the process handles don't work with kernel 2.4. Any insight on what causes this? The address of the task struct _should_ remain constant through the lifetime of a process. In principle it could be reallocated and then the process list links adjusted, but this wouldn't make sense in a performance-critical component like the kernel.

LauraLMann:
No, @m000, I don't really have any good ideas as to why on_get_current_process_handle didn't work before these changes with the 2.4 kernel.  (I was mainly interested in not breaking what already worked.)  I was wondering if perhaps there was a merge problem - that a simple merge wasn't the answer when it was merged in.  I noticed that per the dates on the commits, the on_get_current_process_handle changes and the changes by @nathanjackson that split the kernel profile into old (2.4) and new profiles were done at about the same time.
With the changes in this pull request, the value returned by get_current_process_handle can be turned into a process handle for my kernel 2.4 guest, and the process it finds has the same name as the process found directly via get_current_process.  This seems like an improvement, even though I don't know why it didn't work before these changes.

m000:
@moyix @LauraLMann Has this PR been tested on any 4.4 32bit ubuntu kernel? I've updated my kernel configuration on my (PR461-test](https://github.com/m000/panda/commits/PR461-test) branch, but the test code fails from the first attempt to retrieve processes.

```
osi_linux.cpp:515: int osi_linux_test(CPUState*, target_ulong, target_ulong): Assertion `ps != NULL && ps->len > 0 && "no processes retrieved"' failed.
```

m000:
If on top of [PR461-test](https://github.com/m000/panda/commits/PR461-test) I revert the changes related to this PR, I get back to a working state.

```
git revert 90dccb2f54c8eeaf58a45cb09a59dfd4b7c7a3a0
git revert -m 1 c655b668179baea9efeb49fce76e9c48da6e03d9
```

LauraLMann:
Yes, I tested with UbuntuServer 32-bit kernel 4.4.

LauraLMann:
@moyix and @m000, do you have a specific version to test, or do you have a disk image and recording we can try to duplicate this problem?

m000:
Sure.
Image: https://send.firefox.com/download/1e31d5278b0687e7/#mPPEa3tofxGqXwZHgcHeHQ
Recording: https://send.firefox.com/download/02771c3f639de0a8/#q6IIno3A1KT94iGdTSCoBw
Command line: `panda-system-i386 -m 512 -display none -panda 'syscalls2:load-info=y;osi;osi_linux' -os linux-32-ubuntu:4.4.0-130-generic -pandalog my.plog -replay ubuntu16-simple`

m000:
Ah. And `panda/panda` as credentials.

m000:
@LauraLMann @moyix Ok. I've figured it out! There was a difference between rr environment and the kernel profile extraction environment:
1. **kernel profile extract environment**: KVM enabled, 4GB of memory
2. **record/replay environment**: KVM disabled, 512MB of memory

I've always been doing that without issues. The extracted profile was the same in either configuration.

However, with the new PR `task.per_cpu_offset_0` is affected by either KVM acceleration or the amount of memory. I haven't been able to verify which of the two is the culprit, but you can stop looking for now.

I'll try to verify the source of the problem and add a fat warning somewhere. I *wish* that the problem is with KVM acceleration. Because otherwise it means that the kernel profiles are now *configuration specific*.

nathanjackson:
@m000 I know you're on a 32-bit OS, but you might try disabling ASLR.

https://askubuntu.com/questions/318315/how-can-i-temporarily-disable-aslr-address-space-layout-randomization

We knew expected this was going to be an issue for 64-bit, but maybe its also a 32-bit issue with these variables?

Ideally, we could figure out how to get around ASLR.

m000:
@nathanjackson That is for user-space ASLR which doesn't kernel memory layout. I remember having concerns about the effects of ASLR on osi_linux, but I haven't seen any problems so far: Only  some base addresses are randomized. As far as we rely on offsets, we should be fine.

Here is per_cpu_offset_0 for 3+1 different configurations.

```
# 512 MB - no kvm
#task.per_cpu_offset_0 = 0x1DE18000
task.per_cpu_offset_0 = 501317632

# 512 MB - kvm
#task.per_cpu_offset_0 = 0x1DE17000
task.per_cpu_offset_0 = 501313536

# 4096 MB - kvm
#task.per_cpu_offset_0 = 0x34129000
task.per_cpu_offset_0 = 873631744

# 4096 MB - no kvm
# can't boot! stops after "switched to clocksource tsc" on console
```

m000:
I've opened #476 regarding this change in the properties of kernel profiles. Any discussion can resume there.

PR #462, opened 2019-08-13
Minor bug fixes

MarkMankins:


panda/plugins/osi_linux/osi_linux.cpp
```diff
@@ -367,7 +367,9 @@ void on_get_libraries(CPUState *env, OsiProc *p, GArray **out) {
 	return;
 
 error0:
-	g_array_free(*out, true);  // safe even when *out == NULL
+	if((out != NULL) && (*out != NULL)) {
```
m000: What is the use of the second check? I remember that the first check will hush some glib warnings that may be annoying depending on the use. But I'm not sure for the use of the second check.

m000: The first check is redundant. If `out == NULL` you're not calling the function properly. It is vain to try to make the code *not* crash.

Also, can you check for other similar cases in the file and add the `*out != NULL` check?

MarkMankins: You are correct.  I'll commit an updated fix to address both issues.  Thanks!

MarkMankins:
Travis take two

moyix:
Thanks! These look good.

PR #463, opened 2019-08-15
panda_get_retval() inline

m000:
Somehow this was still missing. Yet another wrapper function that abstracts platform details.


m000:
Note: I've googled for the proper ARM/PPC return values. It would be nice if someone with hands-on experience on the platforms could make a quick check.

nathanjackson:
I don't know about PPC, but you're correct about ARM. R0 is typically used for return value.

PR #464, opened 2019-08-19
Build on OSX with Ansible

m000:
The PANDA build on OSX is broken because many dependencies have been dropped from homebrew. This is an effort to fix building on OSX. To make this maintainable in the long run, the build uses an [Ansible](https://en.wikipedia.org/wiki/Ansible_(software)) playbook rather a shell script. Ansible playbooks may provide a viable alternative for Linux builds as well.

Configure script had to be reworked in some places (better detection of llvm, proper detection of protobus), and while I was at it, I also cherry-picked some relevant changes from the QEMU repo. Clang is much more verbose with warnings, so I tried to reduce them to a reasonable level.


m000:
At this point, you can again build PANDA on OSX! Just install the formulas from playbooks/resources/homebrew. Then use this gist to properly configure PANDA: https://gist.github.com/m000/abf9225f7b8da1de6ccad69937652414

I didn't have time to actually make a test run though.

m000:
I'm finished working on this (sans any fixes to make Travis happy). PANDA now builds on MacOS. The first time you use the script, it should take a fairly long time because LLVM 3.3 will be built from source.

In the end, you will be able to build PANDA. But **it won't run** because of changes to the Cocoa framework. With some strategic pulling and merging, we should be able to have it working on MacOS again. But let's leave this for another PR.

I've tested these changes also in Linux. I didn't notice any side-effects.

m000:
For the record, here is the error message when running `qemu-system-i386`.

```
2019-08-23 00:37:21.510 qemu-system-i386[20850:12095761] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'nextEventMatchingMask should only be called from the Main Thread!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff2eb752fd __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff59242a17 objc_exception_throw + 48
	2   AppKit                              0x00007fff2c0b4f8e -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 4180
	3   qemu-system-i386                    0x00000001061e80d9 cocoa_refresh + 233
	4   qemu-system-i386                    0x00000001061e0948 do_safe_dpy_refresh + 24
	5   qemu-system-i386                    0x000000010606c93c process_queued_cpu_work + 140
	6   qemu-system-i386                    0x0000000105f7d5b2 qemu_tcg_cpu_thread_fn + 226
	7   libsystem_pthread.dylib             0x00007fff5ac042eb _pthread_body + 126
	8   libsystem_pthread.dylib             0x00007fff5ac07249 _pthread_start + 66
	9   libsystem_pthread.dylib             0x00007fff5ac0340d thread_start + 13
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

m000:
I believe I understand Travis' problem. Commit 5173bbf switches from the traditional `-I` for adding header paths to the modern `-iquote` (for which I can't find any documentation).

This makes all angled includes e.g. `#include <linux/kvm.h>` to always look for the file out of the source tree. Apparently, until now these headers were picked from the PANDA source tree. Hopefully this can be resolved by adding another package to the CI environment.

m000:
@moyix , can you try adding `linux-libc-dev` to the CI environment and see if the problem goes away? Apparently me adding it to the PR won't do the trick.

m000:
@moyix , can you try adding linux-libc-dev to the CI environment to see if this will pass the build tests?

moyix:
I don't see anything in the Travis control panel. Have you tried adding the .travis.yml change to your PR?

m000:
@moyix Yes, I rebased last night. I tried rebasing again just in case I missed something the first time.

m000:
I first tried pulling some newer headers in the build environment, but that caused datatype conflicts. Then I tried disabling some stuff, but still no luck. Currently missing some definitions in `vfio.h` which were added at later kernel versions.

I'm not sure if it's worth trying more. Since Ubuntu LTS 14.04 is no longer officially supported, maybe it's time to consider following suit for the Travis build environment. Otherwise, I'll see what can be salvaged from this PR without breaking Travis and save the rest for later.

moyix:
I tend to agree. Maybe we can try to do a separate PR that moves the Travis CI environment to 16.04 (or even 18.04?), get that working, and then rebase this PR on that?

m000:
Ok, let's go that way.

m000:
Travis environment bumped up in #478.

m000:
Same problems. At least after bumping Travis to 16.04LTS I can try debugging them locally without creating a new image just for this.

m000:
@moyix I've fixed the building issues. Turned out that the linux header inclusion flags should have continued to use "-I". I had accidentally changed them to "-iquote". Along the way, I've added fixes for building on x86 (in case anyone ever needs it).

Travis seems to take too long for the `check-qtest-*` tests. On my workstation those run in ~145sec, but Travis environment doesn't seem to be as fast.

m000:
This is ready to merge, ideally after merging #474 which has been tested by multiple parties.

Changes in this PR affect mostly the build machinery. There are changes in source files, but it is mostly type fixing to make the clang compiler shipped with osx happy. I've done some testing on traces I have I didn't spot any issues.

moyix:
Looks like merging #474 added some conflicts, do you want to rebase?

m000:
Rebased and fixed some minor issues with the playbook and the llvm formula. Also simplified code searching for kernelinfo a bit.

PR #466, opened 2019-08-22
Add option to replaymovie plugin to put instruction counter on each movie frame

LauraLMann:
The title pretty much says it all.


LauraLMann:
In case you're wondering, this little counter is helpful in looking for a good place to cut out a section of a recording via the scissors plugin.

PR #473, opened 2019-09-16
Add runtime plugin/plugin dependencies

MarkMankins:
Update the plugin build system to read an optional plugin_plugin.d file
in each plugin directory.  This file contains dependency information for
plugins that have runtime dependencies on other plugins.

Runtime dependencies can be required when using C++ classes in a plugin.
If plugin_1 wishes to make use of C++ classes that are defined in plugin_2,
one could update plugin_1's Makefile to add a runtime dependency on plugin_2
by modifying the LIBS symbol, such as:
LIBS+=-L$(PLUGIN_TARGET_DIR) -l:panda_plugin_2.so

However, this is not enough to guarantee a clean build.  The build system needs
to know to build panda_plugin_2.so prior to building panda_plugin_1.so.

Makefile.panda.target has been modified to look for an optional
plugin_plugin.d file in each plugin directory where these dependencies can
be defined.  In this case, a plugin_plugin.d file would be required in
plugin_1's source directory that contained:
plugin-plugin_1 : plugin-plugin_2
or if plugin_1 and plugin_2 live in the extra panda plugins directory:
extra-plugin-plugin_1 : extra-plugin-plugin_2

### DIFF
--- a/panda/Makefile.panda.target
+++ b/panda/Makefile.panda.target
@@ -63,6 +63,10 @@ PANDA_CONFIG_FILES=$(SRC_PATH)/panda/plugins/config.panda \
 				   $(SRC_PATH)/panda/plugins/config.llvm.panda \
 				   $(if $(EXTRA_PLUGINS_PATH),$(EXTRA_PLUGINS_PATH)/panda/plugins/config.panda,)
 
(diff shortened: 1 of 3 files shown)


moyix:
Good catch – thanks!

PR #474, opened 2019-09-16
making kernel profiles grea^H^H^Heneric again

m000:
Addresses #476.
On the side, 06c0cce adds `get_struct()` C++ templates for offset-based struct traversal. This enables the eventual elimination of the `IMPLEMENT_OFFSET_GET*` hack. Some work has been done towards this end.


m000:
@LauraLMann @nathanjackson Can you give this PR a go to see if anything is amiss?

LauraLMann:
Am I missing some new directions?  I tried copying the new utils/kernelinfo/ folder and contents to a 32-bit Unbuntu Server guest where I have successfully built kernel config files in the past.  It has installed linux-headers-generic, and is at kernel 4.4.0-98-generic.  But when I "make" with the new kernel config files, I get a complaint that in kernelinfo.c line 156 that __per_cpu_offsets is not declared.

m000:
@LauraLMann Ooopsie! That was a typo - we use `task.per_cpu_offsets_addr` in the CPU variable is `__per_cpu_offsets`. Commit f30bf60 should fix this.

LauraLMann:
After @m000 's latest changes, I can now rebuild the kernel information files under 32-bit ubuntu server, 32-bit debian, 64-bit ubuntu and 64-bit debian.  However, when I run our regression tests, there are some failures.  I will have to investigate them to see if the cause is related to these changes or to something else.
Both of the tests using 64-bit osi failed.  One of them is using a kernel configuration which I generated.  The other is using one from the kernelinfo.conf file that comes with the osi_linux plugin, and so was updated in this pull request.  (I assume it's matching "ubuntu:4.4.0-154-generic:64", as the os specified in the failing test is "linux-64-ubuntu:4.4.0-154-generic".)

LauraLMann:
I think something has gone wrong in this pull request, at least for 64-bit guests.  The osi_test results with these changes result in ridiculously large parent process IDs.  I had originally verified a sampling of parent process IDs by doing "ps -eLf" in the guest immediately after taking the recording that I use for testing, and comparing what "ps -eLf" displays with what osi_test outputs.  For example, "ps -eLf" says that for process Xorg that the process ID is 385 and the parent process ID is 376.  But the new osi_test results say the process ID is 385 (good), but the parent process ID is 1614907703672.  This is on the "Current process" line, which outputs the name, PID and PPID fetched via get_current_process.  The parent process IDs are similarly ridiculous in the process list printed out just afterwards (but are consistent with what get_current_process returned for the PPID).  The numbers in the dynamic libraries list seem OK.
My tests give up due to errors before they get around to testing the get_current_thread, and get_process_handles functions, so I can't tell you if they have similarly gone off in the weeds.  But I can force them to produce that output, if you think it would be helpful.

m000:
@LauraLMann Can you share a VM image so I can test this?

LauraLMann:
Changing target_pid_t to uint32_t does help the 64-bit tests to pass.  I also had to change osi_test.c where it formats the process ID and parent process ID for printing, as the formatting appropriate for a target_ulong is not appropriate for a uint32_t in the 64-built PANDA build.
I am now running the rest of our regression tests to see if this change is OK with them too.

m000:
@LauraLMann Thanks for checking. Does e7ee597 cover all the changes you had to make?

LauraLMann:
Yes.  Our full suite of regression tests are still running (as a variety of them sprinkled about use osi), but so far they are all passing.  I did run some 32-bit and 64-bit linux osi related tests individually, and they all passed.  So I'm relatively confident this change hasn't broken something else.

LauraLMann:
Our full suite of regression tests are passing with the latest changes (well, my draft version of them).  I think this pull request is OK now (although it appears Travis does not agree...).

m000:
This is ready to merge.

moyix:
Merging! :)

PR #479, opened 2019-09-24
Add memorymap plugin (and fix wintrospection bug)

LauraLMann:
The main purpose of this pull request is to add a memorymap plugin to dump the process mapping at a given instruction count or instruction counter (or all of them, if you have the patience).  For each location of interest, the plugin determines which kernel module or dynamic library, if any, the instruction is in and writes that out.
As a little bug in wintrospection got in the way of testing this plugin, it was fixed.  (One path out of the on_get_libraries function didn't restore cr3 like it should.)


moyix:
Looks good, thanks!

Issue #482, opened 2019-09-30
asidstory1 regression test now fails, ASID value differs between runs with same replay


jnmartin84:
The "asidstory1" regression test now fails because on successive runs with the same replay file, it generates different values for ASID for the netstat process.

I was able to reproduce with the latest master on Ubuntu 16.04 (x86-64) as follows:

1)    clone panda from github
2)    cd panda
3)    run ./panda/scripts/install_ubuntu.sh
4)    export PANDA_REGRESSION_DIR=path/to/regdir
5)    modify ./panda/testing/tests/asidstory1/asidstory1-test.py so it uses the correct OS string ("linux-32-debian:3.2.0-4-686-pae")
6)    modify ./panda/testing/tests/config.testing to enable asidstory1 so "init" step creates necessary directories under PANDA_REGRESSION_DIR
7)    cd panda/testing
8)    run python ptest.py init
9)    run python ptest.py bless asidstory1
10)    run python ptest.py test asidstory1

On step 10, the test will fail because the values in the ASID column differ between the blessed output and the test output.

Some sample output:
```diff
$ diff /home/panda/regdir/blessed/asidstory1/asidstory1.out /home/panda/regdir/tmpout/asidstory1/asidstory1.out 
    2c2
-    <     371  2476              netstat2      7f42  1643958  ->  7239683
    ---
+    >     371  2476              netstat2      7f20  1643958  ->  7239683
    8c8
-    <       5  2433                 bash2      7f42  1052940  ->  1092800
    ---
+    >       5  2433                 bash2      7f20  1052940  ->  1092800
```

This is repeatable, with different values in the ASID column (fifth column in output) every time.

jnmartin84:
git bisect points me to the following commit where the test first broke:
```
$ git bisect bad
2a37c18f8fff92080ee15c8ab6b48d0713228230 is the first bad commit
commit 2a37c18f8fff92080ee15c8ab6b48d0713228230
Author: Nathan Jackson <nathan.jackson@ngc.com>
Date:   Mon Sep 9 17:48:09 2019 -0400

    add current process cache to osi_linux
```

nathanjackson:
I think I have a fix on my branch issue-482: https://github.com/nathanjackson/panda/commit/a0078e716bb66c432e652982300f5d37ae765235

@jnmartin84 Would you mind testing this? If it works I can submit a PR.

jnmartin84:
Will do.

jnmartin84:
[...]

PR #483, opened 2019-10-01
add asid to cached process info in osi_linux

nathanjackson:
This PR should address issue #482.


nathanjackson:
If at first you don't succeed, try again.

PR #495, opened 2019-10-30
Add 64-bit linux support to syscalls2 plugin

LauraLMann:
There are some restrictions:
only 64-bit linux was added (not Windows)
system calls marked "ptregs" are not processed by the parsing scripts and so not supported
system calls marked x32 specific (which have ids up in the 500s) are not supported as their signatures couldn't be found
32-bit system calls (ie. via "int 0x80" or "sysenter") encountered while the 64-bit profile is active are not processed - instead a warning is printed out
These changes have only been extensively tested with the file_taint plugin, and a few little problems in that plugin that got in the way of testing were fixed along the way.
Adjustments were made in the loaded plugin so at least it will build and still support the 32-bit guests it used to support, but it was not modified to support 64-bit linux.
Also, 64-bit linux support was enabled in the callstack_instr plugin for the threaded stack type (which was useful for testing purposes).


jnmartin84:
Continuous Integration failure.
"asidstory1" regression test failed.


tcmalloc support  no
jemalloc support  no
avx2 optimization yes
replication support yes
build.sh: Compiling PANDA...

[32m[panda_install][0m [1mPANDA is built and ready to use in panda/build/[arch]-softmmu/panda-system-[arch].[0m
END PANDA BUILD
BEGIN QEMU TESTS
END QEMU TESTS

LauraLMann:
How do I investigate the PANDA Jenkins Build failure when I don't have a Jenkins account?  What's the next step here to get this pull request through the appropriate wickets to be accepted?

jnmartin84:
I am going to disable that check for now and see if I can figure out what is going on with that error.

jnmartin84:
I have adjusted the status to remove the failure indication for the disabled test.

LauraLMann:
Close and reopen to force Travis to try again.

LauraLMann:
One more time...

LauraLMann:
In order to more easily resolve the merge conflicts in files I DID NOT TOUCH, I am closing this PR, and submitting a new one that manually reapplies my changes on top of the latest master.
Given that doing a forced push got us into this mess, I don't think doing a forced push to resolve the new conflicts is a safe way to deal with them.

Issue #491, opened 2019-10-22
tainted_net plugin fails to load due to #487


nathanjackson:
Our regression tests are failing after merging #487.

```
panda-system-i386 -m 1024 \
  -vga cirrus -netdev user,id=net0 \
  -device e1000,netdev=net0 \
  -replay test_tainted_net_semantic_labels \
  -os linux-32-redhat:2.4.18-14 -panda tainted_net:eth_type=0x0800,label_incoming_network=True,bytes=1-3:5,ip_proto=6,semantic=True,ip_dst=192.168.100.60,packets=84,ip_src=192.168.100.1
PANDA[core]:os_familyno=2 bits=32 os_details=redhat:2.4.18-14
PANDA[tainted_net]:adding argument eth_type=0x0800.
PANDA[tainted_net]:adding argument label_incoming_network=True.
PANDA[tainted_net]:adding argument bytes=1-3:5.
PANDA[tainted_net]:adding argument ip_proto=6.
PANDA[tainted_net]:adding argument semantic=True.
PANDA[tainted_net]:adding argument ip_dst=192.168.100.60.
PANDA[tainted_net]:adding argument packets=84.
PANDA[tainted_net]:adding argument ip_src=192.168.100.1.
Failed to load build/i386-softmmu/panda/plugins/panda_tainted_net.so: build/i386-softmmu/panda/plugins/panda_tainted_net.so: undefined symbol: on_replay_handle_packet
FAIL: Unable to load plugin `build/i386-softmmu/panda/plugins/panda_tainted_net.so'
Aborted (core dumped)
```

@m000 I'm not really sure what's going on. Nothing obviously about your changes should have broken the plugin, but reverting 19ce8ae fixes it. Can you take a look?

m000:
Ok. Will take a look. Will send an update later today.

m000:
An objdump is revealing—indeed nothing to do with #487:
```
% objdump -tC ./i386-softmmu/panda/plugins/panda_tainted_net.so | grep on_replay_handle_packet
00000000000041b0 g     F .text	000000000000085e              on_replay_handle_packet(CPUState*, unsigned char*, int, unsigned char, unsigned int)
0000000000000000         *UND*	0000000000000000              on_replay_handle_packet
```

The plugin is C++ and `on_replay_handle_packet()` is declared inside the `extern "C" {}` block. Moreover, its definition in C++ doesn't match the declared signature, so the linker doesn't pick it up.

[...]

PR #487, opened 2019-10-12
Assorted fixes/improvements.

m000:
This PR contains assorted fixes and improvements that have been piggybacked on #464. The fixes are not directly related to the objective of #464. These should be straightforward to review and merge, so that the latter becomes less cluttered.

PR #497, opened 2019-10-31
Cleanup for PANDA callbacks + RR control API

m000:
**PANDA callbacks cleanup**
- PANDA callbacks definitions are moved to `callbacks/cb-defs.h` (plugin-side) and `callbacks/cb-support.h`. The latter is auto-generated using the former, minus some exceptions. This makes it easier to keep the two files in-sync. Before, the callback definitions were thrown in `plugin.h`. Also, the old `callback_support.h` header had several naming discrepancies compared to `plugin.h`. This made it harder to sync the occasional changes between the two.
- Most callbacks were declared to return `int` for no good reason. With very few exceptions, the return value of callbacks was not used. For this, return type of callbacks was changed to `void` for most of the callbacks.
- In many cases the argument types of the callbacks were not semantically correct. E.g. using `uint64_t` for addresses or `uint32_t` for sizes, or `void *` for a byte buffer. Types have been fixed for most callback arguments to reflect the actual meaning of the argument. This should help improve the readability of plugin code.
- Comment block preceding callbacks definitions were not uniformly formatted. They have now been reworked so they can be easily parsed. This would allow e.g. to automatically convert them to Doxygen format, so that PANDA documentation can be generated automatically, rather than having to manually update a bunch of markdown files.

**RR control API**
Starting/stopping record/replay required setting some (undocumented) magic globals. These would be checked by the main PANDA loop and carry on the requested action. The magic globals and the `rr_mode` globals have been merged into a new type `rr_control_t` defined in `rr/rr_types.h`. An API for starting/stopping r/r is provided in `rr/rr_api.h`. The changes between modes are now modelled as an FSM, so code using the API can test whether the requested action can be carried out or not.

**Plugin symbol checker script**
Script `build-check.sh` has been added in `panda/scripts`. The script checks all the plugin shared libraries for missing symbols. Symbols must exist either in the plugin itself, or in the PANDA binary. The script can be integrated with `build.sh` and CI, to intercept issues similar to #491.


m000:
@nathanjackson has already tested this PR. It didn't cause any problems with his test set.

nathanjackson:
@m000 , I've not run the tests against our internal plugins, only the public plugins. I haven't had a chance to go update our internal plugins. Before I go off trying to update them, is it expected that we will have to update them?

m000:
@nathanjackson Some updates will be required but they should be straightforward. From the changes required for the public plugins, the following changes should be enough:
- Fixing the return type of callbacks to `void` and updating `return` statements.
- Fixing the types of callback arguments. You can do this lazily: If the compiler doesn't complain and `build-check.sh` comes clean, you should be fine without changing anything.
- Updating `printf` formats in some cases. `TARGET_PTR_FMT` can be used for guest pointers and `%zu` for `size_t`. Again, if the compiler doesn't complain, leaving the code as-is is fine for an internal plugin. For public plugins it is good to use the proper format so that the plugins can be compiled for all supported host platforms.

nathanjackson:
I've got everything on our end updated and those tests pass now too. The changes to the APIs are good and make much more sense now.

I'm sure @moyix loves it when I poke him to merge stuff 😃 but we need this PR sooner rather than later since it fixes some plugins that we use on a regular basis. Plus we've got some additional PRs already in that we need to revisit in light of #487 and this PR.

PR #499, opened 2019-10-31
Update Manual Section "plugin zoo"

rquarry:
Changed broken "README.md" links for some plugins that exist in v2 repository from old panda1 path.


moyix:
Thanks!

PR #501, opened 2019-11-05
Fix for #500

### LINKED ISSUE #500: Debug Builds are Broken
Sorry @m000, we just ran into another issue but this time with the RR API changes. I didn't try a debug build until now but we're getting a compilation error if you pass `--enable-debug` to the configure script.

```
panda/cpu-exec.c:717:9: error: ‘rr_on’ is static but used in inline function ‘debug_checkpoint’ which is not static [-Werror]
     if (rr_on() && cpu->rr_guest_instr_count >> 17 > counter_128k) {
```

m000:
Declared two functions defined and used locally in `cpu-exec.c` as `static inline`. This should fix #500.

### DIFF
--- a/cpu-exec.c
+++ b/cpu-exec.c
@@ -709,10 +709,8 @@ void debug_counter(void) {
         counter_128k++;
 }
 #endif
-__attribute__((always_inline))
-    inline void debug_checkpoint(CPUState *cpu);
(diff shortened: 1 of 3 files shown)


m000:
Travis is being naughty again.

PR #503, opened 2019-11-05
Update Arch install script to include new GPG keys.

ajgay:
Tested on Manjaro 18.1.1.


moyix:
Thanks for this! Will merge now.

nathanjackson:
Whoa, is this supposed to have the PANDA binaries in it now under `install/`?

nathanjackson:
This probably needs to be backed out... we really don't want binaries in the repo.

ajgay:
I don't believe my initial commit was responsible for that. If you look at the diff for 654c025 you see its very simple. I did notice that `build.sh` did default to the local `install` directory however. That seems to be fixed with the latest merge however, no?

ajgay:
Sorry, just noticed you meant pre-build binaries in `install` not the target of the `build.sh` script. Agreed the bins should be removed. Not sure how they slipped through.

nathanjackson:
This commit added the binaries to the repository: 3acf5d4c67e7e7ad2e77b0df9fd2fc71a2313ac4. Since this is a merge from upstream into your fork, it probably means your master branch was out of date from a few weeks back when someone else I think pushed binaries into the repo.

@moyix You probably want to fix this before accepting any additional PRs by removing the relevant commits from the history and force pushing, otherwise the binaries will always exist in the repo.

If you need help, I more or less know what needs to be done.

m000:
The PR needed rebasing before merging. I can see several commits from previous PRs that have already been merged.
@ajgay use `git pull -r upstream master` to rebase your PRs onto master, rather than merging with master. This generally produces cleaner PRs.

m000:
@moyix I think that the quickest solution would be: (a) doing a hard reset to the HEAD before this was merged, (b) cherry-picking 654c025, (c) force-pushing.

@ajgay 654c025 is the only thing you wanted to include, right?

ajgay:
@m000 Yes, that is the only needed commit. The rest were added as a simple merge as the PR sat for several weeks.

Issue #504, opened 2019-11-06
handle packet old_buf_addr being truncated


LauraLMann:
It appears some of the changes in pull request #497 have broken the old_buf_addr argument to the panda_callbacks_replay_handle_packet function.  It used to be a uint64_t, but now it's a target_ptr_t.  When building i386 on a 64-bit host, target_ptr_t is uint32_t.  If you look at rr_log.c line 833, it appears old_buf_addr is being set to a host address, which on a 64-bit host is a 64-bit number.  But because the old_buf_addr is now 32-bits in all those functions it is passed too (and C/C++ is a really lazy language when it comes to type checking), we quietly lose the top 4 bytes of old_buf_addr.
As old_buf_addr is valuable when trying to coordinate packet processing during replays, I really think we need to keep the original value, and revert to having old_buf_addr be the size of a HOST pointer.

nathanjackson:
Indeed. The `old_buf_addr` type is `uint64_t` in the non-deterministic log.

https://github.com/panda-re/panda/blob/master/panda/include/panda/rr/rr_log.h#L102

This callback needs to be changed back.

m000:
Indeed I may have gotten this wrong. The argument is described in the callbacks definition header as `XXX this is a mystery`, so I went with the wrong guess, thinking the argument was referring to a guest address.

Since this is cast to `uintptr_t` in `rr_log.c`, using that instead of `target_ptr_t` _should_ work. I'll update #501 in a minute to also include a fix for this.

m000:
Looking closer, this is a bit more complicated. In the rare usage scenario where someone attempts to replay on a 32bit host a trace captured on a 64bit host, the cast on `rr_log.c:833` will result in truncation. So, there's that.

m000:
@LauraLMann Can you help me clarify the use of `old_buf_addr`? I'm getting conflicting input from the code.

- As you mention, in `rr_log.c:833`, the value is loaded from the rr log as if it refers to a host address.
- In `manual.md:1898`, `old_buf_addr` is described as "the address that `buf` had when the recording was taken". So again it sounds like a host address.
- But in tainted net plugin (`tainted_net.cpp:122`), the address is used to apply taint. I.e. it is used as a guest address.

Based on the use, I'd say that `target_ptr_t` is the correct type. I'm not sure what kind of loss we have for the 4 top bytes of `old_buf_add`. For a 32bit guest, these should be 0 anyway.

nathanjackson:
[...]

PR #505, opened 2019-11-07
Fix for PANDA_CB_REPLAY_HANDLE_PACKET callback.

m000:
Addresses #504.

Argument `old_buf_addr` corresponds to a host address, captured at the time of recording. Therefore, it needs to be `uint64_t`, rather than `target_ptr_t`. The `uint64_t` type is large enough to fit host addresses of both 32bit and 64bit hosts.

Also, `old_buf_addr` is not a very intuitive name. It was changed to `buf_addr_rec`, which reflects the semantics of the argument better.

In the process, it was discovered that PANDA dumps/loads raw pointers to the trace. This means that traces captured by 32bit builds will not be replayable by 64bit builds. And vice-versa. The top-level `README.md` has been updated to warn about this issue.

Some documentation housekeeping was also performed. `PANDA.md` has been removed as it duplicated information already present in `manual.md`. Also, detailed build instructions have been spun off from `README.md` to separate files in the `docs` directory.

Finally, a condition was fixed in `rr_log.c` that resulted in not having a memory snapshot written when recording starts.


m000:
The fix for PANDA_CB_REPLAY_HANDLE_PACKET should be good (it merely reverts some recent changes).
So, unless there's some feedback for the updates on the documentation, this can be merged.

PR #485, opened 2019-10-08
Add IDAPython plugin to color tainted pseudocode lines

LauraLMann:
This IDAPython plugin works with the output of PANDA's ida_taint2 plugin.  It has only been tested with i386, but it seems useful as is.

PR #507, opened 2019-11-13
Add IDAPython plugin to color tainted pseudocode lines

LauraLMann:
Adds an IDAPython plugin that works with PANDA's ida_taint2 plugin output to colorize tainted lines in the pseudocode window.  It has only been tested with i386, but has been found to be quite useful as is.
This is a replacement for pull request #485, in which numerous conflicts that could not be automatically resolved were introduced into files I DID NOT TOUCH by the acceptance of other pull requests.  Manually replaying my changes on the new master seemed easier than undoing the merge mess.


nathanjackson:
Any chance we could get this merged soon? It doesn't really affect core PANDA.

PR #509, opened 2019-11-14
Add 64-bit linux support to syscalls2 plugin

LauraLMann:
There are some restrictions:
only 64-bit linux was added (not Windows)
system calls marked "ptregs" are not processed by the parsing scripts and so not supported
system calls marked x32 specific (which have IDs up in the 500s) are not supported as their signatures couldn't be found
32-bit system calls (ie. via "int 0x80" or "sysenter") encountered while the 64-bit profile is active are not processed - instead a warning is printed out
These changes have only been extensively tested with the file_taint plugin, and a few little problems in that plugin that got in the way of testing were fixed along the way.
Adjustments were made in the loaded plugin so at least it will build and still support the 32-bit guests it used to support, but it was not modified to support 64-bit linux.
Also, 64-bit linux support was enabled in the callstack_instr plugin for the threaded stack type (which was useful for testing purposes).
This pull request replaces #495, which had numerous conflicts in files I didn't touch for this work.  It seemed safer to manually apply my changes on top of the latest master than to attempt to manually resolve each conflict arising from merging the latest master into my old branch (fewer opportunities for typos).


nathanjackson:
@moyix, any chance we could get this one reviewed? Its predecessor #495 sat idle for a few weeks.

moyix:
Thanks for redoing the PR! Merging.

PR #512, opened 2019-11-22
Fix more network callback issues introduced recently

LauraLMann:
The changes for pull request #505 were incomplete, as the buf_addr_rec passed into the handle packets callback in the tainted_net plugin was used in various contexts that also needed to revert their types to uint64_t to avoid truncating host addresses.
Also, the PANDA_CB_REPLAY_NET_TRANSFER callback had a similar problem (to what PANDA_REPLAY_HANDLE_PACKET originally had) with the source and destination addresses sometimes being for a host (thus 64-bit on a 64-bit host) address and sometimes being for a guest address, and sometimes both being a host address.  This caused host addresses to get truncated, and so taint did not go where it was supposed to.  So, I fixed that.
Could you please take a look @m000, after Travis has blessed it (and hopefully add your blessing as well).

### DIFF
--- a/panda/include/panda/callbacks/cb-defs.h
+++ b/panda/include/panda/callbacks/cb-defs.h
@@ -689,8 +689,8 @@ typedef union panda_cb {
        Arguments:
         CPUState *env:          pointer to CPUState
         uint32_t type:          type of transfer  (Net_transfer_type)
-        target_ptr_t src_addr:  address for src
-        target_ptr_t dest_addr: address for dest
+        uint64_t src_addr:      address for src
+        uint64_t dest_addr:     address for dest
         size_t num_bytes:       size of transfer in bytes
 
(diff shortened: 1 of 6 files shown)


LauraLMann:
One more time...

LauraLMann:
Travis is unhappy "retrieving gpg key timed out" while trying 'sudo -E apt-add-repository -y "ppa:phulin/panda"'.  I don't think that's something I have rights to fix.  If I do, please educate me.
@moyix, or @m000, or @jnmartin84 - anybody have any advice on resolving this Travis issue?

moyix:
Sounds like a network issue on the travis side. I'll kick it.

moyix:
Looks good, thanks for the fix!

LauraLMann:
Thanks so much for the quick response @moyix!

PR #519, opened 2020-01-07
pandalog_taint_query free not freeing

Kojoley:
Because of it there are plenty of memory leaks in plugins

### DIFF
--- a/panda/plugins/taint2/taint_api.cpp
+++ b/panda/plugins/taint2/taint_api.cpp
@@ -471,6 +471,7 @@ void pandalog_taint_query_free(Panda__TaintQuery *tq) {
         }
         free(tq->unique_label_set);
     }
+    free(tq);
 }
 
 extern bool taintEnabled;


AndrewFasano:
Thanks for all these PRs!

When reviewing this one, I noticed there might be another memory leak pri_dwarf. Shouldn't each element in the array returned from `on_get_libraries` also be freed at some point? Not sure if you'd be interested in tracking that down as well or if we should have somebody else should look into it.

Kojoley:
> When reviewing this one, I noticed there might be another memory leak pri_dwarf. Shouldn't each element in the array returned from on_get_libraries also be freed at some point? Not sure if you'd be interested in tracking that down as well or if we should have somebody else should look into it.

Most likely, I have seen it but did not upstream the fix because I have not received response on already opened PRs. Also, there are some memory leaks in QEmu and they seem to not beeing fixed even in upstream, should I file to both repositories or only to QEmu?

AndrewFasano:
Sorry for the slow responses. We're all a bit busy right now, but we still try to review the PRs every couple of weeks.

For QEmu bugs, you can just report them upstream and we'll eventually get their fixes merged in. But if you're using PANDA and they're causing issues for you (or if they're particularly bad), you could also report them here.

PR #526, opened 2020-01-10
Github Actions

AndrewFasano:
Use Github Actions to:
1) Build new docker containers for **new commit to master**. Creates a container from Ubuntu 18.04 and 14.04. Makes containers available at https://github.com/panda-re/panda/packages, tagged with the commit hash.

2) Validate that commits are building by doing an incremental build **for all pushes**: starting from the latest pre-built docker container from master, checkout the commit to be tested, and just run make. This avoids reinstalling dependencies and remaking unmodified parts of PANDA.

3) Validate that **all pull-requests** are building by running a full build of PANDA from scratch.

Includes some commits from upstream to ensure we can build with gcc7. Disables some tests related to crypto that are failing when we run `make check`.

Note that these checks don't run any sort of test suite, they just try to build. We'll leave the testing to Jenkins in the future.


AndrewFasano:
Having some issues, this isn't quite ready to merge

AndrewFasano:
I think this is ready to be merged now

PR #529, opened 2020-01-23
Avatar panda

mariusmue:
This PR bundles the [avatar-](https://github.com/avatartwo/avatar2)-additions to PANDA, rebased on latest master of PANDA (1a4d9c5a436e0a1ce6ec563f992f0f9e9cee1424).

More specifically, it brings (among others) the following features to panda-system-arm:
* configurable machine: a board which can be flexibly configured via JSON.
* avatar-remote-memory: a qemu-peripheral which allows forwarding of MMIO accesses to the avatar2-framework via message-queues.
* pyperipheral2: Another peripheral which takes a python2 file as input for prototyping peripherals, following the[ avatar2-peripheral](https://github.com/avatartwo/avatar2/blob/master/avatar2/peripherals/avatar_peripheral.py) API

Most of the changes are selfcontained inside a hw/avatar directory.

The included commits can be squashed, the detailed history can be seen in the [avatar-panda ](https://github.com/avatartwo/avatar-panda) repository, which will be deprecated once this PR is mainlined. On the note of stability, the additions of this PR successfully passed the tests included in avatar's CI, but haven't been tested beyond this.

I'm happy to maintain the according subsystem in case it finds it ways to upstream!

Cheers,
Marius


tnballo:
Thanks @mariusmue! Really appreciate these additions. Couple of questions from taking a quick look:

- Can you provide an example JSON that the configurable machine would ingest? (Maybe there is one in the avatar repo?)
- If the avatar-panda repository will be depreciated, are there tests from avatar's CI we could use to make sure that future updates within `hw/avatar` are OK? The configurable machine is something I'd be interested in using and possibly modifying, so I'd be nice to ensure any future modifications don't break things for avatar users.
- `hw/avatar/python_peripheral.c` - does the input need to be python2 as opposed to python3? Latter would be preferred :P
- `hw/avatar/configurable_machine.c ` - What is the purpose  of `dummy_interrupt` that all QEMU devices are wired to? For PANDA usecases of this configurable machine, we'll probably want to create QEMU supported peripherals, map their MMIO at some specific addresses, and then wire IRQs appropriately with `sysbus_connect_irq` - is that something the configuration JSON would allow?

mariusmue:
Thanks for the response, I'm also looking forward in seeing those changes mainlined instead of having yearly rebases on panda-latest. :)

Let me answer the questions in order:
1) In the current use-case of avatar, panda is not run standalone - the json is always dynamically generated by avatar. I can for sure dump one of those JSON's and add them
2) The [CI of the avatar2-python core](https://travis-ci.org/avatartwo/avatar2) checks out avatar-panda (and it future panda) builds it, and runs it's test against it. The corresponding test can be found [here](https://github.com/avatartwo/avatar2/blob/master/tests/test_qemutarget.py). Does the PANDA-CI supports pip and nosetests? If so, we could "invert" the test and have the PANDA CI checking out avatar from pypi, which should be stable releases, and use this test for verification. We could probably also come out with a test decoupled from avatar, though.
3) pyperipheral2 needs python2. I have a fork of avatar-panda which has support for pyperipheral3 (which uses python3), and a switch in the configure-file to select which one to build (if at all). I'll see to have it integrated into this PR.
4) The IRQ subsystem with the configurable machine is limited. In the long run, I think the configurable machine will need a refactor, in which we could also account for better interrupt support. However, I thought it's better to have the base functionality merged first.

tnballo:
Closing and re-opening this PR to kick off Jenkins CI.

tnballo:
Jenkins comment was a little confusing, FYI this PR passed all tests in `./panda/testing`.

jnmartin84:
I can take a look tomorrow and see what happened during the build

PR #530, opened 2020-01-23
Network plugin tweaks to get it to compile with Wireshark 3.2.2

jmcarter9t:
Added some changes to accommodate the deprecation of `wiretap_dump_open_ng` in Wireshark version 3.0 (it still existed prior).  Also changed the preprocessor conditionals to include the major version change.  I assumed the output file from the plugin should NOT be compressed -- this is now a hard-coded option.

**I did NOT test the plugin because I am still getting up to speed on plugins.**


AndrewFasano:
Closing and reopening to try kicking the CI scripts

AndrewFasano:
Despite the Github CI being a bit messed up, it looks like it did find an actual issue when it finally ran [here](https://github.com/panda-re/panda/pull/530/checks?check_run_id=496648211):

```
  CXX     i386-softmmu/panda/plugins/network/network.o
In file included from /panda/panda/plugins/network/network.cpp:22:0:
/usr/include/wireshark/config.h:428:10: fatal error: ws_diag_control.h: No such file or directory
 #include <ws_diag_control.h>
          ^~~~~~~~~~~~~~~~~~~
```

Does the install_ubuntu.sh script install wireshark but not whatever package provides this file? Or are the includes paths incorrect?

AndrewFasano:
I think this PR might also contain some of the code for #564. The network plugin tweaks are probably worth pulling in, but I think we should remove the OSI changes first

AndrewFasano:
I think this was fixed in #605 which started with this PR and then made some other changes. Feel free to reopen if there's anything that + #564 don't capture

PR #533, opened 2020-01-29
Adjust RPATH in .so files during "make install"

LauraLMann:
Certain plugins cannot be loaded when using a build of PANDA created using "make install".  This is because in some situations the RPATH in an .so file is used to load plugins, but the RPATH does not refer to where the .so files for the plugins were installed, but instead to where they were built.  The "chrpath" utility is used during "build install" to adjust the RPATH in the .so files that are placed in the deb package.
Of course, this means you need to have chrpath installed on your machine if you ever use "make install".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ addons:
       #- sourceline: 'deb http://ppa.launchpad.net/phulin/panda/ubuntu xenial main'
     packages:
       # Build dependencies
+      - chrpath
       - libaio-dev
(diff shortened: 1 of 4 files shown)


LauraLMann:
As the error message above says to ignore the test suite results, I was doing so.  But as there has been no movement on this pull request, I now wonder - is there a real problem I need to take care of?  If so, please inform me.  The Travis results are just that it timed out, which usually is not indicative of a real error.

PR #535, opened 2020-01-30
Pypanda/flex installer

mariusmue:
Currently, the setup.py for pypanda strictly expects PANDA to be built with build.sh, and LLVM support enabled.
This is not always the case (e.g., in the avatar-context). Hence, this little PR allows for more flexibility on side of the installer.


mariusmue:
Please do not merge now, I found a bug related to the plog2 imports

PR #536, opened 2020-01-30
MMIO callback and plugin updates

tnballo:
* Cherry picks 3a0b1c31c22979 from `pypanda-dev` (updates `PANDA_CB_MMIO_AFTER_READ` so that plugins can modify values read)
* Updates `PANDA_CB_MMIO_AFTER_WRITE` so that plugins can modify values written. This keeps the signatures of MMIO read and write callbacks consistent, and adds some additional flexibility
* Adds virtual address to both MMIO read and write callbacks (this came up in testing a specific firmware)
* Updates `mmio-trace` plugin to log to JSON, with every access annotated with the name of the MMIO device being accessed

### DIFF
--- a/cputlb.c
+++ b/cputlb.c
@@ -832,6 +832,8 @@ static void io_writex(CPUArchState *env, CPUIOTLBEntry *iotlbentry,
     cpu->mem_io_vaddr = addr;
     cpu->mem_io_pc = retaddr;
 
+    panda_callbacks_mmio_before_write(cpu, physaddr, addr, size, &val);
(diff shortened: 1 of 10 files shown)


tnballo:
Don't merge yet, `PANDA_CB_MMIO_AFTER_WRITE` needs to be changed to `PANDA_CB_MMIO_BEFORE_WRITE`, otherwise any potential mutation of the value written will have no effect.

tnballo:
This is now ready to merge.

PR #542, opened 2020-02-05
Fix multiple memory leaks

Kojoley:


AndrewFasano:
Thanks for submitting these additional fixes! It looks like one of the changes is now causing the build to fail in Travis. Do you think you could figure out where things went wrong?
Thanks!

Kojoley:
Ahh, I was compiling with Clang and did not see them, however, `maybe-uninitialized` is known to produce false-positives, and since I run with `-fsanitize=address` it is most likely a false-positive.

What about `discarded-qualifiers` warnings. May I change `panda_set_os_name(char *)` to `panda_set_os_name(const char *)`? And suppress `g_free` warnings with `g_free((gchar *)p)` (it is GCC/Glib problem and they suppress it inside Glib too https://github.com/GNOME/glib/blob/d44e00fb98a8d247af7d20d2e1123b8f8425cdfa/glib/gutils.c#L397)

Kojoley:
I moved offending change to a separate PR #544

AndrewFasano:
@nathanjackson ran a private test suite he has and this PR caused some crashes.

Nathan, can you post any of the details here? Even just a backtrace would be great.

Kojoley:
The only thing I can do without any info about the crash is is to split the PR, ping me if you wish me to do so.

PR #544, opened 2020-02-06
panda_plugin_names and panda_plugin_files content leakage

Kojoley:


Kojoley:
I have no idea what GCC is trying to say here, most likely it is another `maybe-uninitialized` GCC bug.

AndrewFasano:
I see how that's a stupid error now. I'd like to try to find a way to fix it so we can keep using our CI though. I can take a look into this in the next week or so

Kojoley:
Well, GCC seems to be right on this one, the problem is that
https://github.com/panda-re/panda/blob/3e6fc72f3456ec47f50e5b0c9957178d078cc98a/vl.c#L3095-L3098
that skips the initialization and I had not recognized it previously.

PR #557, opened 2020-03-13
Edge coverage & loaded libs plugins 

hpreslier:


panda/plugins/edge_coverage/edge_coverage.cpp
```diff
@@ -0,0 +1,153 @@
+// Generate a drcov trace file from a panda recording
+// TODO: the output isn't quite right, Lighthouse shows no coverage
```
AndrewFasano: Is this at all true anymore?

hpreslier: Nope. Forgot to delete that.

hpreslier: It’s all pandalog.

Issue #548, opened 2020-02-10
Documentation out of date. Can't run osi_linux example.


janbbeck:
I am following the documentation to run osi_linux as given here:
[https://github.com/panda-re/panda/tree/master/panda/plugins/osi_linux](url)

For testing I spun up a ubuntu-18.04.3-desktop-amd64.iso image in live modeas a guest:

`./panda/build/x86_64-softmmu/panda-system-x86_64 -m 3G -cdrom 'ubuntu-18.04.3-desktop-amd64.iso' -monitor stdio`

Recording works fine, and so does the replay with replaymovie. Next, I wanted to try the osi_linux as per the instructions above. I created a new kernelconf.info entry as described in the readme:

```
[my_kernel_info]
name = 5.0.0-23-generic|#24~18.04.1-Ubuntu SMP Mon Jul 29 16:12:28 UTC 2019|x86_64
version.a = 5
version.b = 0
version.c = 15
task.per_cpu_offsets_addr = 18446744071846983712
task.per_cpu_offset_0_addr = 18446616393938894848
task.current_task_addr = 89088
task.init_addr = 18446744071849473856
#task.per_cpu_offsets_addr = FFFFFFFF90FB7820
#task.per_cpu_offset_0_addr = 0xFFFF8BE03CA00000
#task.current_task_addr = 0x00015C00
#task.init_addr = 0xFFFFFFFF91217740
task.size = 9152
task.tasks_offset = 1960
task.pid_offset = 2216
task.tgid_offset = 2220
task.group_leader_offset = 2280
task.thread_group_offset = 2392
task.real_parent_offset = 2232
task.parent_offset = 2240
task.mm_offset = 2040
task.stack_offset = 24
task.real_cred_offset = 2624
task.cred_offset = 2632
task.comm_offset = 2640
task.comm_size = 16
task.files_offset = 2712
cred.uid_offset = 4
cred.gid_offset = 8
cred.euid_offset = 20
cred.egid_offset = 24
mm.size = 1032
mm.mmap_offset = 0
mm.pgd_offset = 80
mm.arg_start_offset = 304
mm.start_brk_offset = 280
mm.brk_offset = 288
mm.start_stack_offset = 296
vma.size = 208
vma.vm_mm_offset = 64
vma.vm_start_offset = 0
vma.vm_end_offset = 8
vma.vm_next_offset = 16
vma.vm_flags_offset = 80
vma.vm_file_offset = 160
fs.f_path_dentry_offset = 24
[...]

PR #559, opened 2020-03-15
Fix documentation (Issue #548) and add kernels to kernelinfo.conf

janbbeck:
The documentation detailing the osi_linux plugin does not work for me. See Issue #548 for discussion.
I did add an extra example to the plugin documentation that works on my setup (clean Ubuntu 18.04 install). I also added information on how to disable KASLR which is a headache when using newer kernels. Finally, I added kernel configuration parameters for the kernels that ship with Ubuntu 18.04.3 and 18.04.4.

Most of the documentation seems terse, so I kept the extra example very short to fit in with the PANDA style. However, if more documentation,including screen captures is desired, I put that up here: 
https://www.janbeck.com/cybersecurity-challenges-ctfs-and-more/panda-whole-system-tracing
Feel free to pull anything relevant from there.


AndrewFasano:
Thanks for the PR!

PR #566, opened 2020-03-18
Adding support for on_asid_change

### LINKED ISSUE #498: PANDA_CB_ASID_CHANGED implemented only for x86
While working on #497, I noticed that the `panda_callbacks_asid_changed()` helper function is only invoked from `target/i386/helper.c`. This means that apparently the `PANDA_CB_ASID_CHANGED` will only work for x86. 

I'm opening this issue as a reminder. The code for this can probably be found in the PANDA-1 codebase. It should not be hard to port to PANDA-2, but I can't pick this up now.

lacraig2:
Adding support for the callback `asid_change` on multiple architectures. I'll be satisfied with adding ARM and PPC. This commit adds ARM support so far. CLOSES #498.


AndrewFasano:
Do we also need to update some documentation?

PR #569, opened 2020-03-19
Add AFTER_VMLOAD callback and make OSI Linux reinitialize after vmload

AndrewFasano:
Should merge 567 prior to this.


panda/pypanda/panda/include/panda_datatypes.h
```diff
@@ -603,14 +604,14 @@ typedef union panda_cb {
...
-        0 otherwise
+        true if the asid should be prevented from being changed
+        false otherwise
 
        Notes:
         The helper is only invoked for x86. This should break a lot of the
         plugins which rely on this callback to detect context switches.
     */
-    int (*asid_changed)(CPUState *env, target_ptr_t oldval, target_ptr_t newval);
+    bool (*asid_changed)(CPUState *env, target_ptr_t oldval, target_ptr_t newval);
```
AndrewFasano: This change really should've been in 4b8ff5c96705f47a535246a15ffd642593ba05d1 but it's an autogenerated file so it's no big deal

PR #572, opened 2020-03-22
recctrl plugin: initial commit

m000:
This plugin allows controlling start/stop of recording from within the VM. See the plugin README file for details.


m000:
Seems I'm missing the implementation of panda_vm_quit(). This was probably saved only in local commits on the hard disk that was stolen from my computer earlier this week. I'll check for any backups, or I'll re-implement it and add it here.

m000:
This should be good to merge.

PR #575, opened 2020-03-24
merge pypanda-dev into master

lacraig2:
Just a bunch of smaller updates and examples.


AndrewFasano:
We tried to make this PR cleaner in #576. It should still have all the same code changes but a much smaller history.

PR #576, opened 2020-03-24
Cleaner: merge pypanda-dev into master

AndrewFasano:
This is just PR 575 by @lacraig2 but rebased to avoid duplicate commits (that were squashed when merged into master previously in PR 571).

Prior description was:
> Just a bunch of smaller updates and examples.


lacraig2:
That's helpful. Much cleaner. Thanks.

PR #578, opened 2020-03-24
MIPS 32-bit support

lacraig2:
Adding MIPS 32-bit support to PANDA. Several things changed

- Changed build.sh to add mips-softmmu
- Changed internal PANDA functions to support MIPS architecture
- Changed other plugins to support MIPS and/or not fail to build

Still need to do:

- Update test suite
- Additional testing


lacraig2:
Running into an issue where the machine stalls after getting through to `start_kernel`. Best to not merge just yet.

lacraig2:
I've fixed this issue here and userspace in multiple applications works flawlessly. Next up is adding python support. Best to not merge yet.

jnmartin84:
I'll disable that pypanda check

jnmartin84:
You might want to bounce your PR to kick Jenkins again and get to passing state

lacraig2:
Still failing in replay. Some non-determinism issues.

target/mips/op_helper.c
```diff
@@ -696,212 +700,566 @@ static void sync_c0_entryhi(CPUMIPSState *cpu, int tc)
...
 
 target_ulong helper_mfc0_count(CPUMIPSState *env)
 {
-    int32_t count;
-    qemu_mutex_lock_iothread();
-    count = (int32_t) cpu_mips_get_count(env);
-    qemu_mutex_unlock_iothread();
-    return count;
+    target_ulong count;
+    //qemu_mutex_lock_iothread();
```
AndrewFasano, line 1058: I'm scared of disabling this lock and subsequent unlock- mutex locks seem like something we shouldn't be messing with

AndrewFasano: Looks like that change is in a few of the op_helper functions - I had a bad time messing with mutexes for the original pypanda merge (ultimately I undid all my mutex-related changes) and I'd be hesitant about tweaking this logic since I assume the mutexes were added for a reason in upstream

lacraig2: We talked about this a long time ago. These disappear later on in upstream qemu. They aren't necessary.

AndrewFasano: Yeah, I remember that and agree

lacraig2:
Planning to merge this without record/replay support.

PR #591, opened 2020-04-03
Stability fixes for file_faker

AndrewFasano:
Change FakeFile->HyperFD mapping to more closely resemble how files and file descriptors are actually related.

Handle the case where ASID changes between entering and returning from a syscall.


AndrewFasano:
These changes make the plugin more stable but it looks like it still fails to identify filenames more than I'd like. I'll keep working on this branch and merge later once it's fixed better.

AndrewFasano:
I think most of this has been re-implemented in master, but there might be some stuff still worth pulling in from here. Whoops.

PR #596, opened 2020-04-07
Correctly set pid_t type for linux system calls.

m000:
`pid_t` was incorrectly mapped to an unsigned type. The syscall prototype parser has been updated to fix that. Related files have been regenerated.


m000:
@AndrewFasano this should be good to merge. Changes only switch from an unsigned to a signed type and add some missing callbacks previously present only in x64.

PR #600, opened 2020-04-08
improvements in configure/build system

m000:
- Proper LLVM detection in configure script.
- Proper probing for protobufs in configure script.
- Capstone support (ported from QEMU upstream).
- Dropped support for some legacy platforms as hosts (ported from QEMU upstream).


AndrewFasano:
This looks fantastic, thanks for this maintenance-heavy PR!

m000:
These were side-changes I made while trying to make PANDA compile/run on MacOS.

But that branch became too cluttered with side-changes, so I spun them off separately. Will now try to rebase the MacOS branch and see if it looks any better.

AndrewFasano:
This PR appears to have broken `-d in_asm`. Previously it would output disassembly, but after this we just get the raw bytes:
```
----------------
IN:
0xfffffff0:
OBJD-T: ea5be000f0
```

It's probably related to some config flag and/or capstone. I'll try to track it down and fix it.

PR #603, opened 2020-04-09
Docker container handling PANDA/QEMU symbol generation for python

lacraig2:
Previously the PANDA/QEMU symbols imported into `cffi` for python were generated through an awful python script I wrote. This doesn't change that. 

However, this PR makes that ugliness work consistently though docker. This is helpful because it is remarkably difficult get set up properly on its own.

This PR adds the following files:
- `Dockerfile` - This sets up the container, requirements, and copies scripts in.
- `build.sh` - This sets up the docker container from the `Dockerfile`.
- `run.sh` - This runs the script and outputs things into the `include` directory under `python` for use with `cffi`.

I do not expect other people to *want* to build this, but I would like others to be able to build these symbols.


lacraig2:
Simple. Touches only files I have personally edited. Merging myself.

AndrewFasano:
Do you think it would be worth splitting these into a separate repo? Seems like it's not necessarily something we need to ship with PANDA and it's kind of pretty cool as a project on it's own.

Issue #595, opened 2020-04-06
Support for clean build on Ubuntu 20.04


janbbeck:
The next LTS release of Ubuntu is just around the corner, and I think we should work to make sure panda builds on that out of the box. Right now, I think one has to suppress `Werror ` at the least.

I'm spinning up the latest daily of 20.04 to see what needs to be done, and I am hoping that we can discuss any issues that come up here.

First question: When running the `install_ubuntu.sh` script in `panda/panda/scripts` it downloads a complete second copy of panda. Is that intentional?

AndrewFasano:
I think when that script was made the ideas was that it could be downloaded and run on a Ubuntu system and just install panda. But when panda's already cloned, it definitely shouldn't clone it again.

janbbeck:
> I think when that script was made the ideas was that it could be downloaded and run on a Ubuntu system and just install panda. But when panda's already cloned, it definitely shouldn't clone it again.

Thanks. I really don't want to mess with that file as it seems to support several OS versions and I don't want to test against all of them. Can I just add a new script, `install_prerequisites_in_Ubuntu_20.04` , or something?

janbbeck:
Update:
I have panda booting a live DVD on 20.04 now with just a little trickery.

Currently, in addition to disabling all werror, this requires disabling the `network` plugin, as that won't build with any compiler I tried. 

The network plugin fails with 4 errors:
```
/media/ubuntu/tmp/panda/panda/plugins/network/network.cpp: In function ‘bool init_plugin(void*)’:
/media/ubuntu/tmp/panda/panda/plugins/network/network.cpp:64:18: error: ‘wtap_dump_open_ng’ was not declared in this scope; did you mean ‘wtap_dump_open’?
   64 |     plugin_log = wtap_dump_open_ng(
      |                  ^~~~~~~~~~~~~~~~~
      |                  wtap_dump_open
/media/ubuntu/tmp/panda/panda/plugins/network/network.cpp: In function ‘void handle_packet(CPUState*, uint8_t*, size_t, uint8_t, uint64_t)’:
/media/ubuntu/tmp/panda/panda/plugins/network/network.cpp:129:24: error: aggregate ‘handle_packet(CPUState*, uint8_t*, size_t, uint8_t, uint64_t)::wtap_pkthdr header’ has incomplete type and cannot be defined
  129 |     struct wtap_pkthdr header;
      |                        ^~~~~~
[...]

PR #605, opened 2020-04-12
fix network plugin for support of ubuntu 19.10

janbbeck:
Tested on 16.04, 18.04, 19.04 and 19.10 by replaying the same recording on each system.

In each case I replayed the system into a .pcap file with and without the patch. Except for 19.10, of course, which will not build without the patch.
In each case I translated the pcap file (because of binary differences between different wireshark versions, the actual pcap files are always different between versions, even without the patch, I think) using tshark:
``` tshark -r foo-stock1904.pcap -Tfields -e ip.src -e ip.dst -e tcp.srcport -e tcp.dstport  > cap1.txt ```
I then checked the resulting text files against each other and they all agree.

I took code from #530 which needed work, see #595

### DIFF
--- a/panda/plugins/network/network.cpp
+++ b/panda/plugins/network/network.cpp
@@ -36,10 +36,25 @@ wtap_dumper *plugin_log;
 
 bool init_plugin(void *self) {
     panda_cb pcb;
-
+    int err;
     int i;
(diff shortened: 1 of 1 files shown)


janbbeck:
Update: This also works on 20.04

PR #606, opened 2020-04-13
Ubuntu install script support for 20.04

janbbeck:
Tested to 'work' on 16.04 18.04 19.04 and 20.04 (assuming #605 is applied).

'work' means successful installation/compilation and boot of a 20.04 live DVD as guest.


AndrewFasano:
Just merged in the changes from master, mostly your code from #601 - looks like you had checked the version in two conflicting, ways, using `lsb_release` with `-r` and `-d`. I kept the `-r` version - let me know if that's incorrect. Otherwise I'll merge once the CI finishes

janbbeck:
> Just merged in the changes from master, mostly your code from #601 - looks like you had checked the version in two conflicting, ways, using `lsb_release` with `-r` and `-d`. I kept the `-r` version - let me know if that's incorrect. Otherwise I'll merge once the CI finishes

Thanks!
-r was correct. This way, even the beta works, not just the final release.

PR #610, opened 2020-04-14
Set GIT_VERSION and BUILD_DATE in preprocessor during build

AndrewFasano:


AndrewFasano:
Note this PR was a dumb idea, at least one of these variables is already set by qemu

PR #616, opened 2020-04-16
restoring a snapshot now raises an exception when there is a memory mismatch

bigandian:


AndrewFasano:
We should probably be checking for errors in all of our monitor interactions (e.g., when we try to save/restore snapshots). Maybe we should switch to using QMP instead of the monitor so it would be easier to identify failures.

This is a good start to making these errors more manageable, thanks!

PR #618, opened 2020-04-17
kernelinfo_gdb typo

tnballo:
Fix tiny typo in `kernelinfo_gdb`, so plugin can load the generated config.


lacraig2:
This is correct. Thanks for catching this @tnballo.

Issue #506, opened 2019-11-10
osi_linux: cannot generate kernelinfo.conf for ARM target


kasperis7:
I was generating kernelinfo.conf for a Debian 10 kernel of armvirt target.
output:
```
root@debian:~/kernelinfo# make
make -C /lib/modules/4.19.0-5-armmp-lpae/build M=/root/kernelinfo modules
make[1]: Entering directory '/usr/src/linux-headers-4.19.0-5-armmp-lpae'
  CC [M]  /root/kernelinfo/kernelinfo.o
/root/kernelinfo/kernelinfo.c: In function ‘init_module’:
/root/kernelinfo/kernelinfo.c:158:71: error: ‘current_task’ undeclared (first use in this function); did you mean ‘current_umask’?
 rintk(KERN_INFO "task.current_task_addr = %llu\n", (u64)(uintptr_t)&current_task);
                                                                     ^~~~~~~~~~~~
                                                                     current_umask
/root/kernelinfo/kernelinfo.c:158:71: note: each undeclared identifier is reported only once for each function it appears in
make[4]: *** [/usr/src/linux-headers-4.19.0-5-common/scripts/Makefile.build:315: /root/kernelinfo/kernelinfo.o] Error 1
make[3]: *** [/usr/src/linux-headers-4.19.0-5-common/Makefile:1539: _module_/root/kernelinfo] Error 2
make[2]: *** [Makefile:146: sub-make] Error 2
make[1]: *** [Makefile:8: all] Error 2
make[1]: Leaving directory '/usr/src/linux-headers-4.19.0-5-armmp-lpae'
make: *** [Makefile:4: all] Error 2
```
kernel information:
```
root@debian:~/kernelinfo# uname -a
Linux debian 4.19.0-5-armmp-lpae #1 SMP Debian 4.19.37-5+deb10u1 (2019-07-19) armv7l GNU/Linux
```
`current_task` is an export symbol in /arch/x86/kernel/cpu/common.c. As for arm target, this symbol doesn't exist.

m000:
Yes, the location of the `current_task` seems to be arch-specific. So we need to adapt the plugin for this. If you are familiar with ARM, can you come up with a PR? Maybe check how Volatility3 gets this information.

jmcarter9t:
[...]

PR #620, opened 2020-04-17
Updates to in-guest OSI profile generation for ARM

AndrewFasano:
This is @jmcarter9t's PR #506 but with some unrelated code changes removed.

@m000 had some thoughts about removing the `archtype` field - is that a change we should make before merging?


AndrewFasano:
The GDB-based approach to generating OSI makes up a different value for `current_task` and I'm not sure which is correct - https://github.com/panda-re/panda/blob/4ba8bf38223a76fddc212c328679734046ac8606/panda/plugins/osi_linux/utils/kernelinfo_gdb/extract_kernelinfo.py#L53. Anyone know if it matters?

AndrewFasano:
As described in #651, I think these fixes will need to be a bit more significant so I'm going to close this PR but keep the branch around for when we begin tackling that issue.

AndrewFasano:
I have a better understanding of this code now and think maybe it should actually be merged

AndrewFasano:
I'm about to merge in #768 but wanted to note that it will probably affect this PR - Might need to change something there once osi-profile generation supports multiple architectures because `#define current_task per_cpu__current_task` probably won't work for non-x86.

AndrewFasano:
I believe this is no longer necessary after #868

PR #648, opened 2020-05-04
added initial ppp_remove_cb prototype

### LINKED ISSUE #644: PPP callbacks can't be disabled
The `PPP`-style interface allows one plugin to register a function to run at a callback provided by another plugin, but there's currently no clean way to unregister a function. For standard callbacks, we provide `panda_unregister_callback()`- it would be nice if we could provide something similar for `PPP` callbacks.

lacraig2:
Adding `ppp_remove_cb` to C so that with the function pointer a function can be removed from the list of functions. Also adds support for pypanda to use `panda.disable_ppp('ppp_cb_name')` similar to `panda.disable_callback('cb_name')`

Fixes #644


AndrewFasano:
I just added python support. I think this is ready to merge now. @lacraig2 want to review?

PR #649, opened 2020-05-05
Virtual Memory Read type mismatch

lacraig2:
`vaddr` is always a `uint64_t` according to qemu source. On 32 bit targets `target_ulong` leads to a `uint32_t` and therefore a type mismatch.


lacraig2:
This is how the qemu source does it. Even though it doesn't make sense I'd rather make similar mistakes to the source there.

Issue #652, opened 2020-05-08
Exception Handling in the Python Interface


AndrewFasano:
When python raises an exception in a PANDA callback, the errors and traceback are currently a mess.

Ideally, we should stop execution, then raise the error. There's some code that tries to do this here https://github.com/panda-re/panda/blob/bd7d4f6d6ab89c42a9746b0345a380cb3207bb78/panda/python/core/panda/callback_mixins.py#L45

but it doesn't currently have a way to figure out what return type the C-code is expecting and when it gets that wrong, it makes things even worse.

If we don't do that and just ignore the error, the guest will continue to run indefinitely and the tracebacks will be printed every time the callback runs.

I think the best approach would be to return a junk value of the right type in `_run_and_catch`, but I haven't been able to figure out a clean design for that.

AndrewFasano:
I believe this was fixed by #779.

Issue #638, opened 2020-05-01
OSI Linux can't specify non-default arguments


AndrewFasano:
We can't currently change the OSI path using arguments to OSI Linux because OSI must be loaded first and it automatically loads osi_linux with default arguments if it's a Linux OS.

We want to be able to do something like:
```
panda.load_plugin("osi")
panda.load_plugin("osi_linux", args={"kconf_file": "my_kernelinfo.conf", "kconf_group": "debian:4.9.99:32"})
```
but that doesn't work currently

AndrewFasano:
It seems like the kconf_file and kconf_group arguments are only used by OSI Linux so it would be difficult to pass those directly to OSI.

Instead, I'm thinking of adding an argument to OSI that, when set, disables auto-loading of the os-specific OSI library:

```
panda.load_plugin("osi", {'disable-autoload': True})
panda.load_plugin("osi_linux", args={"kconf_file": "my_kernelinfo.conf", "kconf_group": "debian:4.9.99:32"})
```

Anyone have a better idea?

Issue #651, opened 2020-05-08
OSI Linux- Architecture specific configurations


AndrewFasano:
The kernelinfo.conf files we use for OSI linux contain values specific to x86. We should add a way to specify architecture to these files and then figure out what different values we need to be recording for other architectures.

To start, ARM systems need to know the value of `PAGE_SIZE` in order to find the task struct. That value (typically 2048 or 4096) is not currently set in our configs. We could probably try to figure it out in dynamically, but it would be nice to have in a config. Other fields like `current_task` and `task.per_cpu_offset_0_addr` appear to be meaningless on ARM and should be removed.

AndrewFasano:
`__per_cpu_offset` (and thus `__per_cpu_offset[0]`) appear to not exist for mips

PR #653, opened 2020-05-08
ARM OSI

AndrewFasano:
This PR has a number of fixes that enable OSI on ARM.

Most significantly, this includes a bugfix for panda_virtual_memory_rw which was failing on ARM when trying to read kernel memory. The underlying qemu functions have comments that make them sound like they'll be able to access any guest memory, but on ARM (at least), that wasn't the case and the permission bits in the CPSR register were affecting if data could be read or not. I suspect other architectures may have the same problem so I tried to design a generic interface for switching into privileged mode as necessary.

Other various OSI updates include:
* The ability to load osi_linux with non-default arguments #638
* Changing OSI to initialize after the first syscall which ensures the kernel is setup and we're not trying to introspect on a system during early-boot. This can be disabled via an optional argument.
* When OSI_linux can't find the kgroup you specify, it will now list the available groups in your configuration. I'm still not a fan of these names, but it's slightly better now.
* A few pypanda enhancements for working with OSI and updates to python plugins to use OSI.

Other updates: 
* Add a missing makefile dependency
* Pypanda: Change error handling slightly, but it's still pretty bad (See #652)
* Pypanda: Add an interface to trigger a breakpoint in the guest. It's not as clean as I'd like but it's a start. Future work to connect to a gdbstub immediately from python (without waiting for any new guest code to run) would be really nice.

Edit: this isn't quite ready to merge, seeing some weird behavior with asidstory on arm

### DIFF
--- a/panda/plugins/osi_linux/default_profile.cpp
+++ b/panda/plugins/osi_linux/default_profile.cpp
@@ -1,5 +1,19 @@
 #include "osi_linux.h"
 #include "default_profile.h"
+#ifdef TARGET_ARM
+/**
+ * @brief Returns the current kernel stack pointer for ARM guest
+ */
+target_ptr_t get_ksp (CPUState* cpu) {
+    if ((((CPUARMState*)cpu->env_ptr)->uncached_cpsr & CPSR_M) == ARM_CPU_MODE_SVC) {
+        return ((CPUARMState*)cpu->env_ptr)->regs[13];
+    }else{
+        // Read banked R13 for SVC mode to get the kernel SP (1=>SVC bank from target/arm/internals.h)
+        return ((CPUARMState*)cpu->env_ptr)->banked_r13[1];
+    }
+}
+#endif
+
 
 /**
  * @brief Retrieves the task_struct address using per cpu information.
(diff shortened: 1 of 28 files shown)


panda/plugins/osi_linux/default_profile.cpp
```diff
@@ -1,14 +1,39 @@
...
 {
     struct_get_ret_t err;
+    target_ptr_t current_task_addr;
     target_ptr_t ts;
-    err = struct_get(cpu, &ts, ki.task.current_task_addr, ki.task.per_cpu_offset_0_addr);
+#ifdef TARGET_ARM
+    target_ptr_t kernel_sp = get_ksp(cpu);
+
+    // XXX: This should use THREADINFO_MASK but that's hardcoded and wrong for my test system
+    target_ptr_t task_thread_info = kernel_sp & ~(0x2000 -1);
```
AndrewFasano, line 30: This logic is somewhat concerning. We should fix it with #651

AndrewFasano:
Spent some time debugging the ASID story diff - turns out current master outputs

```
  Count   Pid                  Name      Asid    First         Last
    392  2476               netstat   7a4d000  1464180  ->  7265436
     88  2433                  bash   7ae4000    27333  ->  1341493
     19  2476                 bash2   5202000  1111322  ->  1464074

                bash : [##############                                                             ]
               bash2 : [          #####                                                            ]
             netstat : [              ##########################################################   ]
```

Which is much closer to the output generated by this PR. The diff introduced by the PR just changes the first instruction we identify bash starting at. I'm okay with that diff though.

panda/plugins/osi_linux/osi_linux.cpp
```diff
@@ -591,10 +712,17 @@ bool init_plugin(void *self) {
     PPP_REG_CB("osi", on_get_current_thread, on_get_current_thread);
     PPP_REG_CB("osi", on_get_process_pid, on_get_process_pid);
     PPP_REG_CB("osi", on_get_process_ppid, on_get_process_ppid);
-    LOG_INFO(PLUGIN_NAME " initialization complete.");
+
+    // By default, we'll request syscalls2 to load on first syscall
+    if (!osi_initialized) {
+      panda_require("syscalls2");
```
AndrewFasano, line 718: This dependency should be in the docs - OSI doesn't need syscalls but it will try to use it if there's no OS when OSI is loaded

Issue #387, opened 2018-12-03
`update_cb` is often incorrectly setting `last_literal` in taint_ops.cpp


wangray:
`last_literal` should be the argument of an LLVM instruction that determines the bitmask applied to another operand.

There are two bugs in how `last_literal` is being calculated, which affects how taint is propagated with controlled bits. 

One: In the case of the LLVM instruction `shl i32 1000000000, %tmp1_v9`, last_literal is set to the first constant observed while iterating the arguments, i.e. 1000000000. So the assertion 
` tassert(last_literal < 64);` fails at `update_cb_switch.h:159`. 

I observe this issue regardless of `TAINT2_DEBUG` option — taint does not work at all for me. 

Two: https://github.com/panda-re/panda/issues/371

wangray:
Example replay where I see an assertion failure 

https://www.dropbox.com/s/95qj8arfajxes7v/test-decode-to-global-rr-nondet.log?dl=1
https://www.dropbox.com/s/na499jygzm52qj2/test-decode-to-global-rr-snp?dl=1

nathanjackson:
@wangray are you building against a debug build of LLVM? Something about debug builds messes with the C-library assertion (I assume an LLVM header ends up defining NDEBUG somewhere when in release mode). If you've built against a debug build, that probably explains why you see the assert regardless of the `TAINT2_DEBUG` option.

I've gotten around this by not using a debug build, but that's definitely not a long term solution as sometimes you have to use a debug build for testing.

wangray:
Yes, I'm using a debug build of LLVM. But the assert makes sense, so the bug still affects taint correctness.

Issue #371, opened 2018-10-31
taint2: Assertions when TAINT2_DEBUG is Enabled 


nathanjackson:
The following minimal code produces an assertion failure when TAINT2_DEBUG is set.

```
#include <stdint.h>
#include <stdio.h>

int main()
{
  int32_t src = 12345;
  float dest = -9999.0;

  dest = src;
  if (dest > 1234.5) {
    printf("greater than %f\n", 1234.5);
  }

  return 0;
}
```

Console Output:
```
PANDA[stringsearch]:adding argument name=fild.
PANDA[core]:initializing stringsearch
PANDA[core]:loading required plugin callstack_instr
PANDA[core]:initializing callstack_instr
search strings file [fild_search_strings.txt]
stringsearch: added string of length 4 to search set
PANDA[core]:initializing tstringsearch
PANDA[core]:loading required plugin stringsearch
PANDA[core]:panda_stringsearch.so already loaded
PANDA[core]:loading required plugin taint2
PANDA[core]:initializing taint2
PANDA[taint2]:propagation via pointer dereference ENABLED
PANDA[taint2]:taint operations inlining DISABLED
PANDA[taint2]:llvm optimizations DISABLED
PANDA[taint2]:taint debugging DISABLED
PANDA[taint2]:detaint if control bits 0 DISABLED
PANDA[taint2]:maximum taint compute number (0=unlimited) 0
PANDA[taint2]:maximum taintset cardinality (0=unlimited) 0
PANDA[core]:loading required plugin callstack_instr
PANDA[core]:panda_callstack_instr.so already loaded
loading snapshot
... done.
opening nondet log for read :   ./fild_taint_bug-rr-nondet.log
./fild_taint_bug-rr-nondet.log:   100668978 instrs total.
fild_taint_bug:     1006691 (  1.00%) instrs.    1.83 sec.  1.70 GB ram.
fild_taint_bug:     2013382 (  2.00%) instrs.    3.33 sec.  1.73 GB ram.
fild_taint_bug:     3020071 (  3.00%) instrs.    4.48 sec.  1.75 GB ram.
READ Match of str 0 at: instr_count=3524173 :  82873f89 82846e13 00000000
enabling taint at instr count 3524173
PANDA[taint2]:taint2_enable_taint
taint2: Allocating large fast_shad (34359738368 bytes).
taint2: Allocating small fast_shad (12800000 bytes) using malloc @ 7f09f59c9010.
[...]

PR #655, opened 2020-05-18
64-bit Taint Propagation Improvements

nathanjackson:
Initially, this work started out to support 64-bit multiply. Before, there was a warning message that would indicate that taint was not propagated through a 64-bit multiply due to the extension to 128-bits. However, we found multiple issues with 64-bit taint propagation and those issues are now fixed and I'll try to summarize everything going on here:

- Taint was not propagated through 64-bit multiply since the values were extended to 128-bit by TCG-LLVM. This issue was fixed by passing the low and hi bits of a 128-bit value to `taint_mul_compute`. To preserve the control bit mask calculations, we've updated the underlying data type to `llvm::APInt` which handles arbitrarily sized integers. We initially tried using the `unsigned __int128` type in GCC and Clang, but this was buggy.

- While testing the handling of 128-bit values, we discovered that the `cmpxchg16b` helper used 128-bit types, so we thought that this was a good candidate for testing helpers. It turned out that `cmpxchg16b` was not actually propagating taint due to an unhandled case where a host memory write to guest RAM was not considered in the host copy taint op. We fixed this issue by handling RAM pointers in this taint operation. This enables `cmpxchg16b` and others to propagate taint. With this change, you can expect some differences in your results, but we think they're correct.

- On the control bits side of things, we know there are some deficiencies that still have not been addressed (see: #371, #387). To test the 64-bit taint propagation, I had to implement some workarounds that would otherwise require significant re-work. There are a couple of issues here. First, `last_literal` is assumed to be the amount to shift by in shift instructions, so the assertion fails when built with `TAINT2_DEBUG` and LLVM is built with assertions enabled. However, this isn't the case and it is technically valid to shift more than the width of the value being shifted so I removed the assertion. Instead, we now explicitly handle this case by zeroing the control bits, which should be equivalent to the current behavior when assertions are disabled.

With these changes, you can expect some differences in taint results, but they seem to be reasonable. We've tested x86 and x86-64 fairly extensively and I feel confident in the results for these architectures. I do not have any testing in place for ARM and only the most basic tests for PPC (basically it doesn't crash).


panda/plugins/taint2/tests/update_cb_switch/update_cb_switch.cpp
```diff
@@ -36,28 +40,32 @@
  *    size:  The number of bytes the LLVM instruction operates upon
  *    orig_cb_mask:  The original cb mask for the bytes being operated on
  *    orig_zero_mask:  The original zero mask for the bytes being operated on
+ *    wj/liters
```
AndrewFasano, line 43: Is this a typo?

nathanjackson: Yes. Will fix.

nathanjackson:
It looks like rebasing fixes `make check`, converting to draft while I re-test.

nathanjackson:
@AndrewFasano This one is ready to be merged\another review. My tests pass except for ARM replay (due to the snapshot issue) as discussed.

jnmartin84:
sorry for junking up the "conversation" view of your PR with close/reopens and failures, forgot some dependencies on the Jenkins side

nathanjackson:
@jnmartin84 No Problem, I understand that my PR is a guinea pig :hamster:

jnmartin84:
panda was hung again so I killed the taint test

PR #658, opened 2020-05-20
taint2 plugin: extend hypercall mechanism to add more operations and support more platforms

jnmartin84:
As part of the work I've done on trying to unit-test the taint2 plugin, a need arose to be able to label bytes and query them for taint directly from guest code running inside of PANDA.

I looked at extending the taint2 hypercall code in order to maintain compatibility with existing use-cases (LAVA) while adding new types of commands/operations. I also extended it to support i386, x86-64 and ARM instead of i386 only with an ARM stub.

It can now do the following operations based on eax/r0 value: 
6: enable taint processing in taint2 plugin
7: label buffer, already existed
8: label buffer pos, already existed
9: query buffer, assert the existence or non-existence of a label on a byte and log details to disk
10: label register, set a taint label on a register instead of memory
11: query register, assert the existence or non-existence of a label on a byte and log details to disk
12: log, writes a message to log on disk

The logged info ends up in current working direct "taint2_log" similar to the way the asidstory plugin writes things to disk.

This has been extensively tested so far for i386. There are a suite of tests that exercise the new functionality in the panda_test repo. They are dependent on these new additions to taint2. They get compiled, copied into PANDA VM and run on guest, producing output on the host useful for CI/regression testing purposes.

Thinking this may be useful for other people, so created a PR to offer it up for general project use. Please let me know what you think.


jnmartin84:
Jenkins froze, bouncing this

PR #672, opened 2020-06-04
python extension improvements

lacraig2:
- make memory access failure more descriptive
- fix relpath bug in symbol resolution
- changes hooks to use enable/disable by name like callbacks


AndrewFasano:
FWIW, I think `panda_virtual_memory_read_external` only can return the error -1. But maybe this is future-proofing it in case we add other errors later.

Once the CI finishes, I'll merge

PR #676, opened 2020-06-09
Capture named process's stdout, stderr, file writes (rehosting extras)

tnballo:
PyPanda rehosting extras:

- New: for a named process, capture stdout/stderr and any file writes from the hypervisor, mirror results to log directory. See example: `panda/python/examples/proc_write_capture.py `
- Update: optionally suppress logs for `ioctl_faker`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,6 @@ trace-dtrace-root.h
 trace-dtrace-root.dtrace
 trace-ust-all.h
 trace-ust-all.c
+
+# Pypanda Examples
+pwc_log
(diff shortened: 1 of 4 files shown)


AndrewFasano:
File descriptors 0-2 aren't always stdin/stdout/stderr - Could you just use OSI's `fd_to_filename` logic instead of treating those cases special?

tnballo:
@AndrewFasano I'd considered that, but 0/`stdin`, 1/`stdout`, 2/`stderr` are defined by the POSIX standard and this plugin uses `osi_linux`. The advantage of the special casing is that `stdout` and `stderr` capture can happen without the need to traverse any kernel data structures for name resolution - that seems more robust/reliable and I'd be very surprised if a Linux system broke POSIX for something as fundamental as file descriptors.

What do you think? If you disagree it's an easy change I don't mind making!

PR #678, opened 2020-06-12
Misc Pypanda bugfixes

AndrewFasano:
Just a few minor fixes.

Also adds a `speedtest` plugin to see how quickly panda is executing guest code.

### DIFF
--- a/panda/plugins/osi_linux/default_profile.cpp
+++ b/panda/plugins/osi_linux/default_profile.cpp
@@ -27,6 +27,7 @@ target_ptr_t default_get_current_task_struct(CPUState *cpu)
     target_ptr_t kernel_sp = get_ksp(cpu);
 
     // XXX: This should use THREADINFO_MASK but that's hardcoded and wrong for my test system
+    // We need to expose that as a part of the OSI config - See issue #651
     target_ptr_t task_thread_info = kernel_sp & ~(0x2000 -1);
 
     current_task_addr=task_thread_info+0xC;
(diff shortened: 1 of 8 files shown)


AndrewFasano:
@tnballo Want to take a look and possibly object to the changes to ioctl faker?

The code is a few weeks old so I'll let it speak for itself instead of trying to explain it too much, but I found that -25 was the return code for whenever an IOCTL is issued which isn't registered to any kernel driver. But if you want it to support every IOCTL then we should probably back that change out...

tnballo:
@AndrewFasano being more conservative with `IoctlFaker` and only changing on `-25` is a probably good idea, no objections! That seems to account for the vast majority and we likely shouldn't be changing other errors - who knows what the ramifications are. 

2 tiny bugs I noticed:
- `self._log` (bool) is no longer used. The new `ignore` might cut down on logging, but it should still be possible to suppress.
- The `__init__` has `log = false` instead of `log = False` - I think that'll be a sneaky type error in Python.

AndrewFasano:
Oh thanks for catching those!

Also, I think making it only work on -25 is the kind of idea that seemed good to me at the end of day on a Friday but now on Monday it sounds like needlessly removing features. Maybe it should default to only -25 but have some way to change that if a user wants...

I'll try to fix the bugs you mentioned and then maybe find a way to support filtering/unfiltered error codes.

Thanks!

AndrewFasano:
Did a rebase instead of merge and that fixed both those issues. Then I also I added support for specifying which ioctls you intercept. I think it's all good now, pending CI

tnballo:
Looks good to me!

PR #679, opened 2020-06-15
Support more qcows from panda-re.mit.edu site

AndrewFasano:
Update the "generic" qcows we provide with for use with pypanda. For each of the following targets, we provide a qcow hosted on panda-re.mit.edu with a "root" snapshot where the guest is logged in. And the default kernelinfo.conf we ship with PANDA supports these targets. Full details of each snapshot are available in [qcows.py](https://github.com/panda-re/panda/blob/06df47aafd5867ce7ff866d13457183b3fa7d50e/panda/python/core/panda/images/qcows.py). The following targets are now supported:
```
name (alias)
----
i386_wheezy
x86_64_wheezy
ppc_wheezy* (ppc)
arm_wheezy (arm)
mips_wheezy* (mips)
mipsel_wheezy* (mipsel)
i386_ubuntu_1604 (i386)
x86_64_ubuntu_1804 (x86_64)
```

\* Note that we currently lack OSI support for MIPS [#707] or PPC (#708) so those targets won't work with OSI.

This PR also cleans up the qcows.py file a bunch and adds adds a test script to run each generic qcow via CLI and python to test basic functionality plus OSI and syscalls2.

Plus a new plugin from @tleek (mostly for debugging) to query taint labels when a specific file is written to disk.


AndrewFasano:
Things I've tested:
x86 boots but with a 'root' snapshot called 'cmdline'
x86_64 boots and OSI works
mips/mipsel boot

PR #686, opened 2020-06-26
Bump gcc

AndrewFasano:
#681  introduced code that depends on `-std=c++1z` which conflicted with the cflags we set in build.sh and wouldn't build in our Xenial container.

This changes our minimum required GCC version from 5 to 7 and updates our Xenial container to install gcc-7.

I'm not too confident in the changes to build.sh but it worked for me and in the containers...

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -68,38 +68,12 @@ fi
 
 msg "Using python2 at: $PYTHON2PATH"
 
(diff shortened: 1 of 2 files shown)


AndrewFasano:
Added a bugfix for this in bf7c9e480cd032bc371bfb45e625f4b096bd3b76

PR #660, opened 2020-05-22
Add Missing Scripts\Artifacts to "make install"

nathanjackson:
This PR adds more scripts to the installed version of PANDA.

General scripts will get installed under: `usr/share/panda/scripts/`.

Plugin\architecture specific artifacts get installed under: `usr/lib/<target architecture>/<plugin>/`

PR #692, opened 2020-07-09
Minor bugfixes

### LINKED ISSUE #687: 'Unicorn Mode' can't read memory for x86
This minimal example should read the contents of memory at 0x1000 into the register EBX. But it appears to get 0 instead. I don't think this has ever actually worked, but I just noticed today.

Note you'll need to quit by typing `q` and pressing enter

```py
#!/usr/bin/env python3

import keystone
import os

from panda import Panda, ffi
from panda.x86.helper import dump_regs, registers

CODE = b"""
mov eax, 0x1000
mov ebx, [eax]
hlt
"""
ks = keystone.Ks(keystone.KS_ARCH_X86, keystone.KS_MODE_32)
ADDRESS = 0x1000
X86_CODE, count = ks.asm(CODE, ADDRESS)
print([hex(x) for x in X86_CODE])

# Set some debugging opts here - dump tcg ops and cpu state. Expose qemu monitor directly
panda = Panda("i386", extra_args=["-M", "configurable", "-nographic", "-d", "op,cpu"], raw_monitor=True)

@panda.cb_after_machine_init
def setup(cpu):
    # map 2MB memory for this emulation
    panda.map_memory("mymem", 2 * 1024 * 1024, ADDRESS)

    # Write code into memory
    panda.physical_memory_write(ADDRESS, bytes(X86_CODE))

    # Set up registers
    cpu.env_ptr.regs[registers["EAX"]] = ADDRESS

    # Set starting_pc
    cpu.env_ptr.eip = ADDRESS

panda.run()
```

AndrewFasano:
* Don't rebuild osi_linux artifacts unless they've changed (introduced in #660)
* Update the i386 configurable machine to be in 32-bit mode instead of 16-bit mode. Eventually we could allow 16-bit as an option, but 32-bit is a better default. Fixes #687 
* Don't change terminal settings to disable echo when running in library mode. Now when you kill a pypanda script, the terminal should be more usable.
* Change our x86 disassembly to intel mode because everyone everyone I've talked to about it prefers it. This is subjective and we can revert it if there are objections.

### DIFF
--- a/chardev/char-stdio.c
+++ b/chardev/char-stdio.c
@@ -49,9 +49,15 @@ static void term_exit(void)
     fcntl(0, F_SETFL, old_fd0_flags);
 }
 
+extern bool panda_library_mode;
+
 static void qemu_chr_set_echo_stdio(Chardev *chr, bool echo)
 {
     struct termios tty;
+
+    if (panda_library_mode) {
+      return;
+    }
 
     stdio_echo_state = echo;
     tty = oldtty;
--- a/disas/i386.c
+++ b/disas/i386.c
@@ -3682,7 +3682,7 @@ static char scale_char;
 int
 print_insn_i386 (bfd_vma pc, disassemble_info *info)
 {
-  intel_syntax = -1;
+  intel_syntax = 1;
 
   return print_insn (pc, info);
 }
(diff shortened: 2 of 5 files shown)


AndrewFasano:
Uh oh, that Jenkins error might actually be something real. I assumed that was failing because the Jenkins tests are slowly being moved to github actions. Will investigate and maybe push a fix shortly.

chardev/char-stdio.c
```diff
@@ -49,10 +49,16 @@ static void term_exit(void)
     fcntl(0, F_SETFL, old_fd0_flags);
 }
 
+extern bool panda_library_mode;
```
AndrewFasano, line 52: This broke `make check` Fixed in #697

PR #695, opened 2020-07-14
Fix tcg to LLVM translation error

LauraLMann:
A macro used by TCG to LLVM translation was emitting invalid LLVM instructions.  This was noticed with a PowerPC recording, after turning some taint2 assertions into warnings so we can use debug builds on our regression tests.  Fortunately, S2E had fixed that translator back in 2014, so it was easy to 'figure out' what to do to fix it.


LauraLMann:
Have no idea with asidstory1 should fail.  My changes only impact taint2 and if someone used the -llvm switch, and the asidstory1 tests I found does not use those.

LauraLMann:
The QEMU Checks and Taint Unit Tests are failing due to "Docker login" failing.  I can't fix that.  It's a configuration issue out of my control.
I can't see why the Jenkins build is failing, because my employer has the website blocked that the Details link takes me to.
And as I explained above, as asidstory1 doesn't use the taint2 plugin or the -llvm switch, I don't see how my changes could have caused that to fail.
Can anybody see why Jenkins is upset and tell me what is going on?

AndrewFasano:
I think these are just some CI bugs, not problems with your code. I'll look into it

AndrewFasano:
Ah, sorry Laura - there's still another CI bug. Should have it fixed in a minute

AndrewFasano:
This passed the checks I care about so I'll merge it! Thanks for putting up with our CI debugging, I promise this isn't a normal occurrence so next time you submit a PR things should go smoother.

PR #697, opened 2020-07-14
CI Bugfixes

AndrewFasano:
Don't do a docker login for github actions except for when we need to push a new docker container.

Cleanup the `panda_library_mode` used by pypanda to be a function instead of a boolean in vl.c which fixes issue with make check introduced in #692


AndrewFasano:
Not sure what's up with the CI - maybe it didn't like all the force pushes? But the errors are about code that's no longer in this PR...

I ran the tests locally and everything passed so hopefully merging this is okay and the CI will just work for future PRs 🤞

PR #698, opened 2020-07-16
Fix CI using stale containers

tnballo:
Subtle bug @AndrewFasano pointed out, with `update.sh` not running and some tests using the latest public docker build instead of the commit to be tested. Caused a lot of headache for #695.


tnballo:
The Jenkins `!!! Couldn't read commit file !!!` error was due to some state becoming inconsistent when a job was manually canceled. Re-trying now.

PR #699, opened 2020-07-17
Update Kenelinfo Documentation

AndrewFasano:
SVN export seems to no longer work (due to a github change?). Adds instructions for downloading a pre-packaged tarball for the kernelinfo generator.


AndrewFasano:
I think the Jenkins error was because Jenkins was unhappy - I doubt the docs update caused a test to fail

PR #700, opened 2020-07-17
Adding osi start_time per process

lacraig2:
This PR adds the `start_time` field to osi.


lacraig2:
moving to local `osi_time` branch.

PR #701, opened 2020-07-17
osi_time merging into mainline

lacraig2:
This adds the `create_time` field to `OsiProc`.


AndrewFasano:
Is this backwards compatible with old kernelinfo files?

lacraig2:
@AndrewFasano It is not.

lacraig2:
Probably a good idea to support. Though optional support for things doesn't seem to be an osi supported feature so far.

AndrewFasano:
Yeah - I think we'd annoy a bunch of PANDA users if we suddenly invalidated all their OSI profiles. Do you think it would be much work to make it backwards compatible?

lacraig2:
I'm looking into a way to make it being missing more of a warning. Until I can do that successfully this will remain a draft PR.

lacraig2:
I've added `IMPLEMENT_OPTIONAL_READ_INT` to make it produce a warning for `start_time` instead of making an error. This seems to work. Ready for review from whoever.

PR #705, opened 2020-07-21
Add missing line feed to print statement in osi_linux

LauraLMann:
Yes, I know this is a little thing, but it was causing one of our regression tests to fail.  A message we were expecting to be on a line by itself was instead concatenated onto the end of the message from osi_linux.

### DIFF
--- a/panda/plugins/osi_linux/osi_linux.cpp
+++ b/panda/plugins/osi_linux/osi_linux.cpp
@@ -693,7 +693,8 @@ bool init_plugin(void *self) {
             goto error;
         }
     }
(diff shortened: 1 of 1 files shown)


AndrewFasano:
I just noticed this a minute ago - I think that printf should just be removed entirely but I'm happy to merge this in for now and then delete it later.

LauraLMann:
I did think it kind of curious that the new printf with missing linefeed was a duplicate of the LOG_INFO line just below.  But I figured someone had a reason for it, so I didn't want to just remove it.  Apparently whatever reason prompted the addition of the printf is now moot.  I'll have no heartburn if someone removes the printf entirely later.

PR #712, opened 2020-07-24
Before Codegen Callback and Coverage Performance Improvements

nathanjackson:
The intent here is to improve the coverage performance and to (hopefully) provide a common infrastructure for coverage plugins since PANDA now has three. As such this PR does the following:

- Adds a callback to PANDA that enables inspection and modification of the TCG block before host code generation.

- Refactors the coverage plugin to use the new callback and insert a callback directly into the TCG Ops. This is enables a significant speed increase for live coverage collection.

- Adds edge coverage into the coverage plugin. Although it should be noted there may still be issues with interrupts.

- Adds various options to filter which blocks should be instrumented.


AndrewFasano:
After this PR, `insn_exec` callbacks are no longer called when an `insn_translate` callback returns true for arm guests. It appears to still work fine for other architectures.

In this example, the `on_insn` function should run because the `should_run_on_insn` returns true.
```py
#!/usr/bin/env python3
import os
from panda import Panda, ffi

panda = Panda(generic="arm")

@panda.cb_insn_translate
def should_run_on_insn(env, pc):
    print("Should run on_insn? Yes")
    return True

@panda.cb_insn_exec
def on_insn(cpu, pc):
    print("Everything works!")
    os._exit(0)

panda.run()
```

@nathanjackson would you be able to look into this or should I?

nathanjackson:
That's strange, I wouldn't have thought these changes could have affected the other callbacks (especially if they're not being used). I can take a look either this afternoon or tomorrow. Any chance you can give me a replay that demonstrates the issue? I have basically no ARM artifacts that I can test with.

AndrewFasano:
If you used the python interface you wouldn't need any ARM artifacts because it automatically gets you everything you need ;)
But I'll generate a small demo in C with a replay (and also confirm the bug isn't just in the python interface).

AndrewFasano:
Unfortunately record/replay is broken for the arm machine but it's easy enough to demo on a live system (the bug shows up on the first instruction executed). I threw together the files you need to reproduce it here: https://panda-re.mit.edu/.other/armbug_poc.tar.gz. If you copy the `armbug` directory into your plugins folder and rebuild panda, you should see the bug when you run `./run.sh`.

Prior to this PR being merged, the plugin will exit on the first instruction (in the insn_exec callback). After this merge, it doesn't (the insn_translate callback is run which prints, but then the insn_exec callback never runs).

PR #714, opened 2020-08-03
CI Fix for Travis

AndrewFasano:
I'm not sure why we're even using Travis for testing anymore but it was failing due to an old GCC version.

Hopefully this PR will get gcc-8 in the Travis environment. Can't (easily) test without creating a PR so this might need some more work...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,17 @@
 dist: bionic
 language: c
-compiler:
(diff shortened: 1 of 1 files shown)


AndrewFasano:
I think travis only tests on master so I'm merging this even though it might not change anything 😞

PR #722, opened 2020-08-10
fix bug in loaded_libs plugin (missing check on get_current_process)

ruaronicola:


AndrewFasano:
Thanks for the bugfix!

PR #726, opened 2020-08-13
rrpack.py prints the wrong number of recorded instruction 

pagabuc:
The number of instructions of a rr-nondet.log file seems to be stored at offset 0, not 16 (but please double check this because I compiled panda for the first time yesterday, so I am not entirely sure  :)


AndrewFasano:
So I'm not too familiar with that format and a comment in rrpack.py describes it differently, but I just tested your patch and I think you're right. I want to get a better understanding before merging but I'll try to look into it more soon.

PR #747, opened 2020-09-04
Fix over-tainting for And/Or instructions

buszk:
Remove taint when it is masked away with immediate And/Or instructions.
e.g. And reg 0xffff0000 should remove taint from the lowest bytes.


AndrewFasano:
That jenkins error is maybe valid, but very unrelated to this PR. I'll merge this

PR #748, opened 2020-09-04
x86_64 configurable machine

AndrewFasano:
This PR adds support for the x86_64 configurable machine

(Rebased on #743 so merge that first)


AndrewFasano:
Looks like github can't see that a bunch of these commits are on `master` after merging #743. I'll reissue this PR

PR #751, opened 2020-09-09
Add MIPS support to callstack_instr plugin

tnballo:
- Adds MIPS support to `callstack_instr`
- Includes a temporary workaround for a [`capstone` bug](https://github.com/aquynh/capstone/issues/1680)
- Tested for big and little endian MIPS - does detect calls (improvement from prior state) but not guaranteed to be 100% accurate (needs usage and testing)


tnballo:
CI error not directly related (PyPanda `cffi.CDefError: cannot parse "struct GHashTable {};"`), but will look into it.

PR #752, opened 2020-09-09
Update dockerfile to pull in newer pycparser

AndrewFasano:
Fixes a bug introduced by ecb5fbf54546cb15da7dc34bd6f0b32acfc1f570 where
`apt install python3-pycparser` was pulling in an old version. Instead we now use pip
to get the latest version of the package.


AndrewFasano:
The CI is noticing the docker failure _now_ instead of when it should have in #743. The container it's pulling for testing doesn't have the right dependencies which is what this PR fixes. In other words, I'm going to merge this because those failures are meaningless here.

PR #753, opened 2020-09-09
Docker bugfix for pycparser

AndrewFasano:
Made a mistake with #752


AndrewFasano:
Failing tests are because the currently published docker container is broken. This will hopefully fix that

PR #762, opened 2020-09-14
Fix panda_current_sp

lacraig2:
`panda_current_sp` currently tries to return the kernel stack pointer every time. This isn't really what is described in the function. 

I've added `panda_current_ksp` to keep the existing logic and changed over the `osi_linux` 2.4 profile to use it. I can't find any other uses in mainline panda.


AndrewFasano:
Can you also update OSI linux to use the generic ksp function? Right now it has its own version

lacraig2:
@AndrewFasano Sure. It looks like callstack_instr has its own too! https://github.com/panda-re/panda/blob/01cc96ada0eefd7a0249e8208d03e00d4fe7e9a9/panda/plugins/callstack_instr/callstack_instr.cpp#L155

PR #766, opened 2020-09-21
Changes to support pushing pypanda to Pypi

AndrewFasano:
This commit changes a bunch of paths and install logic in pypanda so we can build a package and upload to pypi with the following (slightly terrible) code:
```sh
cd panda/python/core
python setup.py install
python setup.py sdist
python3 -m twine upload dist/*.tar.gz
```

Most notably, it changes the pypanda package name to `pandare` (the best name that was available on pypi) and updates the docs/examples to use this new name. Note out-of-tree pypanda plugins will need to be updated.

Now PANDA users can just run `pip3 install pandare` 🎉


panda/python/examples/asid.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 from sys import argv
-from panda import Panda, blocking, ffi
+from pandarere import Panda, blocking, ffi
```
lacraig2, line 4: This seems to be the only instance of it, but this is wrong. Should be `pandare`.

PR #767, opened 2020-09-22
pypanda ghidra mixins

lacraig2:
Simple mixins to support ghidra integration from within panda.

### DIFF
--- a/panda/python/core/panda/main.py
+++ b/panda/python/core/panda/main.py
@@ -2337,5 +2337,54 @@ class Panda():
 
             return wrapper
         return decorator
+    
+         ############# GHIDRA MIXINS ###############
+    def delete_all_ghidra_memory_segments(self,memory, monitor):
+        for block in memory.getBlocks(): 
+            memory.removeBlock(block,monitor)
+
+
+    def populate_ghidra(self,cpu, pc, bridge, analyze=True):
+        tid = currentProgram.startTransaction("BRIDGE: Change Memory Sections")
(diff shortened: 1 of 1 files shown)


AndrewFasano:
Can this be moved into panda.extras? Seems like like it's more of a plugin than a core feature

lacraig2:
There may be some confusion here. 

The way that this is architected requires a user to create a bridge and pass that to panda. It does not assume that ghidra is connected to Panda any more than we assume that volatility is always intended to be used in any given project. That is to say we use a mixin that doesn't establish a hard dependency unless the functionality is called.

I think this makes a lot more sense as part of the core panda functionality.

AndrewFasano:
Hm, I think I was a bit confused but I still disagree that this should be a part of the main pypanda module.

The advantage of isolating specific features like this into their own classes is that the code and documentation would be  cleaner. There'd be less code in the main pypanda module and the main pypanda docs would be easier to follow. And then we can refer interested users to the `pandare.Ghidra` module where there would be all the ghidra functions and docs nicely grouped together. The main documentation for that module could describe how to setup ghidra-bridge and how it works.

At some point in the future I'd like to go through the pandare module and pull out submodules where possible. I'd think this code would nicely fit into a new module (see example below) as would the functions related to volatility, pyperipherals, and probably some other places


Example use case as a separate module
```py
from pandare import Panda, Ghidra

panda = Panda(...)
ghidra = Ghidra(panda)
ghidra.populate()
print(ghidra.decompilation(...))
```

PR #768, opened 2020-09-24
Add support for kernel v2.6.X to kernelinfo module

mberemand:
The symbol `current_task` is not defined in a way that is accessible to `kernelinfo.c` in the 2.6 kernel headers. However, the expansion of it, `per_cpu__current_task` is. Other than that, the kernelinfo module seems to be compatible with 2.6. This was tested with CentOS 6.3 running kernel version 2.6.32-279, and entries for i386 and x86_64 were added to `kernelinfo.conf`.


AndrewFasano:
Hey, thanks for the PR I just ran into this exact issue earlier today so I really appreciate the fix. I'll merge once the CI finishes

AndrewFasano:
Looks like we're having some CI bugs after a change I made yesterday. Obviously those errors aren't related to your code so I'll just merge now.

Thanks again!

PR #770, opened 2020-09-25
Add missing system calls for MIPS syscalls2

tnballo:
- As a quick sanity check, verified that a previously missing syscall, 4090 (`sys_mmap`), now trigger callbacks
- Still needs to be tested with `syscalls_logger` once that makes it to `master`
- Minor fix to `syscalls_parser.py` for Python3


AndrewFasano:
I like this solution but I'm a little scared someone will modify and rerun the generate-prototypes scripts and we'll accidentally throw away these hand-crafted files sometime in the future. Just something to keep an eye out for! Merging

PR #773, opened 2020-10-01
LLVM 3.3 -> LLVM 10

MarkMankins:
WIP: LLVM 10 :)

First - thanks to Matthew Nichols for laying out the initial roadmap for the conversion.  Without his guidance I'm sure I'd still be sifting through documentation at this point trying to figure everything out.

Also a big shout out to Laura Mann for volunteering to tackle the most difficult aspects of the conversion.  Laura dug into the guts of LLVM in order to figure out how to pull out information that was exposed more directly in LLVM 3.  Her solution for replacing the TJITMemoryManager is elegant and should remove the barrier for future LLVM upgrades.

This isn't ready for merging yet, but is stable enough for testing, and I could really use some help identifying what is broken.

Overview of changes:

No longer carrying a reference to the LLVM Function in struct TranslationBlock.  In LLVM 10 the function object is deleted after it is handed off to the JIT compiler.

TCGLLVMContext/TCGLLVMContextPrivate became TCGLLVMTranslator.

In LLVM 3, a single LLVM Module contains all TCG blocks that are translated to LLVM.  In LLVM 10 a new LLVM Module is created for every block translated from TCG to LLVM.  This complicates things as the original code was written to assume everything used a single module.  Function pass managers need to be installed on each new module that is created.

The JIT interface used by LLVM 3 doesn't exist in LLVM 10.  The code was updated to use LLLazyJIT.  The calculations to retrieve the LLVM function address and size were reworked to use the new interface.  The TJITMemoryManager class was removed.

Build process no longer creates panda_taint2_ops.bc.  llvm_taint_lib.cpp no longer loads panda_taint2_ops.bc.

A handful of LLVM 3 taint ops functions expected an Instruction * as a function argument.  These functions iterated over the operands in the instruction as part of the taint processing (see update_cb()).  This was problematic in LLVM 10 because in some cases LLVM 10 would optimize the instruction away.  This left the Instruction * in the IR pointing to a deleted object.  The taint ops interface was updated to pass the instruction opcode and the instruction operands.  For example, taint_copy was defined as:
void taint_copy(Shad *shad_dest, uint64_t dest, Shad *shad_src, uint64_t src, uint64_t size, llvm::Instruction *I)
and is now defined as:
void taint_copy(Shad *shad_dest, uint64_t dest, Shad *shad_src, uint64_t src, uint64_t size, uint64_t opcode, uint64_t instruction_flags, uint64_t num_operands, ...)
PandaTaintVisitor::addInstructionDetailsToArgumentList() is used to add the operands to the taint_ops function's argument list.

Items that still need to be investigated/addressed:

Passing the operands to taint ops functions instead of passing the instruction and iterating over the operands has the side effect that now operands can be casted to ConstantInt.  In the LLVM 3 version when taint ops functions iterated over operands, operands that were instructions could not be casted to ConstantInt.  At first I thought this was a good thing - although now I'm not so sure.  For example, this affects a recent change made to taint_copy (ecc3b1d3c93bb93a161b708b437d161c936d743b).  If we need to know which operands to instructions are constants and which are not, this change might need some rework.  In any case, @buszk  can you check and see how badly I broke this commit?

docs/manual.md - How to use it for analysis section is no longer correct, likely other documentation is out of date

compilation of helpers with clang: optimization level dropped from -O2 to -O1
        at -O2, clang 10 introduces lots of vector operands that aren't handled well by panda
        at -O1, clang 10 isn't aggressively inlining functions
        would be nice to get the clang 10 to inline functions like clang 3 was for performance
        not inlining int128 helpers causes issues, see note and workaround in int128.h
        
configure script updates could be improved
        not sure we want to hard code llvmver=10
        would be nice to determine if llvm::Value::dump() is available at compile time
        the code almost compiles in llvm 11, should be able to detect which version is in use and set some preprocessor symbols

Not sure if the stack smash protection fix in helper_call_modifier is still needed.

Not sure if the issues that existed when unloading plugins in LLVM 3 still exists in LLVM 10.  The fix used in LLVM 3 which involved calling llvm::PassRegistry::unregisterPass() won't work since that function no longer exists.

Noticed the TranslationBlock structure is mirrored in the pypanda code.  Not sure the changes have been mirrored properly/completely, as have not been able to use pypanda yet.

For speed, we might want to consider compiling taint_ops.cpp at -O2 even in debug builds (or at least have an option to do this.)


buszk:
From my understanding, the patch eliminates all llvm::Instruction pointers in panda/plugins/taint2.
Instead, the llvm::Instruction is represented by opcode and operands. However, the operands
only work for constants (variables won't work).

In general, I think this design works, given LLVM's well structured instructions. The commit
you worried could be easily reworked (because constants is usually the second operand). The
patch would obscure variable-typed operands, but the functions called in taint_ops.cpp already
indicate type of operands (to some extend). 

Thank you for the amazing patch!

MarkMankins:
Thanks for checking this out @buszk.  I can easily switch the operand handling for the commit in question if you think that's the right thing to do.

PR #776, opened 2020-10-05
Don't ignore signals when in library mode

### LINKED ISSUE #540: Problems with signal-handling for threaded panda.run() (pypanda)
Hi all,

I'm banging my head against this since yesterday, and have no good solution available (yet), hence this issue.
When running the run() function from pypanda inside a thread, a SIGINT (ctrl+c) will not lead to termination of the process.
Small script to reproduce (modified version of an old unicorn.py, working on current master):
```python                                                               
#!/usr/bin/env python3

import capstone
from sys import argv
from os import path, remove as os_remove
import threading
import time

from panda import Panda, ffi
from panda.arm.helper import dump_regs, registers


code = [ # While true, inc R0. With a nop because I'm bad at shellcoding
    b"\x01\x00\x80\xe2", # add r0, #1
    b"\x01\x10\xa0\xe1", # mov r1, r1 (NOP)
    b"\x10\xf0\x4f\xe2", # sub r15, #16
]
ARM_CODE = b"".join(code)


config_path = "/tmp/config_panda.json"
def make_config(addr=0x1000, size=0x1000):
    # TODO: for each user-specified memory mapping, make it

    # Always make a new config file
    if path.isfile(config_path):
        os_remove(config_path)

    with open(config_path, "w") as f:
        f.write("""{{"memory_mapping": [ {{
          "is_special": false, "is_symbolic": false, "forwarded": false,
          "permissions": "rwx", "name": "mem1", "size": 4096,
          "address": {addr}}} ],
      "entry_address": {addr},
      "kernel": "/dev/null"
    }}""".format(addr = addr))

# Setup memory mappings file
ADDRESS = 0x8000000
make_config(ADDRESS)

# Initialize emulator
args = ["-M", "configurable",
        "-kernel", config_path,
        "-nographic"]
panda = Panda(arch="arm", extra_args=args)

# On machine init, setup reg_writes and mem_writes
@panda.cb_after_machine_init
def my_init(env):
    # Map code into memory
    panda.physical_memory_write(ADDRESS, ARM_CODE)

    # Set registers
    env.env_ptr.regs[registers['R0']] = 0
    env.env_ptr.pc = ADDRESS


# VMI: before each block, print disassembly
md = capstone.Cs(capstone.CS_ARCH_ARM, capstone.CS_MODE_ARM)
block_count = 0
@panda.cb_before_block_exec
def before_block(env, tb):
    global block_count
    print(f"Block 0x{block_count:x}")

    pc = panda.current_pc(env)
    code_buf = ffi.new("char[]", tb.size)
    code = panda.virtual_memory_read(env, tb.pc, tb.size)

    for i in md.disasm(code, tb.pc):
        print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))

    if block_count == 10:
        print("After 10 blocks, R0 = ", hex(env.env_ptr.regs[registers['R0']]))
        #panda.end_analysis()
    block_count += 1

t = threading.Thread(target=panda.run, daemon=True) # daemon or not doesn't make a difference here
t.start()

# Simulate bunch of other operations in main thread
while True:
    time.sleep(1)
print("Finished emulation")
```

When sending SIGINT/SIGTERM/SIGHUB to the process, it hangs, instead of terminating.
My intuition is it that it is related to the way QEMU handles those signals, as it registers its own signal-handler for it, potentially overwriting the ones from python.

I know that running pypanda from a thread is not the intended way of using it, however, in the context of avatar2, execution is driven asynchronously and pypanda is not only interfaced via python, but also via QEMUs gdbstub and QMP (which, impressively, works out of the box).
Hence, please let me know if you are interested in solving this issue, or whether I have to look out for an avatar-local workaround.

Suggested solutions/workaround:
- In `os-posix.c:os_setup_signal_handling`, an additional check whether we are in library mode could be added. If so, don't register QEMUs signal handler. This would break the currently pypanda.was_aborted() functionality, but allows "proper" signal handling from python.
- Introduction of a non-blocking version of pypanda.run(), so that it can be called from the mainthread.

Side observation:
While digging into this, I also noticed that pypandas athread is not stopped by default on end_analysis(). While this is not a problem due to the daemonic nature of that thread, it may be worth considering stopping it as well?

//edit: typo fix

AndrewFasano:
When the python interpreter is blocking on a `panda.run()` call, if a user presses Ctrl+C we want to raise a `KeyboardInterrupt` exception and end the panda execution. Currently we just end the panda execution with no indication of interruption.

This PR is a work in progress - it disables the PANDA signal handlers when in library mode. But it looks like the exception is being caught and dropped by CFFI instead of passed to anything useful.

This is trying to fix #540

### DIFF
--- a/os-posix.c
+++ b/os-posix.c
@@ -51,11 +51,14 @@ static int daemon_pipe;
 void os_setup_early_signal_handling(void)
 {
     struct sigaction act;
-    sigfillset(&act.sa_mask);
-    act.sa_flags = 0;
-    if (panda_get_library_mode()) { // Don't ignore signals in library mode
+
+    if (panda_get_library_mode()) {
+       // Don't ignore signals in library mode
       return;
     }
+
+    sigfillset(&act.sa_mask);
+    act.sa_flags = 0;
     act.sa_handler = SIG_IGN;
     sigaction(SIGPIPE, &act, NULL);
 }
@@ -68,6 +71,11 @@ static void termsig_handler(int signal, siginfo_t *info, void *c)
 void os_setup_signal_handling(void)
 {
     struct sigaction act;
+
+    if (panda_get_library_mode()) {
+       // Don't ignore signals in library mode
+      return;
+    }
 
     memset(&act, 0, sizeof(act));
     act.sa_sigaction = termsig_handler;
--- a/os-posix.c
+++ b/os-posix.c
@@ -53,6 +53,9 @@ void os_setup_early_signal_handling(void)
     struct sigaction act;
     sigfillset(&act.sa_mask);
     act.sa_flags = 0;
+    if (panda_get_library_mode()) { // Don't ignore signals in library mode
+      return;
+    }
     act.sa_handler = SIG_IGN;
     sigaction(SIGPIPE, &act, NULL);
 }


AndrewFasano:
This was done better by #779

PR #784, opened 2020-10-13
Move TCG utility functions

mberemand:
Allows for `insert_call()` to be used by other plugins.


AndrewFasano:
Thanks for the PR!

PR #798, opened 2020-10-26
Conditional ci

AndrewFasano:
Slight update to readme.

If a commit message contains "no ci" don't run ci or rebuild container (e.g., for doc updates)


AndrewFasano:
Hm, this strategy won't work with required tests so maybe it's not worth it.

PR #802, opened 2020-10-28
Catch SIGABRT in C from Python

lacraig2:


lacraig2:
Yeah. This is a bad idea. Don't do this. ¯\_(ツ)_/¯

PR #815, opened 2020-11-13
Fix protobuf version

AndrewFasano:
Importing `pandare.plog_pb2` was failing because the installed version of protobuf was outdated. Updates docker containers to use latest. This was calling our documentation-updating CI to fail.

I had hardcoded 3.0.0 to make sure the python version matched the system version, but now I don't think that was correct/necessary.

Also fixes a broken pypanda test.


AndrewFasano:
I think there still might be a bit of a bug/annoying edge case here as described by [this post](https://stackoverflow.com/a/30175210). Ubuntu 18.04 protobuf from apt was built with an older GCC so the API doesn't match when newer gcc is used. Perhaps this can be fixed by adding `-D_GLIBCXX_USE_CXX11_ABI=0` to our cflags somewhere when we detect an old protobuf install.

Issue #675, opened 2020-06-06
plog_reader binary does not show all field


DOT850:
You all might be aware of this, but I kind of got lost on this. The plog_reader in panda/arm-softmmu/plog_reader only seems to show a limited output of data (instr and PC). You need to use the python script to get the full output. A fix might not be required, but might be helpful to update the manual which says you can use either?

PR #819, opened 2020-11-16
Plog documentation

eneytcheva:
Fixed issue #675 by updating the manual and renaming plog_reader.cpp to example_plog_reader.cpp


AndrewFasano:
CI failed because it ran out of disk space, not at all your fault.

I just fixed that so I'll have the CI run again.

PR #820, opened 2020-11-17
Pypanda cleanup for Jupyter Notebooks

AndrewFasano:
General pypanda cleanup, mostly to improve UX with Jupyter notebooks


AndrewFasano:
I think the change in hooks to using before-block-exec instead of before-block-exec-invalidate-opt was wrong. BBE-io fires before a block is executed every time, not just at translation. The commit message and code I changed seem to suggest I was confused about that.

PR #821, opened 2020-11-19
Coverage Plugin: Edge Instrumentation Improvements

nathanjackson:
This PR fixes several issues with edge coverage in the `coverage` plugin. In the new approach I've taken an approach similar to Frida Stalker (https://frida.re/docs/stalker/). Instead of emitting an edge every other block, the edge coverage mode now actually instruments control flow instructions. We hook control flow instructions and then statically or dynamically figure out what the control flow destinations are. So far, this seems to make the edge coverage mode a lot more robust, but it does require OSI support because we have to know when a thread change occurs since threads can change at basic block boundaries. I'm still not sure this fixes issues with interrupts completely, but I've not found any edges that I would consider to be wrong. There's a couple of limitations though. First, it looks like QEMU (and PANDA by extension) caches jumps that occur in jump tables so the corresponding block of code in a jump table is never truly executed. In these cases, the edges will be omitted, but I don't think there's much we can do about it. I'm not sure it affects the global view of coverage that much either. Second, this approach is currently limited to X86, but I don't think there's a way to do this by just looking at the TCG block since we lose some context by just looking at TCG ops.

Since this new way of instrumenting edges is fairly intensive because of the use of OSI, I sought to make some performance improvements there. The solution was to add the ability for OSI to notify plugins that the current task has changed. To do this, we hook specific parts of the guest's context switch after the registers have been restored. This has been pretty well tested on Linux guests and Windows 2000 + XP. I did some Windows 7 testing and it seems to work, but I don't have a lot of recordings or cases I can test with. Since we insert code via TCG to support task change notifications, this required me to make some small changes to TCG -> LLVM and taint2 by extension. Before, TCG -> LLVM would assume that the only time a call was occurring was for a call to a helper function. Since the calls can be arbitrary now, I had to handle that case in both TCG -> LLVM and the Taint Visitor.

All of my own tests pass, but any input is appreciated.


AndrewFasano:
I'm trying to use this branch and seeing some non-deterministic build failures. There's a chance it's caused something else, but I wanted to note it here just in case it's relevant:
```
...
  API     x86_64-softmmu/panda/plugins/winxpx86intro/winxpx86intro_ext.h
  API     x86_64-softmmu/panda/plugins/taint2/taint2_ext.h
  File "/home/fasano/panda/panda/scripts/apigen.py", line 191, in <module>
    generate_api(sys.argv[1], sys.argv[2], sys.argv[3:])
  File "/home/fasano/panda/panda/scripts/apigen.py", line 177, in generate_api
    func_spec += (arglist[func_spec[1]],)
KeyError: 'void'
```

PR #824, opened 2020-11-23
Use cpu->panda_guest_pc in llvm mode

buszk:
LLVM translation updates `panda_guest_pc` before each instruction emulation.
Current panda_current_pc return pc of the basic block's start, because it is updated for every basic block.

It is possible to cause problems for a plugin that expects `panda_guest_pc` consistent before and during llvm mode.
However, no one exists to my knowledge.


nathanjackson:
This change looks like it may have had some unintended consequences, namely it seems to change the results for various plugins that use Taint\Callstack. Is there a reason you aren't going through the CPU state to obtain the accurate PC?

My understanding was that if you needed a more accurate PC, you were supposed to obtain it from the CPU state and enabling the precise program counter.

nathanjackson:
@buszk The documentation indicates that you should go through the CPU state: https://github.com/panda-re/panda/blob/master/panda/docs/manual.md#precise-program-counter

buszk:
Oh. Sorry for missing the document. Intuitively, I thought
panda_current_pc() should return the precise value after
panda_enable_precise_pc().

However, I think the change is necessary. Qemu updates PC
undeterministically, meaning two records of same execution may end up at
different PC point. That may relate to how Qemu split basic blocks.

In my case, I expect taint_branch return the address of branch instruction,
rather than the address of the basic block(which might change across
records).

PR #840, opened 2020-12-18
make panda handle errors in the blocking thread gracefully.

lacraig2:
I merged and then unmerged because I wanted a review from @AndrewFasano first, but forgot.


AndrewFasano:
Looks good to me

PR #853, opened 2021-01-12
Documentation updates

LauraLMann:
Made some code changes in plugins so that the documentation lines up with the plugins.  Some examples called out in the pre-existing documentation did not work as advertised.  These issues seem related to plugins failing to require dependencies correctly.  Some little documentation tweaks were also made to make it easier to process the markdown in pandoc (but without breaking the markdown as it appears in Git's web interface).
@MarkMankins must be thanked for most of this work - I'm just carrying on where he left off.


AndrewFasano:
This is fantastic, thanks for the PR!

PR #854, opened 2021-01-15
callstack_instr: Explicitly fail for arm guests when capstone version insufficient

### LINKED ISSUE #641: callstack_instr: fails to work in Ubuntu 18.04 for ARM targets
The version of libcapstone on Ubuntu 18.04 (v3.0.4) doesn't classify ARM instructions well. The latest version (v4.0.1) appears to do a better job with ARM. I didn't track down when they actually resolved the issue, but I did find a bug report that may be related.

Potentially related: https://github.com/aquynh/capstone/issues/1072

Regarding panda sources, the if statement at line 353 within in callstack_instr.cpp, is never reached. Specifically, `cs_insn_group(handle, end, CS_GRP_CALL)` at line 295 is never true.

References to source code:

https://github.com/panda-re/panda/blob/master/panda/plugins/callstack_instr/callstack_instr.cpp#L295

https://github.com/panda-re/panda/blob/master/panda/plugins/callstack_instr/callstack_instr.cpp#L353

I wrote the following simple application to verify that capstone would classify
correctly. It will print "call" on success and "not call" on failure. Compiled
with `gcc -Wall -lcapstone main.c`.

```
#include <stdio.h>
#include <capstone/capstone.h>
#include <capstone/arm.h>

int main(){
    csh handle;
    unsigned char buf[4] = {0x56,0x34,0x00,0xeb}; /* BL	#0xD160 */
    int size = 4;
    cs_insn *insn;

    if (cs_open(CS_ARCH_ARM, (cs_mode)CS_MODE_LITTLE_ENDIAN, &handle) != CS_ERR_OK)
        return false;

    cs_option(handle, CS_OPT_MODE, CS_MODE_ARM);
    cs_option(handle, CS_OPT_DETAIL, CS_OPT_ON);

    cs_disasm(handle, buf, size, 0x1000, 0, &insn);

    printf("capstone version: %d.%d\n", CS_API_MAJOR, CS_API_MINOR);

    if (cs_insn_group(handle, insn, CS_GRP_CALL)) {
        printf("call\n");
    } else {
        printf("not call\n");
    }
}
```

Output on Ubuntu 18.04 (v3.0.5):
```
$ ./a.out
capstone version: 3.0
not call
```

Output with latest release (4.0.1):
```
$ ./a.out
capstone version: 4.0
call
```

AndrewFasano:
Fixes #641

### DIFF
--- a/panda/plugins/callstack_instr/callstack_instr.cpp
+++ b/panda/plugins/callstack_instr/callstack_instr.cpp
@@ -526,6 +526,13 @@ bool init_plugin(void *self) {
 #elif defined(TARGET_ARM)
     if (cs_open(CS_ARCH_ARM, CS_MODE_ARM, &cs_handle_32) != CS_ERR_OK)
         return false;
+
+#if CS_VERSION_MAJOR < 4
+        printf("\n[ERROR] Capstone versions prior to 4.0.1 are unusable with ARM so callstack instr will fail! Please upgrade your libcapstone install to use this plugin\n\n");
+        return false;
+#endif
+
+
 #elif defined(TARGET_PPC)
     if (cs_open(CS_ARCH_PPC, CS_MODE_32, &cs_handle_32) != CS_ERR_OK)
         return false;


AndrewFasano:
This detects build-time errors, but we should also detect runtime errors

PR #857, opened 2021-01-21
proper freeing of qemu_file in PANDA_MAIN_FINISH

mariusmue:
qemu_file is also free'd in panda_set_qemu_path (https://github.com/panda-re/panda/blob/master/panda/src/panda_api.c#L97).
In corner cases, this can lead to a double-free as in PANDA_MAIN_FINISH qemu_file is neither checked against NULL, nor set to NULL after free'ing it.
This PR is a minimal fix for this.


tnballo:
CI should be back up, re-running.

tnballo:
Good catch, thanks!

PR #859, opened 2021-01-22
expose end_replay to the main panda.py object

mariusmue:
As pypanda allows now mulitple replays to be run, `end_replay` should be a function provided by the main panda object - which is added with this PR.


tnballo:
CI should be back up, re-running.

tnballo:
Looks great, thanks!

Issue #865, opened 2021-01-31
Support enabling/disabling PPP-style callbacks


AndrewFasano:
Currently regular PANDA callbacks (i.e., `before_block_exec`) can be created, enabled, disabled, or deleted.

However, PPP-style callbacks (i.e., `syscalls2`'s `on_sys_open_enter`) can only be created and deleted.

It would be nice to support enabling and disabling these types of callbacks.

PR #867, opened 2021-02-02
dynamic symbols and hooks updates

lacraig2:
This PR totally redesigns the hooks plugin and adds a new plugin `dynamic_symbols`. 

## hooks

Hooks now have a different prototype and structure. They are now defined using the following structure:
```
typedef struct hook {
    target_ulong start_addr;
    target_ulong end_addr;
    target_ulong asid;
    panda_cb_type type;
    hooks_panda_cb cb;
    enum kernel_mode km;
    bool enabled;
} hook;
```

When setting a hook a user may select from a list of `block`-type callbacks (e.g. `BEFORE_BLOCK_EXEC`, `AFTER_BLOCK_EXEC`, etc). This will be the callback type they receive.

On receiving a hooked callback a user will receive, in addition to the normal callback arguments, a pointer to the hook that called the function. Users may, at this point, change the hook in any way they choose. If they choose to set `enabled` to false the hook will be removed from the pool.

This removes some functionality from hooks as well. We remove the ability to filter by `procname` and `libname` for a few reasons. 
  - Checking anything with OSI, even with caching, on every basic block is far too slow and the old model probably didn't work very well anyhow.
  - We have shifted to a callback scheme where callbacks triggered can be changed.

Filtering can still be achieved with ASID filtering. You must simply disable hooks whose ASIDs are associated with terminated processes to avoid hooks being triggered on recycled ASIDs.

For example:

```
@panda.hook(0x0, length=0x200, enabled=True, kernel=True, cb_type="after_block_exec")
```

This hook will trigger on every "after_block_exec" where it is in the kernel and the program counter is between 0x0 and 0x200.

## dynamic_symbols

`dynamic_symbols` is a Linux-based PANDA plugin which automatically resolves the symbols for each ELF in memory for each process with the help of `osi`. Users can use `hook_symbol` to hook symbols which may be defined in the future.

For example:

```
@panda.hook_symbol("libc", "write")
def hook_write(env, tb, h):
    arg1 = panda.arch.get_arg(env,1)
    try:
        trout = panda.read_str(env, arg1)
    except:
        trout = "?"

    print(f"{panda.get_process_name(env)} write \"{trout}\" 0x{arg1:x}")
```

For a full example see: [here](https://github.com/panda-re/panda/blob/proc_lib_cov/panda/python/examples/bins.py)


panda/plugins/dynamic_symbols/dynamic_symbols_int_fns.h
```diff
@@ -0,0 +1,48 @@
+#ifndef __DYNAMIC_SYMBOLS_INT_FNS_H__
+#define __DYNAMIC_SYMBOLS_INT_FNS_H__
+
+// BEGIN_PYPANDA_NEEDS_THIS -- do not delete this comment bc pypanda
+// api autogen needs it.  And don't put any compiler directives
+// between this and END_PYPANDA_NEEDS_THIS except includes of other
+// files in this directory that contain subsections like this one.
+#define MAX_PATH_LEN 256
```
AndrewFasano, line 8: Does this compiler directive break the pypanda cffi parser stuff? I assume it doesn't, but it directly violates the rules in the comment above it which is amusing

panda/plugins/dynamic_symbols/tester.py
```diff
@@ -0,0 +1,392 @@
...
+    return 0
+
+def hook_write(env,tb):
+    fd = get_arg(env,1)
+    buf = get_arg(env,2)
+    n = get_arg(env,3)
+    #try:
+    #    qq = panda.read_str(env,buf)
+    #except:
+    qq = "[??]"
```
AndrewFasano, line 170: Shouldn't this try reading? Also, maybe only read/print up to `n` bytes?

panda/plugins/hooks/USAGE.md
```diff
@@ -10,10 +10,40 @@ Arguments
...
+```
+
+Changes made to the hooks in callbacks are propagated. Hooks disabled are removed.
+
+
+struct hook
+------------
+
+The `hook` struct allows a user to set up a hook on any of several callbacks (see: `hooks_panda_cb`). They may specify the type with the `type` parameter. A basic hook will set a `start_addr` and an `end_addr` which covers a region in which callbacks should be triggered. Users may also specify an `asid` to filter on the basis of or specify that the `asid` is 0 to trigger on all asids. Users may also use `enum kernel_mode` to set `km` to filter on the basis of the kernel mode state. Lastly, users may set the `enable` boolean in the struct. Structs which are disabled after a user has seen them will never be used again and so are removed.
+
```
AndrewFasano, line 47: So here disabled means deleted? No way to disable then later re-enable? If so the docs are clear, just double checking

panda/plugins/hooks/hooks.cpp
```diff
@@ -15,130 +15,432 @@ PANDAENDCOMMENT */
...
-                ++i;
-            }
+vector<hooks_panda_cb> symbols_to_handle;
+
+void handle_hook_return (CPUState *cpu, struct hook_symbol_resolve *sh, struct symbol s, OsiModule* m){
+    int id = sh->id;
+    hooks_panda_cb resolved = symbols_to_handle[id];
+    struct hook new_hook;
+    target_ulong offset = s.address;
+    printf("handle_hook_return @ 0x%llx for \"%s\" in \"%s\" @ 0x%llx ASID: 0x%llx\n", (long long unsigned int)rr_get_guest_instr_count(), sh->name, sh->section, (long long unsigned int) s.address, (long long unsigned int) panda_current_asid(cpu));
```
AndrewFasano, line 97: Should this be guarded by like `if(debug)` or commented?

panda/plugins/hooks/hooks.cpp
```diff
@@ -15,130 +15,432 @@ PANDAENDCOMMENT */
...
-	hooks[value].push_back(hook);
 }
 
-void enable_hook(hook_func_t hook, target_ulong value){
-	update_hook(hook, value);
-
+bool vector_contains_struct(vector<struct hook> vh, struct hook* new_hook){
+    for (auto &h: vh){
+        if (memcmp(&h, new_hook, sizeof(struct hook)) == 0){
+            printf("contains\n");
```
AndrewFasano, line 135: Comment/remove?

panda/plugins/hooks/hooks.cpp
```diff
@@ -15,130 +15,432 @@ PANDAENDCOMMENT */
...
+                        (h.start_addr <= tb->pc && tb->pc + tb->size <= h.end_addr))){
+                        (*(h.cb.before_tcg_codegen))(cpu, tb, &h);
+                        if (!h.enabled){
+                            it = before_tcg_codegen_hooks.erase(it);
+                            continue;
+                        }
+                        memcpy(&(*it), (void*)&h, sizeof(struct hook));
+                    }
+                }
+            }
```
AndrewFasano, line 215: I think this logic is duplicated a bunch? Could it be a macro? Like you could specify the `before_tcg_codegen_hooks` as a string and then parameterize the whole block using that as a variable?

AndrewFasano: I guess the return types aren't all the same, but maybe it could be two versions (bool or void)? maybe not worth it

panda/python/core/pandare/panda.py
```diff
@@ -2404,142 +2410,127 @@ def clear_breakpoint(self, cpu, pc):
...
-            if h.is_kernel:
-                continue
 
-            if h.program_name:
-                if (h.program_name != name):
-                    if h.is_enabled:
-                        self.disable_hook(name)
-                    continue
+        def decorator(fun):
+            # Ultimately, our hook resolves as a beforde_block_exec_invalidate_opt callback so we must match its args
```
AndrewFasano, line 2421: Is this still true? Also there's a typo `beforde`

AndrewFasano (changes requested):
This looks fantastic. I'm really impressed by how complex this challenge is and the solutions you came up with. I was interested in this enough to actually read through all the code - so I left a bunch of comments, but they're all fairly minor and certainly don't need to be addressed unless you want to.

AndrewFasano:
I guess I "requested changes" but if you want to ignore them all, that's fine by me.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+            }
+            ptr+=1;
+        }else{
+            break;
+        }
+    }
+    return buf;
+}
+
+void gdb_helper(){};
```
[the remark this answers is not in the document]
lacraig2: Good catch. For some reason line number breakpoints weren't working for me so I created a blank function and set up a gdb function to break on `gdb_helper` and finish the function. So I could just add `gdb_helper()` wherever I wanted a breakpoint. I'll remove it.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+        //printf("trying gnu_hash\n");
+        int result = get_numelements_gnu_hash(cpu, gnu_hash);
+        if (result != -1)
+            return result;
+    }
+    // we don't actually have the size of these things 
+    // (not included) so we find it by finding the next
+    // closest section
+  //  target_ulong strtab_min = strtab + 0x100000;
+    printf("continuing onto the end\n");
```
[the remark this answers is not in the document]
lacraig2: Yes. It's debugging. I was checking to see if it ever got there. It hasn't yet, which is good, this is the failsafe if ever the first two methods don't work. I'll remove.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+        int numelements_dyn = dynamic_phdr.p_filesz / sizeof(ELF(Dyn));
+        //assert(false);
+        //printf("numelements_dyn: %d\n", numelements_dyn);
+        
+        // iterate over dynamic program headers and find strtab
+        // and symtab
+        ELF(Dyn) tag;
+        target_ulong strtab = 0, symtab = 0, strtab_size = 0, dt_hash = 0, gnu_hash = 0;
+        int j = 0;
+        // 100 for sanity
```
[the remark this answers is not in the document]
lacraig2: It's an old comment from when I hadn't quite figured out the bounds and it was running off the end and I would set a maximum value. Removed.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+                    s.address = m->base + a->st_value;
+                    //printf("found symbol %s %s 0x%llx\n",s.section, &strtab_buf[a->st_name],(long long unsigned int)s.address);
+                    symbols_list_internal.push_back(s);
+                    check_symbol_for_hook(cpu, s, m);
+                    //printf("%s %s %llx\n", m->name, s.name, (long long unsigned int)s.address+ a->st_name);
+                }
+            }
+            proc_mapping[name] = symbols_list_internal;
+            symbols[asid] = proc_mapping;
+        }else{
```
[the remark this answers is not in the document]
lacraig2: Fair.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+void hook_program_start(CPUState *env, TranslationBlock* tb, struct hook* h){
+    //printf("got to program start 0x%llx\n", (long long unsigned int)rr_get_guest_instr_count());
+    update_symbols_in_space(env);
+    h->enabled = false;
+}
+
+bool first_require = false;
+
+void bbe_execve(CPUState *env, TranslationBlock *tb){
+    if (unlikely(!panda_in_kernel(env))){
```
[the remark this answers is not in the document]
lacraig2: `!panda_in_kernel` could not, itself, be characterized as likely or unlikely on a generic system.

You have to understand it in the context. We enable `bbe_execve` after an `sys_execve` type call. It then enters the kernel and waits for the very first userland call. When this function executed we expect hundreds of kernel blocks before a single userland block. That's why its unlikely that we're in userspace.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+                if (panda_virtual_memory_read(env, sp+(ptrlistpos*sizeof(target_ulong)), (uint8_t*) &entrynum, sizeof(entrynum)) != MEMTX_OK || panda_virtual_memory_read(env, sp+((ptrlistpos+1)*sizeof(target_ulong)), (uint8_t*) &entryval, sizeof(entryval))){
+                    panda_disable_callback(self_ptr, PANDA_CB_BEFORE_BLOCK_EXEC, pcb_bbe_execve);
+                    return;
+                }
+                ptrlistpos+=2;
+                if (entrynum == AT_NULL){
+                    break;
+                }else if (entrynum == AT_ENTRY){
+                    if (!first_require){
+                        panda_require("dynamic_symbols");
```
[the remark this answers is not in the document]
lacraig2: This should be `hooks`. Good catch. I copied that code over from the other plugin.

panda/plugins/dynamic_symbols/dynamic_symbols_int_fns.h
lacraig2: It does not break things (though at one point it did break pycparser).

I don't think I wrote that autogenerated comment, but the "compiler directives" that are at issue are things like `ifdef` and `if` and such.

panda/plugins/hooks/hooks.cpp
lacraig2: yeah.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+                break;
+            }else if (dynamic_phdr.p_type == PT_NULL){
+                //printf("hit PT_NULL\n");
+                return;
+            }else if (j == phnum -1){
+                //printf("hit phnum-1\n");
+                return;
+            }
+        }
+        //char* dynamic_section = (char*)malloc(dynamic_phdr.p_filesz);
```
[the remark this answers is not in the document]
lacraig2: sure.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+    }
+    struct symbol blank;
+    blank.address = 0;
+    char none[] = "NULL";
+    strncpy((char*) & blank.name, none, 5);
+    strncpy((char*) & blank.section, none,5);
+    return blank;
+}
+
+string read_str(CPUState* cpu, target_ulong ptr){
```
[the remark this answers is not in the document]
lacraig2: I seem to write it a lot. It's probably worth doing, but I wouldn't want it to return an `std::string` and I also don't want to manage the memory. A prototype for something like this might be 

`void read_str(CPUState* cpu, target_ulong addr, char* buf, int maxlen)`

idk. probably worth doing at some point.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+                    strncpy((char*) &val.section, section.first.c_str(), MAX_PATH_LEN -1);
+                    return val;
+                }
+            }
+        } 
+    }
+    struct symbol blank;
+    blank.address = 0;
+    char none[] = "NULL";
+    strncpy((char*) & blank.name, none, 5);
```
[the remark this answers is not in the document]
lacraig2: Yeah. I had thought I changed all these out. I can't make them pointers to NULL because they aren't managed memory. They're an inline part of the struct. I have changed this around to `memset` the strings to all `0`. I think that's a better solution.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+#include "hooks/hooks_int_fns.h"
+#include "dynamic_symbols_int_fns.h"
+#include "syscalls2/syscalls_ext_typedefs.h"
+#include "syscalls2/syscalls2_info.h"
+#include "syscalls2/syscalls2_ext.h"
+
+}
+using namespace std;
+
+map<target_ulong, map<std::string, target_ulong>> mapping;
```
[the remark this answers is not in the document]
lacraig2: That's a good note. I'll remove those.

panda/plugins/hooks/USAGE.md
lacraig2: I had previously returned a pointer to users on adding hooks. I removed that functionality so that you can only change hooks when they're actively run.

I suppose the question is, should users be able to maintain an active reference to hooks and, if so, what are the guarantees?

Without the ability to get references to hooks disabling hooks removes them because without a reference they can't be set back to enabled and therefore should be gone.

It's a difficult problem, especially when considering reasonable efficiency. We don't want to run every callback all the time. 

We could do something like give them unique identifiers (not pointers) back to the application and hold onto them and then disable the callback if all of them are disabled and then enable the callback type when something like `hook_enable_callback` is enabled? It's tough developing a model for it and this is a consequence of the one that I have taken. I'm certainly willing to take another look at that model if you have concerns.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -0,0 +1,572 @@
...
+    }
+    // we don't actually have the size of these things 
+    // (not included) so we find it by finding the next
+    // closest section
+  //  target_ulong strtab_min = strtab + 0x100000;
+    printf("continuing onto the end\n");
+    target_ulong symtab_min = symtab + 0x100000;
+    ELF(Dyn) tag;
+    for (int j=0; j< numelements_dyn; j++){
+        if (panda_virtual_memory_read(cpu, dynamic_section + j*sizeof(ELF(Dyn)), (uint8_t*)&tag, sizeof(ELF(Dyn))) != MEMTX_OK){
```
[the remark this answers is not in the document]
lacraig2: That structure certainly exists a fair amount. However, there is still some variation (void) and also, it's an error case, so I'd rather keep it open so I can potentially put debug prints in there.

panda/python/core/pandare/panda.py
lacraig2: It was before we talked about changing it. Now it is not. Removing.

panda/plugins/hooks/hooks.cpp
lacraig2: It's just frustrating enough to not be quite that easy to macro. I mostly just don't want to. Perhaps at some point in the future.

panda/plugins/hooks/hooks.cpp
lacraig2: yeah. that should be debug.

lacraig2: Though I expect it to be the first thing uncommented if something goes awry. But it's excessive in the larger cases, yes.

panda/plugins/dynamic_symbols/tester.py
lacraig2: Yeah. Probably worth doing there. Changed.

panda/plugins/dynamic_symbols/README.md
```diff
@@ -0,0 +1,34 @@
...
+struct symbol resolve_symbol(CPUState* cpu, target_ulong asid, char* section_name, char* symbol);
+```
+
+The `resolve_symbol` function is the core of the dynamic_symbols plugin. It resolves the symbol within a particular ASID. The section name search filters by substring (e.g. "libc" will match "libc.so.1"). The symbol search filter must match exactly.
+
+```C
+void hook_symbol_resolution(struct symbol_hook *h)
+```
+
+You may alternatively get a callback when a specified symbol has been resolved.
```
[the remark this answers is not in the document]
lacraig2: I've changed this around to make it slightly more clear. Let me know what you think.

lacraig2:
Definitely not! I appreciate the time and effort you put in to do a thorough review. I've commented on each of them and made changes where I felt necessary for this round. I resolved the easy ones.

panda/plugins/hooks/USAGE.md
AndrewFasano: Ah okay, this design makes sense! I was just thinking about #865 where I was annoyed that we couldn't disable and later re-enable but this makes sense. Efficiency is important for hooking.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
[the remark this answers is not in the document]
AndrewFasano: Yeah, sounds like a problem for later

AndrewFasano (approved):

PR #868, opened 2021-02-02
kernelinfo.c fix

off-by-1-error:
A simple fix for kernelinfo.c, because it used to only compile on x86 and x64 systems. It now produces the same results as the kernelinfo_gdb script. 

Tested on x86, x64, arm, and mips.


AndrewFasano:
Looks great. Nice job with this!

PR #869, opened 2021-02-04
Fix memory leak in OSI in_shared_object

jamcleod:
According to the tests in osi_linux, the `mappings` GArray is an owned GArray (link below), so this should be correct, however I do not have anything that uses this on hand to properly regression test against.

https://github.com/panda-re/panda/blob/c9b275ec2ffd8cb45f98dff5b4fc90e7ef6b3836/panda/plugins/osi_linux/osi_linux.cpp#L580-L586


tnballo:
Great find, thanks!

PR #875, opened 2021-02-12
Pypanda: Update FileFaker & IoctlFaker

KevinCooper:
This PR is to fix and add the following items:
- Ignore string decode errors in IoctlFaker to prevent crashing during run
- Set the buffer to None instead of an exception if the IoctlFaker is unable to read a memory region
- Allow the user to specify a custom file on disk for `replace_file` instead of being hard coded to `/etc/passwd`
- Move the log statement down for `file_faker.write` since it requires `new_data`.  Removed the `self.` for `offset` since it is a local variable, not a class variable.
- Update `file_fake` test to include a test for writing data to the faked file in the emulated OS.

### DIFF
--- a/panda/python/core/pandare/extras/file_faker.py
+++ b/panda/python/core/pandare/extras/file_faker.py
@@ -217,7 +217,7 @@ class FileFaker(FileHook):
             for name in names:
                 self._gen_fd_cb(name, arg_offset)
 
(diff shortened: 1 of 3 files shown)


AndrewFasano:
Thanks for the well-documented PR!

PR #879, opened 2021-02-16
Makefile: fix length calculation with recent 'make'

Hyask:
'make' >= 4.3 has a breaking change in the way it escapes the '#' character. A hack is needed to make it work with
multiple versions of 'make'.
More details can be found in the 'make' changelog (https://lwn.net/Articles/810071/).

This already affects Debian testing and the latest Ubuntu at least.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -626,12 +626,17 @@ endif
 endif
 
 ifneq ($(CONFIG_LINUX),)
+
+# This hack is needed with `make` >=4.3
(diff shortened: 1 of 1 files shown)


AndrewFasano:
I've never seen a backwards-incompatible change in `make` before. Thanks for the fix!

Hyask:
> I've never seen a backwards-incompatible change in `make` before. Thanks for the fix!

Yeah, me neither. Thanks for the quick merge.

PR #882, opened 2021-02-20
Augment Taint2 with symbolic information

buszk:
The PR is migrated from my on-going project where I collect symbolic branches.
I augment TaintData with a field of z3::expr to keep track of symbolic information at byte-level.

The code still requires some fix and cleaning.
The code was written with panda llvm-3.3, and would require more testing on the llvm-10.

Currently, it only prints symbolic branch information. 
With pypanda installed, you can run `python panda/python/examples/unicorn/taint_sym_x86_64.py`
In the output, you can find the symbolic path constraint, like the following
```
PC: 1013
Path constraint: (let ((a!1 (= ((_ extract 15 8)
                (bvadd (concat #x000000000000 val_1 val_0)
                       (concat #x000000000000 val_3 val_2)))
              #x00)))
(let ((a!2 (and (= val_0 (bvadd #x41 (bvmul #xff val_2))) a!1)))
  (not a!2)))
```
which is equivalent to `(256 * val_1 + val_0) + (256 * val_3 + val_2) != 0x41`

Todo list:
- [ ] Plugin structure? As a new plugin rather than mod taint2?
- [x] Some other plugin needs change (tainted_branch)
- [x] More proper API for symbolic info
- [x] Add z3 to dependencies
- [x] Add tests


nathanjackson:
Just throwing some thoughts out there, I tend to agree with your TODO that questions whether or not  this should be a separate plugin. From a maintenance perspective, I think it would be better to have the taint and concolic execution capabilities separated if possible.

Taint2 can use quite a bit of memory already, so I was wondering, how does including `z3::expr` in the `TaintData` struct affect memory usage? This might be another reason to separate the capabilities.

This seems very cool!

buszk:
The extra memory might actually be concern.

To clarify, I augmented TaintData with two z3::expr pointers. I am not sure how much of PANDA’s memory consist of shadow memory for TaintData though. Might be a big portion.

I think the solution can be simply move the two pointers to the existing LabelSetP data structure. Lastly, I do expect all the symbolic data tainted as well, because memory read with symbolic addresses isn’t well supported.

PR #889, opened 2021-02-27
Dynamic Symbols for non-native endian guests

AndrewFasano:
Dynamic symbols: Flip pointers read out of guest memory where necessary if host and guest have different endianness.

OSI_linux: Fix bug in calculating length of `dentry` names with non-native guest endianness.


AndrewFasano:
The example script doesn't seem to work for mips, but the one I added to the tests directory (which is more simple) does. It seems like dynamic_hooks isn't seeing libc get loaded on mips. I'm not sure if there's something weird going on with that qcow or if there's a bug with dynamic symbols on mips.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
```diff
@@ -463,6 +494,10 @@ void hook_program_start(CPUState *env, TranslationBlock* tb, struct hook* h){
 }
 
 void btc_execve(CPUState *env, TranslationBlock *tb){
+    if (unlikely(!panda_is_callback_enabled(self_ptr, PANDA_CB_BEFORE_BLOCK_EXEC, pcb_btc_execve))) {
+        // Callback still runs occasionally after disabled
+        return;
+    }
```
AndrewFasano, line 500: Note this is weird. But sometimes the callback is called, then goes to disable itself and we get an assert because it was already disabled. Maybe there's a race condition in here?

lacraig2: So it looks like there were a few instances where BEFORE_BLOCK_EXEC should have been BEFORE_TCG_CODEGEN. Could that have been the thrown assert?

lacraig2:
I was playing with this on MIPS with the default image. For some reason it's picking up a bunch of libraries, but not libc. Worth investigating.

panda/plugins/dynamic_symbols/dynamic_symbols.cpp
AndrewFasano: I suspect that was the issue

AndrewFasano:
So with the BTC change does this look good to merge?

lacraig2:
Yep. Looks good to me.

PR #892, opened 2021-03-04
Build PANDA with LLVM 11 instead of LLVM 10

LauraLMann:
Note that I could not find a PPA to install LLVM 11 in Ubuntu 19, so support for building under Ubuntu 19 was dropped.


AndrewFasano:
Looks fantastic, thanks for the PR! I'll merge when the CI finishes

PR #908, opened 2021-03-19
PyPANDA: warn if missing call to end_analysis

AndrewFasano:
When a script has finished executing all the queued up, blocking functions that drive guest execution but none of them have called `panda.end_analysis()` print a warning.


lacraig2 (approved):
It looks good. I tested it out. I was thinking through side-effects, but I believe `end_analysis` should be the global end of the emulation so we shouldn't need to reset at some point.

AndrewFasano:
I'd disagree with that - end_analysis is how you end a running analysis, you could then start another. Stop run is an old/internal function nobody should call IMO

AndrewFasano:
This PR appears to be flawed. I'll open an issue later, but if you do time.sleep(30) in a blocking function, you'll get the warning even though the thread is still running. Not sure exactly what's going on because the asyncthread.py should be blocked on the [func() call ](https://github.com/panda-re/panda/blob/f0b4a30bfcdd348160424d54b5d91d47d2e2fbc8/panda/python/core/pandare/asyncthread.py#L91) but it's somehow running the loop again. 

Maybe it's related to the `internal=True` async thread? Not sure exactly how we implemented that

PR #909, opened 2021-03-19
Support Rust plugins via submodule in pandadev container

tnballo:
- Add out-of-tree [Rust plugins](https://github.com/panda-re/panda-rs-plugins) via submodule.
- Update `pandadev` container to install the Rust toolchain, build Rust plugins as shared objects, and copy them into the build directory so that they can be invoked via the command line (as if they were regular C/C++ plugins).


panda/dependencies/ubuntu:18.04_build.txt
```diff
@@ -24,6 +24,8 @@ protobuf-c-compiler
 protobuf-compiler
 python3-dev
 zip
+curl
+sudo
```
AndrewFasano, line 28: Shouldn't need to install SUDO in the container - there's only the root user

Dockerfile
```diff
@@ -50,6 +55,15 @@ RUN git -C /panda submodule update --init dtc && \
...
     make -C /panda/build -j "$(nproc)"
 
+# Rust plugin build
+ENV PANDA_PATH "/panda/build"
+RUN echo "export PANDA_PATH=\"/panda/build\" >> /root/.bashrc"
+RUN git -C /panda submodule update --init panda-rs-plugins
+WORKDIR /panda/panda-rs-plugins
+RUN ./install_plugins.sh
+# Cleanup after IL plugin (uses a differnt captstone version during build)
+RUN sudo apt-get -qq update && DEBIAN_FRONTEND=noninteractive apt-get install -y libcapstone3 libcapstone-dev
```
AndrewFasano, line 65: Shouldn't use sudo in container since this runs as root

AndrewFasano:
Looks good to me other than 3 sudo things- once you remove those I'm happy to merge.

panda/dependencies/ubuntu:20.04_build.txt
```diff
@@ -20,6 +20,8 @@ protobuf-compiler
 python3-dev
 libpixman-1-dev
 zip
+curl
+sudo
```
AndrewFasano, line 24: here too

AndrewFasano:
Also now that you've done this for panda-rs I think we should probably do something similar for PyPANDA. It would be nice to move that to out-of-tree

AndrewFasano:
Actually I want to suggest a couple of changes, let me see if I can figure this out...

Dockerfile
tnballo: Done.

panda/dependencies/ubuntu:18.04_build.txt
tnballo: This is needed to support scripts in the `panda-rs-plugins` repo that also run without Docker, specifically [this script](https://github.com/panda-re/panda-rs-plugins/blob/master/install_plugins.sh) gets called and then calls [this one](https://github.com/panda-re/panda-rs-plugins/blob/master/panda-il-trace/setup.sh).

These scripts use `sudo` when run on baremetal for non-container users, so having `sudo` in the container means the `Dockerfile` can invoke them as well!

tnballo:
Looks like CI is failing due to SSH for a submodule init, guess submodules have to be HTTPS? Don't get the same failure locally.

tnballo:
Converting PR to draft until `capstone` discrepancy (4.0.2 built from source by Rust plugin vs 4.0.1 in Ubuntu 20.04 repo) is resolved.

PR #920, opened 2021-03-26
Attempt to fix asid panda_current_asid() when LPAE is enabled. #919

### LINKED ISSUE #919: ARM: ASID issue when LPAE is enabled
I've been investigating an issue where [panda_current_asid()](https://github.com/panda-re/panda/blob/master/panda/src/common.c#L97) returns a different value than what is returned by the asid_changed callback ([vmsa_ttbr_write()(LPAE)](https://github.com/panda-re/panda/blob/master/target/arm/helper.c#L2578) or [vttbr_write()(non-LPAE)](https://github.com/panda-re/panda/blob/master/target/arm/helper.c#L2596)). This appears to impact systems that enable the LPAE (Large Physical Address Extension) MMU (including aarch64). I am not an expert in ARM page tables. I've just been stumbling through QEMU, Arm's reference manual, and Linux's ARM MMU source.

This comes down to [arm_get_vaddr_table()](https://github.com/panda-re/panda/blob/master/panda/src/common.c#L60) which was copied from [get_level1_table_address()](https://github.com/panda-re/panda/blob/master/target/arm/helper.c#L7332). If LPAE is enabled, this function incorrectly masks the ttbr register with `tcr->base_mask`, which is not used in LPAE mode. If you follow the logic within `get_phys_addr()`, you will find that the stage1 lookup is different if LPAE is enabled vs armv5/6.

```
    if (regime_using_lpae_format(env, mmu_idx)) {
        return get_phys_addr_lpae(env, address, access_type, mmu_idx, phys_ptr,
                                  attrs, prot, page_size, fsr, fi);
    } else if (regime_sctlr(env, mmu_idx) & SCTLR_XP) {
        return get_phys_addr_v6(env, address, access_type, mmu_idx, phys_ptr,
                                attrs, prot, page_size, fsr, fi);
    } else {
        return get_phys_addr_v5(env, address, access_type, mmu_idx, phys_ptr,
                                prot, page_size, fsr, fi);
    }
```

The first stage lookup is a bit more complicated for LPAE as you can see in [get_phys_addr_lpae](https://github.com/panda-re/panda/blob/master/target/arm/helper.c#L7807). The first stage can depend on either `TTBR0` and `TTBR1`. Specifically, it depends on the fields `t0sz` and `t1sz` of the TTBCR register. It is also slightly different for aarch64.

In practice, it appears like Linux just uses TTBR0 for the 1st stage lookup on EL0; however, I can't speak for other OS's. Therefore, panda could probably just get away with using TTBR0 as the "asid".

Also, note that the code within [vmsa_ttbr_write()](https://github.com/panda-re/panda/blob/master/target/arm/helper.c#L2578) for LPAE is somewhat misleading; however, it works for Linux. TTBR0 is actually a 64-bit register for 32-bit Armv7 architectures. There are 40 bits the address and an 8-bit ASID is in bits 55-48. It's accessed with 2 32-bit registers using the instruction `MRRC p15, 0, <Rt>, <Rt2>, c2`. Refer to [Arm's V7 reference manual](https://developer.arm.com/documentation/ddi0406/latest/) section
B4.1.154 (page 1723) for details. In the call to `panda_callback_asid_changed()`, it is cast to a 32-bit value (target_ulong). In practice, I don't think this matters because Linux only appears to uses 32 bits for the base address.

The TTBCR register is handled differently in LPAE mode. It uses the Long Descriptor format (pg 17178) by setting the EAE bit. In Armv5/6 mode it uses the short descriptor, which contains the base_mask. If you wanted to make this work in general, vmsa_ttbr_write() would need to account for `t0sz` and `t1sz`.

To match the behaviour in `vmsa_ttbr_write()`, I believe `arm_get_vaddr_table()` can be changed with this:

```
    if (mmu_idx == ARMMMUIdx_S12NSE0 || mmu_idx == ARMMMUIdx_S12NSE1) {
        mmu_idx += ARMMMUIdx_S1NSE0;
    }

    if (regime_using_lpae_format(env, mmu_idx)) {
        *table = regime_ttbr(env, mmu_idx, 0);
    } else {
        /* Note that we can only get here for an AArch32 PL0/PL1 lookup */
        TCR *tcr = regime_tcr(env, mmu_idx);

        if (address & tcr->mask) {
            if (tcr->raw_tcr & TTBCR_PD1) {
                /* Translation table walk disabled for TTBR1 */
                return false;
            }
            *table = regime_ttbr(env, mmu_idx, 1) & 0xffffc000;
        } else {
        ...
    }
    ...
```

References:
* [Writeup on the ARM32 page tables](https://people.kernel.org/linusw/arm32-page-tables)
* [Arm's V7 reference manual](https://developer.arm.com/documentation/ddi0406/latest/)
* Relevant kernel sources:
  * Setting TTBR0 with VMSA VTTBR when LPAE is enabled: https://elixir.bootlin.com/linux/latest/source/arch/arm/mm/proc-v7-3level.S#L45
  * Setting TTBR0 with VTTBR when LPAE is disabled: https://elixir.bootlin.com/linux/latest/source/arch/arm/mm/proc-v7-2level.S#L41
  * AArch64: https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/asm/mmu_context.h#L234

bellma:
This doesn't resolve #919 in for general cases, but I believe it fixes it for Linux.


AndrewFasano:
Thanks for the fix!

PR #926, opened 2021-03-29
PyPANDA example cleanup

### LINKED ISSUE #924: PyPANDA 's examples ` osi_syscalls.py`  ` Memory of size 100 does not evenly divide into 8 byte chunks`  ERROR
The virtual_memory_read  memory of size must be aligned.

panda/panda/python/examples/osi_syscalls.py
```python
@panda.ppp("syscalls2", "on_sys_execve_enter")
def on_sys_execve_enter(cpu, pc, fname_ptr, argv_ptr, envp):
    # Log commands and arguments passed to execve unless in kernel
    if panda.in_kernel(cpu):
        return
    try:
        fname = panda.read_str(cpu, fname_ptr)
        argv_ptrlist = panda.virtual_memory_read(cpu, argv_ptr, 100, fmt='ptrlist')  
    except ValueError: return
    argv = []
    for ptr in argv_ptrlist:
        if ptr == 0: break
        try:
            argv.append(panda.read_str(cpu, ptr))
        except ValueError:
            argv.append(f"(error: 0x{ptr:x})")

    print(get_calltree(cpu) + " => " + ' '.join(argv))
```

AndrewFasano:
1) Use panda.queue_blocking over panda.queue_async everywhere (prior
   code was wrong - could have decorated f with @panda.blocking and then
   queued with panda.queue_async(f), not applied @panda.queue_async as a
   decorator to f.
2) Consistently use the variable name `cpu` for the CPUState object
   being returned from python (this object contains env_ptr which could
   then be stored as the variable `env` but prior code was confusing)

Also fixes #924 because that was just a little change in an example.


lacraig2 (approved):
Looks good.

PR #941, opened 2021-04-01
loaded_libs minor bugfix

tnballo:
Sorry my editor config removes trailing spaces - good for file size but bad for diff readability. 

Minor bugfix in `loaded_libs` - force at least one sample.

Edit:

Regarding sampling determinism across replays - thought there was a bug when `random()` is not explicitly seeded, but [per the man page](https://man7.org/linux/man-pages/man3/random.3.html) it gets seeded with `1` automatically. So the explicit `srandom()` call in `init_plugin()` isn't actually necessary - but might be clearer.

### DIFF
--- a/panda/plugins/loaded_libs/loaded_libs.cpp
+++ b/panda/plugins/loaded_libs/loaded_libs.cpp
@@ -1,5 +1,6 @@
 #define __STDC_FORMAT_MACROS
 
+#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


tnballo:
Looked through the diff, the changes made in `f0be65b` fix the same issue as this PR and then some! So it superceeds: no need to merge this PR, closing.

PR #950, opened 2021-04-09
dynamic symbols changes; tcg hooks changes; proc_start_linux changes.

lacraig2:
- Spruced up algorithms in dynamic_symbols
- Fixed subtle TCG bug in hooks. Processes share blocks a lot so you must insert even if ASID is wrong.
- Added `program_header` to auxiliary values struct to make it easier to have for offsets.


panda/python/examples/proc_start.py
```diff
@@ -52,4 +52,14 @@ def rec_auxv(cpu, tb, av):
     print(f"[bold blue]AT_PLATFORM:     {av.platform}[/bold blue]")
     print("[bold magenta][END PyPANDA on_rec_auxv][/bold magenta]")
 
+    @panda.hook(av.entry)
+    def hook(cpu,tb, h):
+        try:
+            print(panda.virtual_memory_read(cpu,av.program_header,4))
+            print('this should read "\\x7f7FELF" (though sometimes it wont if its not in memory)')
```
AndrewFasano, line 59: Is this print supposed to be here or is it just for debugging?

AndrewFasano:
In the future it would be awesome if these PRs could be split up into more logical chunks - i.e., a different PR for each of the bullet points (and if a PR depends on another being merged first that's fine to note). Or you could just use git rebase to clean up your commits into 3 distinct commits, one for each of the features.

As is, PRs like this are kind of awkward to merge (either we merge/rebase all 12 commits onto `master` and end up with a bunch of messages and diffs only you would care about, or we squash them all into a single commit in which case 3 distinct features end up as 1 commit).

panda/python/examples/proc_start.py
lacraig2: It is. It gives a sense of what you should expect from the program. Sometimes it wont be loaded and it explains that outcome too.

lacraig2:
I certainly agree and aim to have smaller PRs. That said the `hooks`, `dynamic_symbols`, and `proc_start_linux` plugins share a lot of structures and changing part of one results in changing a lot of other parts.

As these things go I'd rather squash it. Two of those are very linked. The last is a fairly small change.

PR #959, opened 2021-04-15
pypanda: add a bunch of types

lacraig2:
Add type information to pypanda (ideally without breaking anything) to every user-facing function and most internal functions. 

NOTE: Not everything is typeable. For example, CPUState struct as CData isn't.


lacraig2:
Don't merge just yet. There's some types i missed that should be optional instead.

lacraig2:
Should be ready to go

AndrewFasano:
If you want to merge this I'd be happy to once the conflicts are resolved

lacraig2:
Ready to merge as far as I can tell.

AndrewFasano:
Can we incorporate typechecking in the CI after this? Something like
```
mypy --python-version 3.7 --ignore-missing-imports panda.py
```

Seems like that currently returns a bunch of errors - not sure if we want to resolve those before merging (like today?) - or we can merge and incorporate this later. Either way, let's merge today so this doesn't go stale!

PR #965, opened 2021-04-16
Adding task_start and task_end callbacks to OSI.

bellma:
Unsure if there is much interest but I added callbacks for when a task starts and ends in OSI. This change hooks  [wake_up_new_task()](https://elixir.bootlin.com/linux/v5.12-rc7/source/kernel/sched/core.c#L3819) for starting tasks and 
[free_task()](https://elixir.bootlin.com/linux/v5.12-rc7/source/kernel/fork.c#L443) for ending tasks. 

This is marked as a draft because I've only tested on an arm-32 system, and the documentation for OSI would need to be updated. Specifically, `get_arg0()` should be tested on other systems. Also, `free_task()` was introduced in 2.5.71, and `wake_up_new_task()` was introduced in 2.6.9, so if support below these versions is desired, I need to find an equivalent functions to hook.

### DIFF
--- a/panda/plugins/osi/os_intro.c
+++ b/panda/plugins/osi/os_intro.c
@@ -47,6 +47,8 @@ PPP_PROT_REG_CB(on_get_process_pid)
 PPP_PROT_REG_CB(on_get_process_ppid)
 
 PPP_PROT_REG_CB(on_task_change)
+PPP_PROT_REG_CB(on_task_start)
+PPP_PROT_REG_CB(on_task_end)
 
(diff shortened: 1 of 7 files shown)


AndrewFasano:
Sorry for the slow response, I saw it was a draft and just assumed it wasn't ready. I like this idea and think this fits in well with OSI. This would inform us when tasks are initially started but also when they're switched to later, right? Seems like a very useful feature to me!

@lacraig2 what do you think? This is similar to `proc_start_linux` but not duplicative as the callbacks on the task getting woken back up are new. And I like that it's in OSI.

PR #971, opened 2021-04-21
dynamic_symbols: fixup entrypoint if it's in ARM's thumb mode.

bellma:
For ARM architectures, the `blx` instruction can use bit 0 to change into thumb mode (https://developer.arm.com/documentation/dui0068/b/ARM-Instruction-Reference/ARM-branch-instructions/BLX). The hook for the entrypoint needs to mask off this bit.


lacraig2:
That's a great catch @bellma! Thanks!

PR #977, opened 2021-04-28
Fix ioctl print error

tnballo:
* Small fix for training


AndrewFasano:
Sorry I broke this. I was thinking we always wanted `panda.ffi` over `ffi` but I was wrong.

PR #985, opened 2021-05-04
Add TCG callbacks to start and end of blocks

lacraig2:
Introduces two new callbacks:
- `start_block_exec` which is a TCG level callback at the start of blocks
- `end_block_exec` which is a TCG level callback at the end of blocks

This is a fairly early draft I'm putting out to get thoughts on.


nathanjackson:
Sorry @lacraig2 I've been doing some non-PANDA specific work lately and haven't been in the Slack to see your message from last week.

I'm definitely in favor of having these. Technically speaking, the "before" and "after" callbacks were actually a bit misleading because a block could abort execution.

The only pitfall I see here is that I think there can be more than one exit from a TCG block (that is, more than one `INDEX_op_exit_tb`). Perhaps we should insert the exit callback for each `exit_tb`.

lacraig2:
@nathanjackson that's interesting. I'll take a look at doing that. Thanks!

PR #1000, opened 2021-05-12
hooks performance optimization

fengjian:
copy to reference


lacraig2:
That's helpful. Though two of those functions aren't actually used anymore when we switched to sets.

Issue #994, opened 2021-05-07
panda_current_asid incorrect for mips


AndrewFasano:
The implementation of panda_current_asid for mips calculates the asid with
```
env->CP0_EntryHi & env->CP0_EntryHi_ASID_mask
```

which seems like it should be correct to me. However, I'm seeing this value change very frequently even within a single process. Using OSI_linux I tried querying the asid field from the OsiProc object and that contains a consistent ASID for each process. So it seems like the Linux kernel is able to identify that the same process is running while `panda_current_asid` is incorrectly saying a process has changed.

I found [this StackOverflow post](https://stackoverflow.com/questions/31948901/getting-pointer-to-the-current-translation-table-on-mips-linux) which is the exact thing we're trying to do and read the referenced chapter of the "See MIPS run" book which describes a portion of the `context` register in CP0 ($4) as:
> Context (PTEBase): This field just stores what you put in it. To exploit the register as its designers intended, write in the high-order bits of an (appropriately aligned) starting address of a memory-resident page table. The starting address must be picked to have zeros in bits 22 and downward—that’s a 4-MB boundary. While it would be grossly inefficient to provide that alignment in physical or unmapped memory, the intention is that this table should be put in the kseg2 mapped region. See below for how this works.

From the StackOverflow post, it sounds like the PTEBase field could be used as an ASID - but this value didn't seem to work as I hoped (with `cpu.env_ptr.CP0_Context & 0xff0000`).

Below is the script I was testing with:

```py
from pandare import Panda

panda = Panda(generic="mips")

asids = {}

@panda.queue_blocking
def driver():
    panda.revert_sync("root")

    @panda.cb_before_block_exec
    def bbe(cpu, tb):

        if panda.in_kernel(cpu):
            return

        # TODO: figure out a better way to get the asid
        asid = panda.current_asid(cpu)
        #asid = cpu.env_ptr.CP0_Context & 0xff0000

        if asid not in asids:
            proc = panda.plugins['osi'].get_current_process(cpu)
            name = "err"
            pid = "?"
            pasid = 0
            if proc != panda.ffi.NULL:
                name = panda.ffi.string(proc.name).decode()
                pid = proc.pid
                pasid = proc.asid


            asids[asid] = {'name': name, 'pid': pid, 'count': 0, "osi_asid": hex(pasid), "panda_asid": hex(panda.current_asid(cpu))}
        asids[asid]['count'] += 1

    print(panda.run_serial_cmd("ps aux | grep ps"))

[...]

PR #1001, opened 2021-05-12
Fixed MIPS ASID bug in OSI

off-by-1-error:
This fixes issue #994  where ASIDs retrieved from `OsiProc` structs were inconsistent with ASIDs retrieved from `panda_current_asid()` with a MIPS guest. ASIDs from OSI now match those from PANDA. The implementation of `panda_current_asid()` is correct, and the ASID changing over the course of a process' execution is [normal MIPS behavior](https://mail-index.netbsd.org/port-mips/2002/11/17/0005.html)


AndrewFasano:
So I'm still a little confused by this - is there an asid stored in the task_struct for MIPS? I can see how using a cached value for that would be incorrect given that a process will have its asid change often. So given that the cached value is incorrect is it just an optimization to read it out of `CP0_EntryHi` or is this change something more than that?

PR #1003, opened 2021-05-13
syscalls2 Change to SBE

lacraig2:
Change syscalls2 to SBE.

Merges after SBE PR.


AndrewFasano:
Rebased onto `master` and force pushed after merging #985 since I squashed those commits and they were also on this branch

PR #1008, opened 2021-05-15
PyPanda Panda.hook: Remove duplicate call to hook_cb_type

pcworld:


AndrewFasano:
Thanks for this PR! Our CI is down for another few hours but I'll merge once it's back.

PR #1016, opened 2021-06-04
Add PANDA API to inject page faults

AndrewFasano:
Currently only supports i386. I'll try adding other architectures before merging anything.

### DIFF
--- a/panda/python/examples/fault.py
+++ b/panda/python/examples/fault.py
@@ -1,0 +1,60 @@
+from pandare import Panda
+
+panda = Panda(generic="i386")
+panda.load_plugin("syscalls2", {"load-info": True})
+
+@panda.queue_blocking
+def drive():
+    panda.revert_sync('root')
+    print(panda.run_serial_cmd("md5sum $(which whoami); find /etc/ | md5sum"))
+    panda.end_analysis()
+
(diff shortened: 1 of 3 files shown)


panda/python/examples/fault.py
```diff
@@ -0,0 +1,60 @@
...
+                print(f"{addr:#x} => {repr(panda.read_str(cpu, addr))}", end="")
+
+        print(sep, end="") # , or )
+    else:
+        print()
+
+@panda.ppp("syscalls2", "on_all_sys_return2")
+def all_ret(cpu, pc, call, rp):
+    rv = panda.arch.get_return_value(cpu)
+    print(f"\t\t==> {rv:#x}")
```
pcworld, line 55: This code seems to be quite useful to trace syscalls. I wonder whether this could be made reusable and put somewhere other than the examples directory, so that projects will not have to copy&paste it?

AndrewFasano: If you're just interested in the syscalls tracing, there's a `syscalls_logger` plugin merged into `dev` now. If you're more interested in the page fault resolution, that's a bit harder - this PR has gone stale because my code to inject page faults seems pretty broken.

PR #1018, opened 2021-06-10
PyPanda: Panda.callstack_callers: Arch-dependent type

pcworld:
callback_instr uses target_ulong for the first parameter of
get_callers().
This commit makes the type dependent on panda.bits so that it works on
architectures other than 32 bits.

--

Note that I have only tested this on x64. I don't know whether there are architectures in which ulong_target is different to what my new code does.


AndrewFasano:
Thanks for the fix!

PR #1022, opened 2021-06-21
Backport qemu fixes to support recent glusterfs

pcworld:
Note that I have not tested glusterfs functionality, but backporting these commits at least fixes compilation issues with Panda on Fedora 34.

PR #1027, opened 2021-06-21
panda/plugins/{callstack_instr,taint2}: Support recent toolchain

pcworld:
Fixes compilation on Fedora 34

---

Together with PR #1022, #1023, #1024, #1025, #1026, this makes it possible to compile Panda on Fedora 34 (which has very recent software) with the following command, after installing the appropriate dependencies from the distribution's repos:

`LLVM_CONFIG_BINARY=llvm-config-11-64 ../build.sh --python x86_64-softmmu  --disable-werror --disable-xen --extra-cxxflags=-I/usr/include/z3`

There are still some warnings which would need to be fixed to get rid of ``--disable-werror``.
panda's xen module is not compatible with recent Xen headers. I have not bothered to find the fixes from upstream qemu, thus I just disable Xen support (don't know what it is needed for on Linux anyway).
Regarding z3 (which is used by the taint2 plugin), the z3 headers on Fedora reside in /usr/include/z3. I have tried a workaround with `__has_include` like in PR #1025, however this does not work because /usr/include/z3/z3++.h on Fedora 34 has an `#include<z3.h>` statement (as opposed to `"z3.h"` as in other headers in the same directory), which might be considered a bug in the Fedora package.

Some of these PRs will probably also be helpful when panda will want to support newer Ubuntu releases.

### DIFF
--- a/panda/plugins/callstack_instr/prog_point.h
+++ b/panda/plugins/callstack_instr/prog_point.h
@@ -14,6 +14,8 @@ PANDAENDCOMMENT */
 #ifndef __PROG_POINT_H
 #define __PROG_POINT_H
 
+#include <functional>
+
 // different ways to segregate the stacks
 enum stack_type {
     STACK_ASID = 0,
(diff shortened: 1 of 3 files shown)


AndrewFasano:
Something got messed up with our CI so I'm going to have to close and re-open all these PRs, sorry for the noise!

PR #1030, opened 2021-06-24
Install scripts bugfix: install pycparser

AndrewFasano:
A new CI test revealed that install_ubuntu wasn't working from a clean docker ubuntu image. This cleans up how that script was managing dependencies (by moving some pip dependencies into our list of apt packages) and adds some that were missing.

These changes don't really need to be backported to the stable branch as they don't affect any core PANDA code, just a helper script for the install process. But they should be incorporated in our next release.

### DIFF
--- a/.github/workflows/parallel_tests.yml
+++ b/.github/workflows/parallel_tests.yml
@@ -97,6 +97,7 @@ jobs:
       run: |
         docker system prune -af --filter "until=72h"
         docker builder prune -af --filter "until=72h"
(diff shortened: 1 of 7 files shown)


pcworld:
Btw, in case you prefer official Ubuntu packages over packages from pip, there's also `python3-pycparser` in Ubuntu, which could be added to the *build.txt files (untested).

AndrewFasano:
Oh that would be better, I'll go with that! Thanks so much. Sorry for leaving the CI broken before all your PRs

PR #1033, opened 2021-06-30
Fix hypercalls on ARM and remove warning that they are untested

collinmay:
The code that would cause a hypercall callback was getting short-circuited by `get_arm_cp_reginfo`'s failure. Moved the panda-specific code earlier in the function.


AndrewFasano:
That CI failure isn't related to your PR (sorry!), I'll fix and rerun

PR #1036, opened 2021-07-12
PyPanda ModeFilter: Fix example indentation

pcworld:
self is not available in the class scope


AndrewFasano:
Thanks for the fix!

Issue #858, opened 2021-01-21
Multiple Replays: Hangs for ARM


mariusmue:
Hi all,

I'm currently encountering issues when using the multi-replay API as demonstrated in PyPanda.
It seems running multiple replays within the same analysis works flawlessly on i386 guests, but becomes problematic on arm guests. More specifically, the guest becomes hanging after finishing the first replay.
I spent already multiple hours with this bug and could trace down *where* things go wrong, but I have no idea why.

In particular, the guest becomes hanging in [`main_loop_wait`](https://github.com/panda-re/panda/blob/master/util/main-loop.c#L531). More precisely, the problematic code is:

```C
int main_loop_wait(int nonblocking)
{
    int ret;
    uint32_t timeout = UINT32_MAX;
    int64_t timeout_ns;

    if (rr_in_replay()) {
        timeout = RR_MAIN_WAIT_LOOP_TIMEOUT_IN_REPLAY;
    }
    else if (nonblocking) {
        timeout = 0;
    }
    [...]
    if (timeout == UINT32_MAX) {
        timeout_ns = -1;
    } else {
        timeout_ns = (uint64_t)timeout * (int64_t)(SCALE_MS);
    }

    timeout_ns = qemu_soonest_timeout(timeout_ns,
                                      timerlistgroup_deadline_ns(
                                          &main_loop_tlg));

    ret = os_host_main_loop_wait(timeout_ns);
```

It seems after the end of a replay, the function gets called blocking, so that `timeout_ns` is set to `-1`  before calling `qemu_soonest_timeout`.
Hence, we need `timerlistgroup_deadline_ns` to return a value to be not `-1`, which is the case on i386.
`timerlistgroup_deadline_ns` holds (to my understanding) a global list of timers, and timerlistgroup_deadline_ns probably gets the value of the next timer to expired (not verified assumption).
However, on ARM, when we go down this codepath, all present timers seem to be set to `-1`, so that `os_host_main_loop_wait` gets called with `-1` as argument as well, leading to an unjustifiable long hang. I wrote a hackish solution, however, it feels like fighting symptoms instead of fixing the cause:

```C
    timeout_ns = qemu_soonest_timeout(timeout_ns,
                                      timerlistgroup_deadline_ns(
                                          &main_loop_tlg));

    /*  In some edge cases encountered in multi-replay runs with PANDA, both
     *  timeout_ns and the return value of timerlistgroup_deadline_ns end up
     *  being -1. If this happens, we would hang here.
[...]

PR #1037, opened 2021-07-15
Fix/multiple replay (#858)

mariusmue:
This is the hot fix for hangs in edge cases for multiple replays, as discussed in issue #858.
As I couldn't come up with a better solution and its still unpatched, I just provide this patch as-is, and also adjusted the multiple-replay python example in case this behavior needs to be tested.

Cheers,
Marius


AndrewFasano:
Thanks for the fix!

PR #1040, opened 2021-07-16
PyPanda Expect: Allow to turn off unansi

pcworld:
Not all shells use ANSI, and also unansi is still incomplete and thus
may sometimes corrupt output.
Add serial_kwargs to Panda() constructor to avoid cluttering the
constructor with new arguments in the future if Expect adds new
arguments.
Further make sure not to use mutable default parameters in Expect.

### DIFF
--- a/panda/python/core/pandare/panda_expect.py
+++ b/panda/python/core/pandare/panda_expect.py
@@ -19,7 +19,7 @@ class Expect(object):
     Designed to be used with the qemu monitor and serial consoles for Linux guests.
 
     '''
-    def __init__(self, name, filelike=None, expectation=None, logfile_base=None, consume_first=False):
+    def __init__(self, name, filelike=None, expectation=None, logfile_base=None, consume_first=False, unansi=False):
         '''
         To debug, set logfile_base to something like '/tmp/log' and then look at logs written to /tmp/log_monitor.txt and /tmp/log_serial.txt. Or directyl access
         '''
@@ -43,9 +43,8 @@ class Expect(object):
         self.update_expectation(expectation)
 
         # If consumed_first is false, we'll consume a message before anything else. Requires self.expectation to be set
-        self.consumed_first = True
-        if consume_first:
-            self.consumed_first = False
+        self.consumed_first = not consume_first
+        self.use_unansi = unansi
 
     def update_expectation(self, expectation):
         if isinstance(expectation, bytes):
@@ -386,8 +385,12 @@ class Expect(object):
                 # Translate the current_line buffer into plaintext, then determine if we're finished (bc we see new prompt)
                 # note this drops the echo'd command
                 if self.current_line.endswith(b"\n"):
-                    # End of line - need to unansi and move into prior_lines
-                    self.unansi()
+                    # End of line - need to potentially unansi and move into prior_lines
+                    if self.use_unansi:
+                        self.unansi()
+                    else:
+                        self.prior_lines.append(self.current_line[:-1].decode(errors='ignore'))
+                        self.current_line = bytearray()
 
                     # Now we have command\nresults..........\nprompt
                     #self.logfile.write(b"\n UNANSIs to: " + repr(self.prior_lines).encode()+b"\n")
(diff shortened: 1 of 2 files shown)


panda/python/core/pandare/panda_expect.py
```diff
@@ -386,8 +385,12 @@ def expect(self, expectation=None, timeout=30):
...
                 # note this drops the echo'd command
                 if self.current_line.endswith(b"\n"):
-                    # End of line - need to unansi and move into prior_lines
-                    self.unansi()
+                    # End of line - need to potentially unansi and move into prior_lines
+                    if self.use_unansi:
+                        self.unansi()
+                    else:
+                        self.prior_lines.append(self.current_line[:-1].decode(errors='ignore'))
+                        self.current_line = bytearray()
```
pcworld, line 393: I'm not completely sure whether this is sufficient in all cases, so feel free to take a careful look.

pcworld:
Also note that this disables unansi in `panda.monitor_console`. I assumed that qemu's monitor does not use ANSI codes when used over a socket, however I might just as well be wrong, in which case we would need to pass `unansi=True` there. (Or alternatively, make `unansi=True` the default in `Expect.__init__`.)

AndrewFasano:
Thanks for helping clean this up! ANSI codes are a pain and I'm not proud of the code that handles them.

I'll do some testing of this before merging, I feel like the qemu monitor was using ANSI but I'll double check that.

And good catch on the mutable default parameter!

AndrewFasano:
Looks like I was wrong, the QEMU monitor isn't doing any ANSI shenanigans. I do notice one difference though - but I think it's probably okay. Previously the lines of output from the qemu monitor were joined by `\n`s while now they're joined with `\r\n`s - do you think that's okay or should we find a way to strip the `\r`s?

Output from running `print(repr(panda.run_monitor_cmd("info registers")[:100]))`:

Old:
```
'EAX=00000020 EBX=c1d878a0 ECX=00000000 EDX=000003f8\nESI=f38b0000 EDI=c14e25b0 EBP=f5c35f38 ESP=f5c35'
```

New:
```
'EAX=00000103 EBX=c1d878a0 ECX=00000000 EDX=000003f8\r\nESI=f38b0000 EDI=0000000e EBP=f5c35f38 ESP=f5c3'
```

pcworld:
> ANSI codes are a pain and I'm not proud of the code that handles them.

I mean, it's cool that it exists.
Though it took me a while to find out *that* it exists (also the magic that attempts to remove the command string). Which is why I've now documented this a bit in Panda's constructor.

> Previously the lines of output from the qemu monitor were joined by `\n`s while now they're joined with `\r\n`s - do you think that's okay or should we find a way to strip the `\r`s?

I think that's okay. If people used Python's [str.splitlines](https://docs.python.org/3/library/stdtypes.html#str.splitlines) for splitting the lines, this would be transparent anyway.
As far as I've understood you're now following a (versioned?) release model with stable releases. Maybe we should introduce some changelog that lists breaking changes such as this? (Would also be relevant for #1041, give me a few days to respond in that issue.)

Btw, I think that removing the command string and unansi'ing might be considered different things (and thus it might make sense to put these behind different options). Though separating these two features would be a bit more work, and I personally don't need it so far, so I would leave it as is in this PR.

AndrewFasano:
Sorry, I imagine that was pretty confusing to track down! Sometimes we (foolishly) assume the features we need for working with our generic qcows are features everyone will always want.

We wrote up our plan for releases on the project wiki [here](https://github.com/panda-re/panda/wiki/PANDA-Branches-&-Versioning) - I think we're supposed to be including version numbers on the dev branch (whoops, we haven't been doing that) and building a changelog. I was thinking we'd just do that when we updated the stable branch. I'll add a changelog and version number in a few minutes and include an entry for this change.

I agree that removing the echo'd command and unansi'ing are different but I'm happy to leave that change for another time when someone actually needs it!

Thanks for all the recent PRs 😄!

AndrewFasano:
Update: I think I was wrong about the qemu monitor not using ANSI commands- see this error here:
https://github.com/panda-re/panda/pull/806/checks?check_run_id=3185790820#step:2:283. That said, this is the only place I've seen ANSI output from the monitor and I tested it a few times. I'll investigate a bit more.

AndrewFasano:
That link seems to have broken since I posted it- the monitor_cmds.py test failed once because it got what looks like an ANSI message back from the qemu monitor when it was expecting a string. I haven't been able to reproduce the error though and it's gone on subsequent runs of the CI.

PR #1045, opened 2021-07-20
Include core callbacks in pypanda documentation

AndrewFasano:
This shuffles some things around so we can show the PANDA callbacks in the autogenerated Python documentation and include arguments, return values, and descriptions of each.

The implementation is based off pulling comments from cb-defs.h and then creating a fake class pandare.Callbacks which has a function defined for each callback with a docstring based off the relevant comment.

This PR also fixes some documentation errors and typos I had found in that file (because they were messing with the parsing). Ultimately we might want to consolidate these comments in another location. If we merge this PR, we'll validate that these comments pass some basic checks (e.g., the names match) as part of our CI.

Finally, this PR removes the `ffi_importer` class which just stored a handle to the libpanda FFI object as a global. The class became unnecessary after some of the refactoring. The FFI object now is explicitly passed as an argument to the TaintQuery constructor since they still needed a reference to it.

image


AndrewFasano:
Despite the great improvement to the docs, this completely breaks pypanda so I guess it's not ready to merge. I'll investigate!

AndrewFasano:
Should be ready to go now, assuming CI passes 🤞

PR #1049, opened 2021-07-22
Pypanda examples cleanup to incorporate changes from #1045 - remove ffi imports

AndrewFasano:
Changes the `from pandare import ffi ... ffi.foo` pattern to skip the import and use `panda.ffi` 

Also changed some `from pandare import blocking ... @blocking` to use `@panda.queue_blocking` instead


lacraig2:
We really need pypanda tests.

This still generally breaks things.

```
luke@luke-panda:~/workspace/panda/panda/python/tests$ python3 taint_reg.py 
  File "/home/luke/workspace/panda/panda/python/tests/taint_reg.py", line 7, in <module>
    from pandare import Panda
  File "/home/luke/workspace/panda/panda/python/core/pandare/__init__.py", line 27, in <module>
    from .autogen.panda_datatypes import get_cb_docs
  File "/home/luke/workspace/panda/panda/python/core/pandare/autogen/panda_datatypes.py", line 15, in <module>
    C = ffi.dlopen(None)
NameError: name 'ffi' is not defined
```

AndrewFasano:
I think you have a stale panda_datatypes.py - can you rerun create_panda_datatypes.py? It should get ffi as a function argument now.

Do you think we could version panda_datatypes so that we could raise an exception for when we make changes to it? That way you'd get an error saying "Rerun create_panda_datatypes.py" instead of this

AndrewFasano:
@lacraig2 just pushed some code to raise a better exception. It's backwards compatible too. Whenever we modify how panda_datatypes.py is generated, we'll bump the `DATATYPES_VERSION` value that's put in there by create_panda_datatypes.py. Then in panda.py we check the version and raise an error if the datatypes needs to be regenerated. We could probably automatically regenerate the datatypes too, but that sounds like more work.

lacraig2:
You're right! Works now. Thanks.

lacraig2:
And adding a version to `panda_datatypes.py` is a good idea. Auto-regenerating is a tough place to be given how slow it is.

PR #1055, opened 2021-07-23
Add support for pypanda plugins

jamcleod:
Notable changes:
* Modifies PANDA (i.e. `panda-system-$arch`) to dynamically link against libpanda rather than statically link in the parts of PANDA that libpanda contains. This is needed such that pypanda can get a handle to it that is compatible with cffi/dlsym
* Adds `snake_hook`, a plugin to run pypanda scripts in a plugin-style manner (see [README](https://github.com/panda-re/panda/tree/python_plugins/panda/plugins/snake_hook#snake_hook) for more documentation on it)


panda/plugins/snake_hook/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        .unwrap()
+        .parent()
+        .unwrap()
+        .join(format!("libpanda-{}.so", ARCH));
+
+    if !plugin_self_path.exists() {
+        panic!("Plugin not found at '{}'", plugin_self_path.display());
+    }
+
+    // Reload self in order to
```
pcworld: in order to what?

panda/python/core/pandare/panda.py
```diff
@@ -163,7 +168,7 @@ def __init__(self, arch="i386", mem="128M",
         self.panda_args.extend(['-m', self.mem])
 
         # Configure serial - if we have an expect_prompt set. Otherwise how can we know what guest cmds are outputting?
-        if self.expect_prompt or (serial_kwargs is not None and serial_kwargs.get('expectation')):
+        if self.expect_prompt or self.expect_kwargs.get('expectation'):
```
pcworld: Either GitHub's diff is misleading or something went wrong during rebase, this shouldn't be changed

panda/plugins/snake_hook/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    }) {
+        // python error
+        Python::with_gil(|py| -> PyResult<()> {
+            Ok(python_err.print(py))
+        }).unwrap();
+    };
+
+    true
+}
+
```
pcworld, line 103: What happens on uninit? I don't really know how inline_python works. Does the Python interpreter stay alive?

panda/plugins/snake_hook/README.md
```diff
@@ -0,0 +1,30 @@
...
+    blocks = 0
+
+    @panda.cb_before_block_exec
+    def before_block_execute(cpustate, transblock):
+        nonlocal blocks
+        blocks += 1
+        print(f"Blocks run: {}", blocks)
+```
+
+The anatomy of a pypanda plugin in its current form (note: this is subject to change) is rather simple: all that is required is declaring an `init` function that takes an `panda` object, which is of type [pandare.Panda](https://docs.panda.re/panda.html#pandare.panda.Panda).
```
pcworld, line 28: In case you're intending to add other functions later on: I wonder if it would be better to let Python plugins globally declare a class rather than global functions. E.g. a PyPlugin class that is supposed to have an init() method (edit: rather than an init() method, I suppose a constructor call would suffice).
This would also allow to later add base classes for Python plugins, e.g. to define a [Python protocol](https://www.python.org/dev/peps/pep-0544/) if type hints are ever added to PyPanda, or an [abstract base class](https://docs.python.org/3/library/abc.html) for plugins to derive from.
Would also remove the need to use `global`/`nonlocal` for variables.

A candidate for a new function which might be added later would probably be an uninit() function to allow unloading Python plugins from the snake_hook plugin.

pcworld (commented):
Nice that this is being implemented.

I wonder, will this design allow extensions in the future such that other plugins (e.g, C or Rust plugins) can load Python plugins and communicate with them, like registering callbacks in a Python plugin?
As far as I understand it's not possible with the current code, which is fine of course, but I wonder whether this design will fundamentally allow to be extended in such a way in the future.

panda/plugins/snake_hook/src/lib.rs
jamcleod: Oops. Fixed

panda/python/core/pandare/panda.py
jamcleod: That is absolutely a mistake during the rebase, great catch!

panda/plugins/snake_hook/src/lib.rs
jamcleod: Good point! I hadn't considered the fact that snake_hook could be unloaded before end of execution.

As for what happens: the python interpreter stays alive (as the shared object can't be unloaded until snake_hook is due to being a shared object dependency), but isn't actively running.

However I believe the only thing that really needs to be handled currently (at least in the current design) is unloading all the callbacks.

panda/plugins/snake_hook/README.md
jamcleod: Yep! this is definitely the direction we intend on taking it. @lacraig2 and I have been discussing how we want to go about this. This PR is mainly about getting out an initial working version to build off of and experiment with, and is intentionally not trying to solve the bigger design questions (i.e. how the class should be structured, how python plugins fit into the greater scheme of things (plugin-to-plugin, should they be usable directly from `-panda` on command line, etc))

panda/plugins/snake_hook/src/lib.rs
jamcleod: I've added the following lines:

https://github.com/panda-re/panda/pull/1055/files#diff-a47498a30851524ea951adb169cac78887db05dd9a60a013e7833d03770753e2R106-R112

to handle unloading the callbacks on snake_hook uninit. In the future if/when we switch to a class-based plugin system, we'll run destructors here as well.

panda/plugins/snake_hook/README.md
```diff
@@ -0,0 +1,30 @@
+# snake_hook
+
+A plugin for running pypanda scripts in a plugin-style architecture. Unlike traditional pypanda usage, this allows PANDA to drive python, rather than python driving PANDA. This allows for a higher level of uncoordinated composability than is possible from using pypanda itself.
+
+### Usage:
+
+```
+panda-system-x86_64 -panda snake_hook:files=print_pcs.py:print_strings.py -nographic
```
AndrewFasano, line 8: According to our docs, plugin argument keys are separated by semicolons - https://github.com/panda-re/panda/blob/dev/panda/docs/manual.md#using-plugins. Is this a typo in the docs or did you implemented the arg parsing differently? I think it would be good to switch to semicolons for consistency.

AndrewFasano (approved):

PR #1056, opened 2021-07-23
Add check to see if Rust exists

jamcleod:


lacraig2 (approved):
Perfect. No notes. ❤️

PR #1058, opened 2021-07-27
Full TCG Syscalls

lacraig2:
Replace hooks return with a TCG hook at the end of the block. This works because the syscall should be the last instruction in the block. It translates:

```
instr a
instr b
syscall
```

into:
```
instr a
instr b
TCG hook enter
syscall
TCG hook return
```


lacraig2:
In addition to this being a draft I'll add that one issue that I know of is that the return addresses (checked in the return callback) will not match up appropriately. I don't have time for it at the moment, but if someone else wants to look at it I want that noted.

That said once this works I expect it to be very very fast.

PR #1059, opened 2021-07-28
Improve data collection for Pandelephant

AndrewFasano:
1) Update syscalls logger to collect return values, data from fixed-length buffers & port information from `bind` syscalls.
2) Create new plugin proc_trace to use OSI's `on_task_change` callback to identify when processes are started/switched to
3) Updates to pandalog_to_pandelephant to:
3a) support importing data collected by the `proc_trace` plugin
3b) handle the case where thread names are missing
3c) support skipping parts of the import process with CLI flags
3d) prefer pandare.plog_reader over the one in the local directory (these are duplicated scripts and the pandare version is newer than what we've left in panda/scripts which is where plog_to_pandelephant lives)

### DIFF
--- a/panda/plugins/config.panda
+++ b/panda/plugins/config.panda
@@ -31,8 +31,10 @@ pri
 pri_dwarf
 pri_simple
 proc_start_linux
+proc_trace
 recctrl
 replaymovie
+rust_skeleton
 scissors
 signal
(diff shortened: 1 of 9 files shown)


off-by-1-error (commented):
lgtm!

PR #1062, opened 2021-07-30
Rework pypanda plugins to use classes

jamcleod:
README changes (explains new pypanda plugin structure) [rendered here](https://github.com/panda-re/panda/blob/e17fcde39b136f4d366cfad86e164d0f4dd458b9/panda/plugins/snake_hook/README.md).


AndrewFasano (approved):
Looks good, thanks!

Issue #1048, opened 2021-07-22
undefined symbol: Z3_mk_seq_concat


celestial-d:
I am using the image from http://panda-re.mit.edu/qcows/linux/ubuntu/1804/x86_64/bionic-server-cloudimg-amd64-noaslr-nokaslr.qcow2 and attempt to use the file_taint on a simple example of just cating a file. What I get is the following output (undefined symbol: Z3_mk_seq_concat)

./panda/build/x86_64-softmmu/panda-system-x86_64 -m 4096 /home/duo/Downloads/bionic-server-cloudimg-amd64-noaslr-nokaslr.qcow2 -replay cat_test -panda file_taint:filename=/root/osi_linux/kernelinfo.conf -panda osi -os linux-64-ubuntu:my_info
PANDA[file_taint]:adding argument filename=/root/osi_linux/kernelinfo.conf.
PANDA[core]:os_familyno=2 bits=64 os_details=ubuntu:my_info
PANDA[core]:initializing file_taint
PANDA[core]:loading required plugin syscalls2
PANDA[core]:initializing syscalls2
PANDA[syscalls2]:using profile for linux x64 64-bit
PANDA[core]:loading required plugin hooks
PANDA[core]:initializing hooks
PANDA[core]:loading required plugin osi
PANDA[core]:initializing osi
PANDA[core]:loading required plugin osi_linux
PANDA[core]:initializing osi_linux
PANDA[osi_linux]:W> failed to read task.switch_task_hook_addr
PANDA[osi_linux]:W> kernelinfo bytes [20-23] not read
PANDA[core]:loading required plugin syscalls2
PANDA[core]:/home/duo/qemu/panda_tmp/panda/build/x86_64-softmmu/panda/plugins/panda_syscalls2.so already loaded
PANDA[core]:loading required plugin taint2
Failed to load /home/duo/qemu/panda_tmp/panda/build/x86_64-softmmu/panda/plugins/panda_taint2.so: /home/duo/qemu/panda_tmp/panda/build/x86_64-softmmu/panda/plugins/panda_taint2.so: undefined symbol: Z3_mk_seq_concat
PANDA[core]:FAILED to load required plugin taint2 from /home/duo/qemu/panda_tmp/panda/build/x86_64-softmmu/panda/plugins/panda_taint2.so
Aborted (core dumped)

[...]

PR #1064, opened 2021-08-04
[symbolic] Add CI support for basic symbolic tracing

buszk:
Previous PR #1061 introduced a regression that causes Segfault.
This PR should have it fixed and adds a basic CI test for symbolic tracing.

Add comments for the symbolic tracing functions to make them more readable

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -2051,6 +2051,29 @@ class Panda():
         expr_str = str(str_bs, 'utf-8')
         return self.string_to_expr(expr_str)
 
+    def taint_sym_path_constraints(self):
+        # Prepare ptr for returned string
+        somestr = b''
+        some_str_ffi = self.ffi.new('char[]', somestr)
+        str_ptr_ffi = self.ffi.new('char**', some_str_ffi)
+        # Prepare ptr for string size
+        n_ptr_ffi = self.ffi.new('uint32_t *', 0)
+
+        self.plugins['taint2'].taint2_sym_path_constraints(n_ptr_ffi, str_ptr_ffi)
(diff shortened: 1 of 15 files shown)


buszk:
Address #1048 
Download and install libz3 4.8.7 to `/usr/local/` on Ubuntu 18

panda/python/core/pandare/panda.py
```diff
@@ -2,6 +2,7 @@
 This module simply contains the Panda class
 """
 
+from ctypes import sizeof
```
AndrewFasano: Is there a way to do this with CFFI instead of adding the dependency on ctypes?

panda/python/core/pandare/panda.py
```diff
@@ -1975,17 +1977,118 @@ def taint_sym_label_ram(self, addr, label):
...
+        self.plugins['taint2'].taint2_sym_branch_meta(n_ptr_ffi, branch_meta_ptr_ffi)
+        # Unpack size
+        n = self.ffi.unpack(n_ptr_ffi, 1)[0]
+        if n == 0:
+            return []
+        meta_ptr = self.ffi.unpack(branch_meta_ptr_ffi, 1)[0]
+        metas_ffi = self.ffi.unpack(meta_ptr, n)
+        # Meta only has a pc field now
+        metas = [
+            hex(meta_ffi.pc)
```
AndrewFasano: Should this return as an int instead of a string? I.e., it could just be `meta_ffi.pc` here?

AndrewFasano (changes requested):

AndrewFasano (approved):

AndrewFasano:
Thanks for making those changes!

PR #1093, opened 2021-09-08
Fix Structs Definitions

### LINKED ISSUE #1080: PyPANDA: Some wrong offsets/sizes for for CPUX86State
There are currently two issues with the generated CFFI definitions for CPUX86State in PyPANDA:

1. QEMU uses zero-sized structs as markers (`start_init_save`, `end_init_save`, `end_reset_fields`). CFFI interprets these as being 1 byte, which throws off a bunch of other offsets
2. The definition of ZMMReg in `assumptions.h` is wrong; it should be `uint8_t ZMMReg[64];` (or even better a union of the correct types like in the original header)

I'm not sure what the best way to fix the first is (aside from a special-case hack to remove those three members), but the second should be pretty straightforward.

lacraig2:
- Change ZMMReg in assumptions.h
- Change run.py to not allow empty structs any longer
- Update struct files to reflect changes

CLOSES: #1080


AndrewFasano:
Nice work!

PR #1094, opened 2021-09-09
Added simpler method to monitor serial port for output and thus to sc…

tleek:
…ript behavior.

Added new function serial_read_until to panda (thx Andre!) that simply
monitors serial port until it sees a byte sequence. Used that to build
a script script that uses this to boot and then log in to guest and
then recreate "root" snapshot, something we had to do by hand
previously.


AndrewFasano:
If you want to change the unansi behavior for a specific panda object, you should create the pypanda object with it set in an argument:

```
panda = Panda(..., expect_kwargs={"unansi": False})
```

This PR was changing the default value for everyone so I switched it back. If that was an intentional change, we can talk about it more.

PR #1095, opened 2021-09-09
Add panda-gdb

jamcleod:
No support for aarch64 (no gdbstub-rs support) and PowerPC (no OSI)

The code is a bit grosser than on the other repo due to needing to feature gate off powerpc/aarch64.

If anyone knows how to keep PANDA trying to build for certain targets that'd probably clean things up a bit.

### DIFF
--- a/panda/plugins/config.panda
+++ b/panda/plugins/config.panda
@@ -12,6 +12,7 @@ dynamic_symbols
 edge_coverage
 filereadmon
 forcedexec
+gdb
 hooks
 hooks2
 libfi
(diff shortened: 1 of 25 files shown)


AndrewFasano (approved):
Looks fantastic, thanks for the PR! It would be awesome to get it working with live guests one day.

AndrewFasano:
The CI failures seem legitimate though - for aarch64 and ppc are you still having the plugin build and just raising a runtime error? I think that's the only way to do it with PANDA's current plugin design :(

PR #1100, opened 2021-09-28
fix exception handling

### LINKED ISSUE #1009: PyPANDA Panda.hook2/_hook_mem exception handling
On 27eb888ce2235debe4e7032a84ccc15d4eebde8a, Panda.hook2 and Panda._hook_mem use these FFI callbacks:
```py
@hook_cb_type # Make CFFI know it's a callback. Different from _generated_callback for some reason?
def wrapper(*args, **kw):
    return fun(*args, **kw)
```
and
```py
@mem_hook_cb_type # Make CFFI know it's a callback. Different from _generated_callback for some reason?
def wrapper(*args, **kw):
    return fun(*args, **kw)
```

Other similar FFI callbacks have special handling for exceptions, for example in Panda.hook:
```py
@hook_cb_type # Make CFFI know it's a callback. Different from _generated_callback for some reason?
def wrapper(*args, **kw):
    try:
        r = fun(*args, **kw)
        #assert(isinstance(r, int)), "Invalid return type?"
        return r
    except Exception as e:
        # exceptions wont work in our thread. Therefore we print it here and then throw it after the
        # machine exits.
        self.hook_exit_exception = e
        self.end_analysis()
        # this works in all current callback cases. CFFI auto-converts to void, bool, int, and int32_t
        return 0
```

I wonder whether Panda.hook2 and Panda._hook_mem would also need special exception handling.
If this is the case, maybe these wrapper functions should be refactored into one separate function to avoid such inconsistencies in the future.

lacraig2:
- Creates a single exception value to exit at
- Checks exception value before entering new code
- standardizes `_run_and_catch` across the 6 code interfaces

FIXES #1009

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -522,21 +522,9 @@ class Panda():
         #self.libpanda.panda_cleanup_record()
         if self._in_replay:
             self.reset()
-        if hasattr(self, "end_run_raise_signal"):
-            saved_exception = self.end_run_raise_signal
-            del self.end_run_raise_signal
-            raise saved_exception
-        if hasattr(self, "callback_exit_exception"):
-            saved_exception = self.callback_exit_exception
-            del self.callback_exit_exception
-            raise saved_exception
-        if hasattr(self, "blocking_queue_error"):
-            saved_exception = self.blocking_queue_error
-            del self.blocking_queue_error
-            raise saved_exception
-        if hasattr(self, "hook_exit_exception"):
-            saved_exception = self.hook_exit_exception
-            del self.hook_exit_exception
+        if hasattr(self, "exit_exception"):
+            saved_exception = self.exit_exception
+            del self.exit_exception
             raise saved_exception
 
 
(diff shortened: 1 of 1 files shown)


panda/python/core/pandare/panda.py
```diff
@@ -3088,8 +3101,7 @@ def decorator(fun):
 
             @hook_cb_type # Make CFFI know it's a callback. Different from _generated_callback for some reason?
```
AndrewFasano: You deleted this decorator above - but left it here. Not sure if that was intentional?

AndrewFasano:
Thanks for cleaning up this mess, code looks good to me

panda/python/core/pandare/panda.py
lacraig2: Good catch. Removed below.

AndrewFasano:
Should we merge this?

AndrewFasano:
I think I stomped on some of this code in a recent PR, I had thought this was already merged and fixed a bug in some of what this changes- sorry!

lacraig2:
Superseded by #1149

PR #1106, opened 2021-10-18
PyPANDA: add option to raise exceptions - use this mode from snake_hook

AndrewFasano:
Previously, PyPANDA would detect exceptions in ppp functions and callbacks and call `panda.end_analysis()` and then later print the exception after the guest has stopped.

However, when running PyPANDA scripts from snake_hook this would result in exceptions being silently swallowed and ignored.

### DIFF
--- a/panda/plugins/snake_hook/src/loader.rs
+++ b/panda/plugins/snake_hook/src/loader.rs
@@ -51,7 +51,7 @@ pub(crate) fn initialize_pyplugins(args: Args) {
     let context: Context = python! {
         from pandare import Panda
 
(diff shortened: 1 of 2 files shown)


jamcleod (approved):
Looks good to me. Worth noting the `cast_rc` name's meaning is unclear to me, but it's no worse than `f` so no nbd to me.

AndrewFasano:
Thanks. I changed f to cast_rc just to make it consistent with the other similar code, I think it's a cast.. return something? 🤷‍♂️

PR #1108, opened 2021-10-19
Standardize PyPlugin interface for rust/pypanda

AndrewFasano:
[Snake_hook](https://github.com/panda-re/panda/tree/dev/panda/plugins/snake_hook) defines a nice class-based API for creating PANDA plugins in pure Python aka **PyPlugins**.

This PR adds support for loading these PyPlugins through PyPANDA or the CLI (with the rust `snake_hook` plugin). In particular, it adds the `pyplugin` property to the pandare class and initializes it on the first access to be an instance of the `pandare.pyplugin`. By using the methods defined in this class, users can load and configure PyPlugins.

Plugins can be registered to run in one of three ways

## 1) Inline classes as PyPANDA plugins
```py

from pandare import Panda, PandaPlugin

panda = Panda(generic="x86_64")

class TestPlugin(PandaPlugin):
    def __init__(self, panda):
        path = self.get_arg('path')
        print(f"path = {path}")
        should_print_hello = self.get_arg_bool('should_print_hello')
        if should_print_hello:
            print("Hello!")

panda.pyplugin.register(TestPlugin, {'path': '/foo'})

@panda.queue_blocking
def drive():
    ...

panda.run()
```

## 2) Files as PyPANDA Plugins with PandaPlugin undefined
Per the original `snake_hook` spec, a plugin could subclass PandaPlugin without importing that object - as such special care must be used when loading these classes from a file. For now we support running with PandaPlugin undefined with the `panda.pyplugin.load_plugin()` APIs as shown below:

myplugin.py
```py
class MyPlugin(PandaPlugin):
    def __init__(self, panda):
        ....
```

runner.py
```py

from pandare import Panda

panda = Panda(generic="x86_64")

MyPlugin = panda.pyplugin.load_plugin_class("myplugin.py", "MyPlugin")
panda.pyplugin.register(MyPlugin, {'path': '/foo'})
# alternatively panda.pyplugin.register(("myplugin.py", "MyPlugin"), {'path': '/foo'})

@panda.queue_blocking
def drive():
    ...

panda.run()
```

## 3) Files as PyPANDA Plugins with PandaPlugin defined
myplugin.py
```py
from pandareimport PandaPlugin

class MyPlugin(PandaPlugin):
    def __init__(self, panda):
        ....
```

runner.py
```py

from pandare import Panda

panda = Panda(generic="x86_64")

from myplugin import MyPlugin
panda.pyplugin.register(MyPlugin, {'path': '/foo'})

@panda.queue_blocking
def drive():
    ...

panda.run()
```

This PR also adds a new script in the PyPANDA examples directory which uses the PyPANDA plugin flask option to show an updating basic block counter in a web app. It also updates an old example (strace.py) of using python classes for basic plugin functionality to use the pypanda plugin model.

If this PR is merged, it means that users should be able to create a single PyPANDA Plugin and load/configure it from either CLI panda (with snake_hook) or PyPANDA (with panda.pyplugin)

In the future, we might wish to consider renaming `snake_hook` to `pyplugin` for consistency.

PR #1106 should be merged prior to this (and then that commit should be dropped from here).

### DIFF
--- a/panda/python/core/pandare/extras/snake.py
+++ b/panda/python/core/pandare/extras/snake.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+
+"""
+Class to manage loading PyPANDA plugins of standard format
+which is compatable with the snake_hook plugin for CLI-panda
+"""
+
+class PandaPlugin:
+    def __init__(self, panda):
+        '''
+        Base class which PyPANDA plugins should inherit. Subclasses may
+        register callbacks using the provided panda object and use
+        PandaPlugin.get_args or PandaPlugin.get_arg_bool to check argument
+        values (note these helpers should be accessed with `self.get_...`).
+
+        This PyPANDA extra aims to mirror the snake_hook (rust) plugin
+        so a single PyPANDA plugin can be used from both pure PyPANDA (with this class)
+        or from CLI panda with snake_hook.
+
+        For more information, read the snake_hook documentation
+        '''
+        pass
+
+    def __preinit__(self, args):
+        self.args = args
+
+    def get_arg(self, arg_name):
+        '''
+        returns either the argument as a string or None if the argument
+        wasn't passed (arguments passed in bool form (i.e., set but with no value)
+        instead of key/value form will also return None).
+        '''
+        if arg_name in self.args:
+            return self.args[arg_name]
+
+        return None
+
+    def get_arg_bool(self, arg_name):
+        '''
+        returns True if the argument is True
+        '''
+        return arg_name in self.args and self.args[arg_name]==True
+
+
+class Snake:
+    def __init__(self, panda, flask=False, port=8080, host='127.0.0.1'):
+        '''
+        panda- a panda object
+        flask - a bool indicating whether to enable the flask server
+        port - a number (0-65535) indicating the port number to host the flask server at
+                (default: 8080)
+        host - host for flask server to listen on
+        '''
+        self.panda = panda
+        self.plugins = {}
+
+        # Flask specific vars
+        self.flask = flask
+        self.port = port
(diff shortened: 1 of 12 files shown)


panda/python/core/pandare/extras/snake.py
```diff
@@ -0,0 +1,226 @@
...
+        plugin = importlib.util.module_from_spec(spec)
+        plugin.PandaPlugin = PandaPlugin
+        spec.loader.exec_module(plugin)
+        if not hasattr(plugin, class_name):
+            raise ValueError(f"Unable to find class {class_name} in {plugin_file}")
+        cls = getattr(plugin, class_name)
+        assert issubclass(cls, PandaPlugin), f"Class {class_name} does not subclass PandaPlugin"
+        return cls
+
+    def register(self, pluginclass, args=None, name=None, template_dir=None):
```
jamcleod: If you're just going to check if `args` is `None` then set it to another default value, why not set the default value to `{}` in the named arguments?

panda/python/examples/bbc_www.py
```diff
@@ -0,0 +1,41 @@
...
+            return """<html>
+            <body>
+                <p>
+                    Basic Block Count: <span id="bb_count">""" + str(self.bb_count) +  """</span>
+                </p>
+            </body>
+            </html>"""
+
+s = Snake(panda, flask=True, host='0.0.0.0')
+s.register(BasicBlockCount)
```
jamcleod, line 32: As previously mentioned, changing this to `panda.snake_hook.register` would be nice.

jamcleod (approved):
Minor changes but after that looks good to me! Approving as I don't think it'll need re-review.

PR #1111, opened 2021-10-19
fix issues #1110

fengjian:
#1110


panda/python/core/pandare/panda.py
```diff
@@ -2377,7 +2377,7 @@ def copy_to_guest(self, copy_directory, iso_name=None, absolute_paths=False, set
 
         # Tell panda to we insert the CD drive
         # TODO: the cd-drive name should be a config option, see the values in qcow.py
```
jamcleod, line 2403: (See this TODO)

jamcleod (changes requested):
I think adding the ability to override the cdrom to use is great, although the mechanisms for making it "just work" in the case of generic qcows are mostly there. The steps I'd take to take to extend this to be more general:

1. in the `Panda` constructor, get the `cdrom` field from the qcow info provided by qcow.py, this is handled in the snippet below:

https://github.com/panda-re/panda/blob/4039432dbf77f5d34d8e879bf9216199fb4937ea/panda/python/core/pandare/panda.py#L106-L111

2. like above, store the cdrom field in something like `self.cdrom`, storing `None` if the user isn't using a generic qcow
3. in `copy_to_guest` change the default value of your optional argument to `None` rather than defaulting to "ide1-cd0"
4. use the aforementioned `None` values to implement fallback (e.g. if the named argument of `copy_to_guest` is `None`, fallback to `self.cdrom`. if that's `None`, fallback to the default value of "ide1-cd0")

Taking these steps would allow pypanda to pull the proper cdrom value from 'cdrom' in qcows.py:

https://github.com/panda-re/panda/blob/4039432dbf77f5d34d8e879bf9216199fb4937ea/panda/python/core/pandare/qcows.py#L71-L81

jamcleod:
Also @fengjian if you don't have the time or bandwidth to make the above changes feel free to let me know and I can try and find time to take care of it

fengjian:
> Also @fengjian if you don't have the time or bandwidth to make the above changes feel free to let me know and I can try and find time to take care of it

Your suggestion is very good, let me refactor the code.

jamcleod (approved):

PR #1112, opened 2021-10-19
Fix bug parsing hook_filter argument

LauraLMann:
It was parsing the hook_filter argument based on where the dash is in the PC argument - not good!


jamcleod (changes requested):
Nice catch.

Also looks like it needs changing in this error message as well:

https://github.com/panda-re/panda/blob/62109e294d5605c27c9a6d30e143c71466c1aaa3/panda/plugins/coverage/coverage.cpp#L233

Outside of that looks great!

LauraLMann:
Yes, good catch.  I have fixed the error message too.  Hopefully the checks will be happy with it too.

jamcleod (approved):

jamcleod:
Sorry this is taking a while, seems like adjustments to the CI box required restarting the workflow. it should get auto-merged after those pass.

PR #1115, opened 2021-10-21
Clean up PyPlugin model, add PPP support & simplify loading via snake_hook

AndrewFasano:
This PR updates PyPlugins (introduced in #1008) to support the PPP interface
1) Expose PPP-style callbacks to other PyPlugins and registering functions to run from them
2) Expose PPP-style direct function calls between PyPlugins

It also simplifies the `snakehook` plugin 
3) Consolidates code that loads PyPlugins by modifying snake_hook to use the panda.pyplugin (PyPluginManage) interface.

This also simplifies how the `snake_hook` plugin loads pyplugins by passing the arguments through to the PyPluginManager in the panda object.

In the long term, it would be nice to expand this interface to somehow allow C/rust plugins to interact with pyplugin-exposed PPP functions.

Note: this changes the names of some of the PyPlugin APIs previously introduced in #1108. In particular, it renames the various functions with `register` in the name to instead use `load` to keep the language consistent with the other plugin interfaces.


panda/plugins/snake_hook/src/lib.rs
```diff
@@ -2,13 +2,12 @@ use inline_python::python;
...
 use panda::prelude::*;
 use pyo3::prelude::*;
+use std::collections::HashMap;
 
 use std::ffi::{CStr, CString};
 use std::path::PathBuf;
-use std::sync::Mutex;
 
 mod loader;
-mod panda_plugin;
```
jamcleod, line 11: `src/panda_plugin.rs` should be removed too

panda/plugins/snake_hook/src/loader.rs
```diff
@@ -36,155 +31,85 @@ fn parse_files(files: &str) -> Vec<PluginFile> {
...
-                        }
-                    }
+                let file_path = std::fs::canonicalize(path).unwrap();
+                let file_path = file_path.to_string_lossy().into_owned();
+                let py_arg_list = args.into_py_dict(py);
+
+                if load_all_classes {
+                    // Load all the pyplugins in file_path
+                    context.run(python! {
+                            panda.pyplugin.load_all('file_path, args='py_arg_list)
```
jamcleod: inconsistent tabbing here

panda/python/examples/pyplugin_ppp.py
```diff
@@ -0,0 +1,46 @@
...
+        def do_add(x):
+            return x+1
+
+class Consumer(PandaPlugin):
+    '''
+    PyPlugin which defines a function to run when Server's `some_f` callback is triggered
+    '''
+    def __init__(self, panda):
+        self.ppp_reg_cb('Server', 'some_f', self.my_f)
+        print(f"Calling Server's do_add(1): ", self.call_ppp('Server', 'do_add', 1))
```
jamcleod: minor design nitpick but is there a reason for going with a string-based API instead of using forwarding objects? one downside of having them be string-addressed is that you cannot bind to the method/function.

So for example if we wanted to store a reference to `do_add`, how could we go about it?

jamcleod (approved):
Only found some minor things, the Rust code looks good to me!

panda/python/examples/pyplugin_ppp.py
AndrewFasano: I'm not sure how we could handle objects instead? `do_add` is undefined here and `Server` might not be in scope?

AndrewFasano:
I think I'm finally happy with this PR - I still plan to do a little documentation cleanup, but I think the PPP design is finalized.

panda/python/core/pandare/pyplugin.py
```diff
@@ -26,6 +70,19 @@ def __init__(self, panda, flask=False, host='127.0.0.1', port=8080, silence_warn
         if self.flask:
             self.enable_flask(host, port)
 
+    @property
+    def ppp(self):
```
AndrewFasano, line 43: Should add some documentation for this property - it's going to be a bit confusing with the dynamically created objects and we want something decent for pdoc

PR #1116, opened 2021-10-22
Add userdata context variable API to PPP callbacks

jamcleod:
See #1105 for additional context as to the rationale. As for the implementation--PPP callbacks were trickier and required a minor breaking change. The typedefs for PPP callbacks must now be generated using the `PPP_CB_TYPEDEF` macro. Unlike the base callback implementation, this implementation doesn't use trampoline generation and simply hold a separate callback list for those with context variables.


jamcleod:
Worth noting for anyone who reviews: only these first 3 files are particularly relevant. Most of the rest of the 15k line diff is just migrating the breaking change

image

panda/plugins/syscalls2/scripts/syscall_parser.py
```diff
@@ -268,7 +268,7 @@ def emit_temp_declaration(self):
         ''' Returns a snippet declaring an appropriate temp
             variable for this argument.
         '''
-        return '{0} arg{1};'.format(self.ctype, self.no)
+        return '{0} arg{1} = 0;'.format(self.ctype, self.no)
```
AndrewFasano, line 271: Can all these types be assigned to an actual 0 or do you want to initialize to an empty value with `{0}`?

AndrewFasano (changes requested):
Code changes look good to me (assuming CI passes) - but you should update the docs for PPP functions too so new plugins will be created with the right macros

AndrewFasano:
Pretty sure the CI failure is just a one-off-bug in the CI, not an issue with this PR. Merging 🙈

AndrewFasano (approved):

PR #1121, opened 2021-11-02
Add option to coverage plugin to exclude a range of PCs from instrumentation

LauraLMann:
A block is not instrumented for coverage if it is partially (or entirely) within a given range of PCs.
While unit testing, I discovered that the utility used to parse a start or end point from a range quietly parses a partial string instead of complaining.  As this is most likely due to a typo on the user's part, I enhanced the parsing utility to throw an error if only part of the start point or end point string is parsed.


jamcleod (approved):
Looks good to me!

PR #1124, opened 2021-11-08
Bugfixes for install_ubuntu.sh

ph4ge:
The minor changes proposed for the install_ubuntu.sh script concern:
- handling _curl_ dependency for Rust before trying to install it + making the shell environment aware of cargo install
- the _make install_ for capstone-4.0.2 requires elevated privileges in order to write to _/usr/lib_

### DIFF
--- a/panda/scripts/install_ubuntu.sh
+++ b/panda/scripts/install_ubuntu.sh
@@ -92,7 +92,7 @@ fi
 progress "Installing Rust..."
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 
-# expose cargo to the running shell/env
+# Expose cargo to the running shell/env
 . $HOME/.cargo/env
 
 # Because libz3-dev for Ubuntu 18 is really old, we download and install z3 github release v-4.8.7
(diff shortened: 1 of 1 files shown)


AndrewFasano:
Hey, these changes look great! I'll let the CI run and once it (presumably) passes, will happily merge. Thanks for the contribution!

ph4ge:
Hi Andrew,

I also noticed that if I didn't perform a complete  apt update && upgrade before the running install_ubuntu.sh
I have a dependency issue, i should have the package information somewhere:

The following packages have unmet dependencies:
 libepoxy-dev : Depends: libegl1-mesa-dev but it is not going to be installed
                Depends: libgl1-mesa-dev but it is not going to be installed
E: Unable to correct problems, you have held broken packages.

To recap, I observe the aforementioned message(libepoxy-dev) if i try launching the install_ubuntu.sh script without having performed a full update && upgrade before.

But I suppose everybody does this. 

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

AndrewFasano:
Hmm. The `apt update` happens automatically in the script which I think should normally set the subsequent apt commands up to succeed. Perhaps something had gotten messed up on your system prior to running the script? Or were you using a brand new system (or vm or container)? I worry that running `apt upgrade` automatically might surprise some users who didn't want other packages on their system upgraded.

ph4ge:
I used a freshly installed system in virtualbox.
I share your concern about the upgrade stuff, that's why I haven't touched it.
Anyway, the "modifications" don't touch/modify "core" stuff.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

AndrewFasano:
Hm, okay. I'll have to think about that then, not sure what would be the best way to fix that case then. For what it's worth, our [CI job that tests install_ubuntu.sh](https://github.com/panda-re/panda/blob/dev/.github/workflows/parallel_tests.yml#L19-L24) just starts from a clean ubuntu 20.04 container and runs the install_ubuntu script - not sure why that would behave differently from your system.

But we don't have to figure out exactly why your system needed the `apt upgrade` to merge this PR so I'll do that now :) Thanks again for the contribution

PR #1131, opened 2021-11-18
Fix webserver for snake_hook

jamcleod:


AndrewFasano:
Did we also want to fix the issue where if you enable flask it will try to call webserver_init even if that function isn't present?

But regardless of that, these changes look great

AndrewFasano (approved):

PR #1132, opened 2021-11-18
my local changes exposting get_file_struct_ptr

ryandmaggio:
I've edited osi_linux to add an exported function called "ext_get_file_struct_ptr" which is just a wrapper around "get_file_struct_ptr," which I needed access to for a plugin I was writing.


panda/plugins/osi_linux/osi_linux.cpp
```diff
@@ -558,6 +558,15 @@ char *osi_linux_fd_to_filename(CPUState *env, OsiProc *p, int fd) {
 }
 
 
+target_ptr_t ext_get_file_dentry(CPUState *env, target_ptr_t file_struct) {
```
AndrewFasano, line 561: The other exposed functions in this file are just called `get_...` think it would make more sense to drop the `ext_` prefix on these?

panda/plugins/osi_linux/osi_linux.h
```diff
@@ -697,4 +697,6 @@ void get_process_info(CPUState *cpu, GArray **out,
 }
 #endif
 
+
+
```
AndrewFasano: I don't think you meant to change this file

AndrewFasano (changes requested):
Congrats on your first PR! Definitely want to merge this but there a couple of little things that could make it better if you have time to address.

panda/plugins/osi_linux/osi_linux.cpp
AndrewFasano: Never mind - this is a wrapper for an inline function so it makes sense to keep the `ext_` prefix

AndrewFasano (approved):

PR #1138, opened 2021-12-02
Use libosi to handle windows introspection

### LINKED ISSUE #1085: Integrate new Windows introspection library
We have a new Windows introspection library we'd like to integrate.  This issue will track development.

The goal for the initial PR is to not break existing plugins, but this will leave PANDA with multiple introspection approaches.  A future issue will address changes required to consolidate to a single introspection library.

Ben-Dumas:
Use [panda-re/libosi](https://github.com/panda-re/libosi) library to handle windows introspection 

1. removes the existing `win*` plugins, and IMO simplifies the wintrospection plugin. 
2. splits out the windows-32-7 profile into windows-32-7sp[01]. 
3. slightly improves handle name stringifying (now registry keys are parsed) 
4. adds 64-bit support for windows7

this has been verified on ubuntu 20.04 for 32/64-bit Windows 7, 32-bit Windows XP, and 32-bit Windows 2000. I tested `osi_test`, `file_taint`, and `syscalls2` with the new library. happy to test anywhere else if i missed something!


jamcleod:
@Ben-Dumas it seems like your usage of `$_` is causing issues while building the docker container:

```
#12 [builder 4/7] RUN cd /tmp &&   git clone https://github.com/panda-re/libosi &&   mkdir libosi/build && cd $_ &&   cmake -GNinja .. && ninja && ninja package && dpkg -i libosi*.deb &&   cd /tmp && rm -rf libosi/
#12 sha256:8422d06c53f2ec6cf74152e23dfa6176552bf7f325dfe8d3bd95f57198ee1c2b
#12 0.338 Cloning into 'libosi'...
#12 0.769 CMake Error: The source directory "/" does not appear to contain CMakeLists.txt.
```
(See the `build_container` test)

panda/plugins/wintrospection/pandamemory.cpp
```diff
@@ -0,0 +1,90 @@
...
+                                uint8_t *buffer, uint64_t size) {
+  auto papm = (PandaPhysicalMemory *)pmem->opaque;
+  auto max_addr = papm->get_max_address();
+
+  for (size_t ix = 0; ix < size; ++ix) {
+    auto current_addr = addr + ix;
+    if ((uintptr_t)current_addr > max_addr) {
+      return false;
+    }
+    buffer[ix] = papm->get_byte(current_addr);
```
lacraig2: This seems like a really inefficient way to read memory. You end up doing a considerable lookup with `panda_physical_memory_rw` every time you call it to read a single byte.

panda/plugins/wintrospection/pandamemory.cpp
```diff
@@ -0,0 +1,90 @@
...
+    m_max_address = region->size + region->addr;
+    rcu_read_unlock();
+  }
+
+  pm_addr_t get_max_address() { return m_max_address; }
+
+  uint8_t get_byte(pm_addr_t addr) {
+    uint8_t retval = 0;
+    if (addr < m_max_address) {
+      panda_physical_memory_rw(addr, &retval, 1, 0);
```
lacraig2: Consider using `panda_physical_memory_read` instead. It's just an inline that places the 0 for `is_write`.

lacraig2: This. To my suprise. Does not exist. I'll write up a quick PR to make it to parity with `panda_virtual_memory_*`.

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -0,0 +1,622 @@
+/* PANDABEGINCOMMENT
+ *
+ * Authors:
+ *  Tim Leek               tleek@ll.mit.edu
+ *  Ryan Whelan            rwhelan@ll.mit.edu
+ *  Joshua Hodosh          josh.hodosh@ll.mit.edu
+ *  Michael Zhivich        mzhivich@ll.mit.edu
+ *  Brendan Dolan-Gavitt   brendandg@gatech.edu
```
lacraig2: I feel like @Ben-Dumas your name should really be up here as well.

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -0,0 +1,622 @@
...
+}
+
+bool init_plugin(void *self) {
+#if defined(TARGET_I386) // only supports i386 and x86_64
+  panda_require("osi");
+  assert(init_osi_api());
+
+  panda_cb pcb;
+
+  pcb.before_block_exec = before_block_exec;
```
lacraig2: Consider using `start_block_exec` over `before_block_exec`. This will allow the plugin to run without chaining disabled.

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -0,0 +1,622 @@
...
+  g_update_task = false;
+  return true;
+#endif
+  return false;
+}
+
+void uninit_plugin(void *self) {
+  printf("Unloading wintrospection plugin\n");
+
+  // if we don't clear tb's when this exits we have TBs which can call
```
lacraig2, line 731: Do you have hooks somewhere that I'm not seeing? Are you doing TCG injection somewhere? This is in reference to the clearing of the TB cache.

lacraig2 (changes requested):
Just some small things to consider.

panda/plugins/wintrospection/pandamemory.cpp
lacraig2: This will be available when PR #1148 closes.

panda/plugins/wintrospection/wintrospection.cpp
Ben-Dumas: ah, this was left over from when i started with `wintrospection.c` plugin

https://github.com/panda-re/panda/blob/dev/panda/plugins/wintrospection/wintrospection.c#L870-L875

Will remove

panda/plugins/wintrospection/wintrospection.cpp
Ben-Dumas: Interesting, I didn't know that. 
Should i call `panda_enable_tb_chaining`, or are you just referring to allowing a live system (vs replay) to run?

panda/plugins/wintrospection/pandamemory.cpp
Ben-Dumas: This was taken from the way we've handled memory snapshots in the past [here](https://github.com/panda-re/libosi/blob/master/src/iohal/memory/pmem_snapshot.cc#L55-L69). But that's fair! I'll check it out and see if If changing it improves performance (or if i can figure out why we did it that way in the first place).

panda/plugins/wintrospection/wintrospection.cpp
lacraig2: Gotcha. Turns out that _was_ necessary when the old plugin was doing TCG insertion for hooking the task change address.

lacraig2: See: https://github.com/panda-re/panda/blob/dev/panda/plugins/wintrospection/wintrospection.c#L740

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -0,0 +1,622 @@
...
+
+  notify_task_change(cpu);
+
+  g_process_manager.reset(new WindowsProcessManager());
+  update_process_manager();
+
+  g_update_task = false;
+}
+
+bool asid_changed(CPUState *cpu, target_ulong old_pgd, target_ulong new_pgd) {
```
lacraig2, line 528: Is it possible to manage tasks changing like the current wintrospection plugin does with hooks on the task change function? https://github.com/panda-re/panda/blob/d4c6508d0b0e39bec20b32429cbdfa9836ded2fe/panda/plugins/wintrospection/wintrospection.c#L740

If so you could likely replace this logic with a hook.

panda/plugins/wintrospection/pandamemory.cpp
lacraig2: I can see it being reasonable in that example. It might be fine, but the overhead of repeatedly doing this logic (https://github.com/panda-re/panda/blob/fcd26ede117337fe68610391d4451a508f41201f/panda/include/panda/common.h#L76-L95) multiplied by the number of bytes seems like it might be an issue.

panda/plugins/wintrospection/wintrospection.cpp
lacraig2: If you don't enable `tb_chaining` the callback `before_block_exec` is not guaranteed to run on every block. That is to say that if PANDA runs a chain of TBs you wont see the intermediary blocks.

It's not always required to disable `tb_chaining` for every `bbe`, but if you want that guarantee you must disable it.

That was the case util we added `start_block_exec`. `start_block_exec` is TCG instrumentation in every block that calls our callback. That way even if PANDA runs a chain of TBs we maintain the guarantee to run on every block.

panda/plugins/wintrospection/wintrospection.cpp
Ben-Dumas: changed names to myself and other two that have worked with wintrospection/libosi in the past! I've always just used the panda standard comment 🤷

panda/plugins/wintrospection/pandamemory.cpp
lacraig2: PR #1148 merged so feel free to add.

panda/plugins/wintrospection/wintrospection.cpp
Ben-Dumas: Is that more accurate than waiting for the ASID to change? I could change it back; I just liked that this method isn't OS-version dependent.

lacraig2: We found that for `osi_linux` hooking on the task change function was not only more performant, but far more accurate.

I can really only speak to the linux side of it, but I do believe there will be windows equivalents. There's quite a few cases where the ASID doesn't accurately reflect the process. One example is `fork`. When the OS creates a new process it doesn't necessarily have to create new page tables for the process through Copy On Write (COW). Another example is `execve`, which may well reuse the same page table address when it "replaces" the process. This might result in the ASID not changing, but the underlying process changing.

lacraig2: If you're looking for something more OS-agnostic you could consider adding logic to the libosi library to handle this and calling it?

Ben-Dumas: what about this newest commit (https://github.com/panda-re/panda/pull/1138/commits/a88a22bebc4570351eaa12e4732560dca74746f1)? Here, I use the address of `nt!SwapContext` *if it is known* by libosi. Otherwise, it falls back to the startblock/asid_change method. This way, a user can always improve their accuracy by determining the address for their OS version, but they have something to fall back on. The offsets are currently set for the profiles that were previously supported.

lacraig2: Looks good. I changed it around to use `hooks`, but other than that it looks great.

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -0,0 +1,679 @@
...
+PANDAENDCOMMENT */
+// This needs to be defined before anything is included in order to get
+// the PRIx64 macro
+#define __STDC_FORMAT_MACROS
+
+#include <dirent.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#include <offset/i_t.h>
```
lacraig2, line 20: Some method of keeping versions of wintrospection and libosi in sync would be good.

lacraig2 (approved):

AndrewFasano:
🥳 Nice work @Ben-Dumas, thanks for upstreaming your code!!

(And thanks as always to @lacraig2 for getting it tested and merged)

Coffeeri:
This looks great, thank you! I am currently trying to use Panda and get the PEB content, Windows API calls of a process.
Is there any documentation on the new API and how to use the new wintrospection plugin and library?

Ben-Dumas:
@Coffeeri I'm glad you might find it useful! feel free to submit issues/PRs to [libosi](https://github.com/panda-re/libosi) for issues you're having! But in general -

The `wintrospection` plugin should give you an example of initializing windows introspection. To walk through it, once the kernel introspection [is initialized](https://github.com/panda-re/panda/blob/dev/panda/plugins/wintrospection/wintrospection.cpp#L621-L637), you can interact with memory using the [`osi::i_t` class](https://github.com/panda-re/libosi/blob/master/include/offset/i_t.h) or the [windows introspection API](https://github.com/panda-re/libosi/blob/master/include/osi/windows/wintrospection.h). You would probably want to [initialize introspection](https://github.com/panda-re/panda/blob/dev/panda/plugins/wintrospection/wintrospection.cpp#L516-L518) for the currently running process and then access the PEB [like this](https://github.com/panda-re/panda/blob/dev/panda/plugins/wintrospection/wintrospection.cpp#L231-L242). After that, we don't currently have a good way of knowing what types are available for you to use, outside of just looking at the code ☹️ For example, looking at windows-64-7sp1, I know i have [these types](https://github.com/panda-re/libosi/blob/master/src/offset/profiles/win7_sp1_x64.cpp#L4315-L4406) for the PEB.  

You can currently get a list of all processes, get a list of all modules, resolve handles to get a pointer to the object, and parse PE files in memory. The libosi library is still in a very early state, and documentation *should* definitely get added into the pipeline; that's a good point.

To your question of how to use `wintrospection`, it's not too different from other plugins (just make sure to give the `os` flag):

```bash
$PANDA_PATH/x86_64-softmmu/panda-system-x86_64 -replay mytrace \
  -m 2G -os windows-64-7sp1 -panda osi -panda wintrospection -panda osi_test
```

If you have any more questions, follow up with me via email (Benjamin.Dumas@ll.mit.edu), the panda slack (preferable, get access here https://panda.re/invite.php), or a new issue now that this PR is closed.

PR #1142, opened 2021-12-08
Add dedicated plugin for hardware-agnostic process ID

AndrewFasano:
panda_current_asid provides a unique, per-process ID for x86, arm, and PPC guests - but not MIPS!

This PR adds a simple plugin, `hw_proc_id` which provides one ppp-function: `int get_id(CPUState*)`  which returns a unique process ID for any of our supported architectures.

For non mips architectures, this is just the asid.
For MIPS, we cache the address of the current_task saved in `r28` whenever we're in kernel mode.

### DIFF
--- a/panda/plugins/hw_proc_id/README.md
+++ b/panda/plugins/hw_proc_id/README.md
@@ -1,0 +1,25 @@
+Plugin: `HW_PROC_ID`
+===========
+
+Summary
+-------
+This plugin aims to provide an architecture-agnostic way to uniquely identify processes through a single API.
+
+For non-MIPS architectures, this plugin is equivalent to using the `panda_current_asid()` function. However, on MIPS guests, the ASID changes frequently for the same process so a different implementation is needed. For these guests, we return the address of the `current` `task_struct` object which will be different for different processes. But, (like with ASIDs) after a process is terminated, another process could end up with it's `task_struct` object in the same location.
+
+
+Arguments
+---------
+None
+
+Dependencies
+------------
+None
+
+APIs and Callbacks
+------------------
+
+`int procid(CPUState*)`: Returns an integer that represents the current process running on the CPU.
+
+Example
+-------
--- a/panda/plugins/hw_proc_id/hw_proc_id.cpp
+++ b/panda/plugins/hw_proc_id/hw_proc_id.cpp
@@ -1,0 +1,63 @@
+/* PANDABEGINCOMMENT
+ * 
+ * Authors:
+ *  Andrew Fasano          fasano@mit.edu
+ * 
+ * This work is licensed under the terms of the GNU GPL, version 2. 
+ * See the COPYING file in the top-level directory. 
+ * 
+PANDAENDCOMMENT */
+
+// Simple plugin to provide an architecture-agnostic, generic capability
+// for uniquely identifying a process withotu OSI.
+//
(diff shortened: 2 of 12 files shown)


panda/plugins/hw_proc_id/hw_proc_id.cpp
```diff
@@ -0,0 +1,63 @@
...
+  return last_r28;
+#else
+  return panda_current_asid(cpu);
+#endif
+}
+
+bool init_plugin(void *self) {
+#if defined(TARGET_MIPS)
+    panda_cb pcb = { .before_block_exec = r28_cache };
+    panda_register_callback(self, PANDA_CB_BEFORE_BLOCK_EXEC, pcb);
```
lacraig2, line 58: Does it need to be `before_block_exec`? Can it be `start_block_exec`?

panda/plugins/hw_proc_id/README.md
```diff
@@ -0,0 +1,25 @@
...
+This plugin aims to provide an architecture-agnostic way to uniquely identify processes through a single API.
+
+For non-MIPS architectures, this plugin is equivalent to using the `panda_current_asid()` function. However, on MIPS guests, the ASID changes frequently for the same process so a different implementation is needed. For these guests, we return the address of the `current` `task_struct` object which will be different for different processes. But, (like with ASIDs) after a process is terminated, another process could end up with it's `task_struct` object in the same location.
+
+
+Arguments
+---------
+None
+
+Dependencies
```
lacraig2, line 15: There isn't a dependency, but you've created a `syscalls2` dependency that I don't see documented.

panda/plugins/hw_proc_id/hw_proc_id.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#ifdef TARGET_MIPS
+target_ulong last_r28 = 0;
+
+void r28_cache(CPUState *cpu, TranslationBlock *tb) {
+  // Whenever the kernel changes register 28 (current task struct)
+  // save it - Unless it's <= 0x80000000- then it's not the task struct(?)
+
+  if (panda_in_kernel(cpu) && unlikely(((CPUMIPSState*)cpu->env_ptr)->active_tc.gpr[28] != last_r28)) {
```
lacraig2, line 36: Should this be `panda_in_kernel` instead of `panda_in_kernel_code_linux`? It makes the assumption that the highest big of PC is set for kernel code.  Though given that we assume that the pointer is > 0x80000000.

You can get slightly more specific and consider checking the region for MIPS. http://www.it.uu.se/education/course/homepage/os/vt18/module-0/mips-and-mars/mips-memory-layout/

lacraig2: Actually, consider adding a macro for `IS_KERNEL_ADDRESS` that would be something like:

```
#define IS_KERNEL_ADDRESS(ADDR) (ADDR & 0x80000000)
```

lacraig2: You can use that for your potential here as well.

lacraig2 (changes requested):
There's some documentation in `osi_linux` and some opimizations that should probably be made, but other than that good.

lacraig2:
Superseded by #1156

PR #1146, opened 2021-12-09
double return for syscalls2

lacraig2:
This PR adda another category of special syscalls called "double return". This category represents those syscalls that return twice in potentially different ASIDs to the same program counter. This is mostly just fork, but it could be other things too.

I also added "sys_exit_group" to the `context_target.json` for noreturns, but really that whole file needs a rethink.


lacraig2:
It's sat for a week so I'm gonna merge it myself.

PR #1149, opened 2021-12-17
Fix pypanda exceptions

### LINKED ISSUE #1009: PyPANDA Panda.hook2/_hook_mem exception handling

lacraig2:
This PR does the following:
- fixes the error wrappers for several functions
- consolidates errors into single type
- checks for errors prior to entry to threaded functions

It does so while attempting to maintain the `snake_hook` exception pass-through.

Supersedes old PR #1100.

FIXES: #1009

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -534,21 +534,9 @@ class Panda():
         #self.libpanda.panda_cleanup_record()
         if self._in_replay:
             self.reset()
-        if hasattr(self, "end_run_raise_signal"):
-            saved_exception = self.end_run_raise_signal
-            del self.end_run_raise_signal
-            raise saved_exception
-        if hasattr(self, "callback_exit_exception"):
-            saved_exception = self.callback_exit_exception
-            del self.callback_exit_exception
-            raise saved_exception
-        if hasattr(self, "blocking_queue_error"):
-            saved_exception = self.blocking_queue_error
-            del self.blocking_queue_error
-            raise saved_exception
-        if hasattr(self, "hook_exit_exception"):
-            saved_exception = self.hook_exit_exception
-            del self.hook_exit_exception
+        if hasattr(self, "exit_exception"):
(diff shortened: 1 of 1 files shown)


jamcleod (approved):

AndrewFasano:
I'm not positive this PR is to blame, but after pulling this commit I'm no longer getting exceptions for syntax errors and other bugs in a pyplugin I'm working with

AndrewFasano:
Just confirmed that it is caused by this commit. For example the following pypanda script exits without printing any errors (or running the print)
```
class Foo(PyPlugin):
  def __init__(self):
    asdf this is invalid
    print("And this doesn't run")

panda.load_plugin(Foo)
panda.run()
```

PR #1151, opened 2021-12-23
Add no_timeout and timeout options for calls

### LINKED ISSUE #1150: Question or feature request: PyPanda API for timeout for panda.record_cmd()
Since some programs being tested might not exit on their own, or might be stuck waiting for something, it would be useful to have an optional timeout parameter for `panda.record_cmd()`. Instead, it seems the program runs forever (it's a simple test app I wrote, doesn't do much) and the recording ends up filling up the partition.

I'm not using the latest version, but checking current sources & examples shows this feature isn't implemented.

Git hash: 713ebc3917
Git date: 16 Nov 2021
Python: 3.8.10
Host OS: Ubuntu 20.04.3
Guest OS: Windows 10 64-bit (using `os="windows", expect_prompt=rb"C*>"` in the `Panda()` constructor in case I'm doing something wrong)

lacraig2:
This PR adds `no_timeout` to `copy_to_guest` and `record_cmd` as well as `timeout` to `record_cmd`.

Related to #1150

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -2356,7 +2356,7 @@ class Panda():
         self.run_monitor_cmd("delvm {}".format(snapshot_name))
 
     @blocking
-    def copy_to_guest(self, copy_directory, iso_name=None, absolute_paths=False, setup_script="setup.sh", timeout=None, cdrom=None):
+    def copy_to_guest(self, copy_directory, iso_name=None, absolute_paths=False, setup_script="setup.sh", timeout=None, no_timeout=False, cdrom=None):
         '''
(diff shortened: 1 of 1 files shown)


AndrewFasano:
Adding `timeout` everywhere looks good, but do we really want to have an int and a bool option for all of these? I'd think we could instead just get rid of the no_timeout flag everywhere. Either way is fine with me though

lacraig2:
@AndrewFasano That's a good point.

It looks like `no_timeout` only shows up in one place with `run_serial_cmd`. Removing it would break things currently using it.

We can certainly make these all take a timeout value.

PR #1162, opened 2022-01-11
Fix windows 7 os naming convention in examples

Coffeeri:
In PR #1138 the windows-32-7 profile was split into windows-32-7sp[01].
This PR simply updates the examples accordingly.


lacraig2:
Nice catch. Thank you.

PR #1169, opened 2022-02-01
Update function prototypes in extern "C" block to match function definitions.

MarkMankins:
I have some plugins that depend on wintrospection that aren't working with the new version of the wintrospection plugin.

I'm getting a runtime symbol resolution error when my plugin is loaded (get_handle_name).

I tracked it down to a mismatch between the function prototypes in the extern C block in wintrospection.cpp.  These didn't match the actual function definitions.  Updating these to match resulted in the functions not getting name mangled in the target dynamic library and my symbol resolution problems disappeared.


lacraig2:
Good catch!

Issue #570, opened 2020-03-20
Update to modern QEMU!


AndrewFasano:
We've been talking about this for a bit but haven't started work on it. Creating this issue to track progress.

We're currently forked off of Qemu at version 2.9.1. We should update to 4+. At the time of writing, the latest version if 4.1. 

We'll likely need to disable MTTCG to avoid significantly changing the record/replay model.

The main tasks I see for now are:
- [ ]  Actually do the **git merge** and handle merge conflicts to create a (likely broken) branch with all the commits
- [ ] Get **callbacks** to run
- [ ] Capture correct **recordings**
- [ ] **Replay** recordings correctly
- [ ] **Library mode**: build qemu as a library 
- [ ] **Pypanda** testing
- [ ] Extensive **testing**

janbbeck:
Just to give it my 2 cents:
I recently upgraded qira from qemu 2.x to qemu 4.0  instead of 4.1. At least for that effort, the changes after 4.0 were getting so severe that porting to 4.0 before going to 4.1 made sense.
This may, of course, not be true for PANDA, but it may be easier to port PANDA to some intermediate version of qemu than going to 4.1 all at once.

hanetzer:
judging by the amount of monkeypatching I needed to do to get pandas as it currently
exists to build on my system, I almost think it would be easier to collect what is different
between pandas and the qemu version its based on and just rewrite it against qemu4.

AndrewFasano:
@janbbeck Thanks for the tip. 4.0 sounds like it might be a more realistic target.

@hanetzer It's unlikely that any of us have time to do a full rewrite given that there are 2k commits to PANDA since we forked. If we had time, that would certainly be the cleanest way. Also- you shouldn't need to do any monkeypatching- if you've had build issues can you open an issue? Our CI has been able to build PANDA on clean machines without any problems.

hanetzer:
@AndrewFasano I'm running on gentoo ~amd64 (read: bleeding on the edge),
so there's a lot of new gcc diagnostics and such that make the build fail in creative ways.

While I have a human looking at this, I'll open an issue about what I'm trying to
do and what is/isn't/idk working.

AndrewFasano:
@hanetzer Ah, okay, that makes more sense then. We just fixed some gcc7-related errors in the last few weeks but that's still probably not new enough for you.

hanetzer:
@AndrewFasano yeah. I'm currently working under a deadline and this
[...]

PR #1170, opened 2022-02-03
Backport: Fix byte-to-interrupt number conversions in armv7m_nvic

luglo:
While playing around with freeRTOS and PANDA I noticed interrupt offset calculation is off. This has been fixed in upstream QEMU for a long time.

This is a backport of QEMU upstream patch qemu/qemu@12fbf1a1639ed916fda948718dac0d30b82b954e


lacraig2:
@luglo nice catch.

Missing upstream fixes like this is a reality of forking QEMU unfortunately. There has been discussion about updating, but it is quite an undertaking. #570

AndrewFasano:
Backporting fixes from upstream is much appreciated though! Thanks for this PR

PR #1114, opened 2021-10-20
Add QEMU logging to PyPANDA

AndrewFasano:
Allows for dynamically changing the log level from pypanda. Useful for enabling things like in_asm logging for specific parts of an execution.

Could be made better with some helper functions which include the mappings between log level names and their integer values.

PR #1175, opened 2022-02-18
move avatar output (i.e., configurable machine) behind log flag

mariusmue:
In case you ever used avatar's configurable machine and where annoyed by all the logging output of created memory regions and similar, this PR provides the solution!
These prints are now guarded by qemu's logging/log_mask infrastructure. This required to adjust the type of the qemu log mask from int to long - but this change should be backwards compatible either way.


include/qemu/log.h
```diff
@@ -4,7 +4,7 @@
 
 /* Private global variables, don't use */
 extern FILE *qemu_logfile;
-extern int qemu_loglevel;
+extern long qemu_loglevel;
```
lacraig2: Do we need to change this to a long? There's definitely still room in the `int` for an AVATAR log.

P.S. PR #1114 will add a `LOG_PANDA` so it'd be good to not use 17. 18-27 are available though!

lacraig2 (changes requested):
Everything else looks good. Not changing to a `long` would simplify this significantly.

lacraig2:
I merged that other old PR and fixed the conflict with this PR.

include/qemu/log.h
mariusmue: Good catch for 18-27, I didn't realize and just appended after the last #define! Updated that branch accordingly

mariusmue:
Uhm. I think the current CI failures are not related to my log_mask changes - how would we move forward here?

lacraig2:
You're right. I'll run it again.

lacraig2:
I went ahead and added this log mechanism to `qemu_logging.py` in pypanda. That'll let you selectively turn on/off various log types dynamically.

Once tests pass this should be good to go. Thanks for the PR!

lacraig2 (approved):

PR #1182, opened 2022-02-28
Expose windows process manager so it can be leveraged by other plugins.

MarkMankins:
This works but it seems like there should be a better way, but it wasn't immediately obvious to me what that might be.

I have some plugins that relied on some functions that were in the pre-libosi version of wintrospection.

I was able to repair those plugins by allowing them to get a handle to the WindowsProcessManager object.

This change allows my plugins to call get_windows_process_manager and cast the return value to a WindowsProcessManager *. From there I as able to repair them to replace the logic that wintrospection used to provide.

Happy to change this to something else if there is a better way to achieve the same outcome.

### DIFF
--- a/panda/plugins/wintrospection/wintrospection.cpp
+++ b/panda/plugins/wintrospection/wintrospection.cpp
@@ -54,6 +54,7 @@ void uninit_plugin(void *);
 int64_t get_file_handle_pos(CPUState *cpu, uint64_t handle);
 char *get_cwd(CPUState *cpu);
 char *get_handle_name(CPUState *cpu, uint64_t handle);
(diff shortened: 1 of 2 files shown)


Ben-Dumas (approved):
Looks good to me; Might need to know what you're working on to brainstorm any better way to handle it, because i _thought_ I added all previous functionality to this plugin. But I don't see anything wrong with this!

Edit: If i did miss some functionality, I could also just add that back in to this plugin

MarkMankins:
Having access to all the cool new stuff in libosi is much better than what I was previously doing - so I don't think there's a need to go back to the old way I was doing things.

PR #1184, opened 2022-03-04
Fallback to using ASID change heuristic to detect task changes on win…

MarkMankins:
…dows 2000.

There are couple of problems with detecting task changes on windows 2000.

First, there are at least five different common versions of the windows 2000
kernel, each with its own unique SwapContext address.  libosi currently
includes the correct offset for windows 2000 service pack 4.

Secondly, for some reason, task changes in Windows 2000 seem to occur in other
places in the kernel outside of SwapContext.

With the older version of wintrosepection, this wasn't a problem as this
was only noticed if one was using task change notifications.

The updated wintrospection brought with it a change that task change
detection must work if one wants to use on_get_current_thread.

If task change detection isn't working, on_get_current_thread will return
the currently running thread id, and the process id of the process running
when the last task change was detected.  Sometimes a thread id would be
returned and 0 for the process id.

I'm seeing this behavior in windows 2000 with plugins that rely on
on_get_current_thread such as callstack_trace, asidstory, and proc_trace.

When using the fallback method the results more closely match the old behavior
of wintrospection.

### DIFF
--- a/panda/plugins/wintrospection/wintrospection.cpp
+++ b/panda/plugins/wintrospection/wintrospection.cpp
@@ -71,12 +71,19 @@ void on_get_modules(CPUState *cpu, GArray **out);
 void on_get_mappings(CPUState *cpu, OsiProc *p, GArray **out);
 
 void initialize_introspection(CPUState *cpu);
+void task_change(CPUState *cpu);
 
 std::unique_ptr<WindowsKernelManager> g_kernel_manager;
 std::unique_ptr<WindowsProcessManager> g_process_manager;
 
 bool g_initialized_kernel;
 uint64_t swapcontext_address;
+
+// last thread id seen from on_get_current_thread
+static uint64_t last_seen_tid = std::numeric_limits<uint64_t>::max();
+
+// true when using asid change heuristic to detect task changes
+static bool using_asid_change_heuristic = true;
 
 // globals for toggling callbacks
 void* self = NULL;
@@ -276,6 +283,19 @@ char *get_handle_name(CPUState *cpu, uint64_t handle) {
   return name;
 }
 
+static inline uint64_t get_current_thread_id(CPUState *cpu) {
+  auto kernel = g_kernel_manager->get_kernel_object();
+  uint64_t tid = kosi_get_current_tid(kernel);
+
+  // if the thread id changed, ensure the pid is also updated
+  if(using_asid_change_heuristic && (tid != last_seen_tid)) {
+      last_seen_tid = tid;
+      task_change(cpu);
+  }
+
+  return tid;
+}
+
 /* ******************************************************************
  PPP Callbacks
 ****************************************************************** */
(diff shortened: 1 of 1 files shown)


MarkMankins:
To be more clear - prior to the change calling on_get_current_thread when running a windows 2000 version other than windows 2000 sp4 just doesn't work.

Running windows 2000 sp4, sometimes the expected result would be returned, sometimes a non-zero thread id would be returned along with a process id of 0, and sometimes a non-zero thread id would be returned along with a non-zero process id, but the process id was not the process id associated with the thread id.

Ben-Dumas (approved):
This would work! If hooking SwapContext doesn't even work for all Windows 2k SP4, then this should be the default, agreed. If it does, and you're just having trouble with other service packs, then I could see 
1. This solution 
2. Adding addresses for other service packs to libosi, and splitting out the windows 2k profile in PANDA like windows XP. 
3. Adding a function to wintrospection that, when called would cause the process manager to update (feels the most hacky, but just listing out ideas in case it helps), such as 
```cpp
update_process_manager();
auto thread = get_current_thread(cpu);
```

MarkMankins:
Still working this one - hope to have something better in a day or two.

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -276,6 +283,19 @@ char *get_handle_name(CPUState *cpu, uint64_t handle) {
...
 }
 
+static inline uint64_t get_current_thread_id(CPUState *cpu) {
+  auto kernel = g_kernel_manager->get_kernel_object();
+  uint64_t tid = kosi_get_current_tid(kernel);
+
+  // if the thread id changed, ensure the pid is also updated
+  if(using_asid_change_heuristic && (tid != last_seen_tid)) {
+      last_seen_tid = tid;
+      task_change(cpu);
```
Ben-Dumas: this would call `notify_task_change` every time the thread updates (or at least every time the subscribing plugin asks for the current state). I thought that was reserved for a new _process_ taking control?

MarkMankins: Good point.  I could add a check in task_change to validate that indeed a new process was detected, and only call notify_task_change if that was true.  Would that work?

Ben-Dumas: That sounds good to me, yep!

panda/plugins/wintrospection/wintrospection.cpp
```diff
@@ -516,13 +549,18 @@ void on_get_mappings(CPUState *cpu, OsiProc *p, GArray **out) {
...
 
 void task_change(CPUState *cpu) {
-  g_process_manager.reset(new WindowsProcessManager());
-
   auto kernel = g_kernel_manager->get_kernel_object();
-  g_process_manager->initialize(kernel,
-                                kosi_get_current_process_address(kernel));
+  uint64_t paddr = kosi_get_current_process_address(kernel);
+
+  if(paddr != last_seen_paddr) {
```
Ben-Dumas: maybe i'm missing something -- why check if the process has changed here? Why not just change the `get_current_thread_id` to only update the `WindowsProcessManager` when the process has changed? Now you're checking the thread and then the process, when the `WindowsProcessManager` only tracks process anyway

MarkMankins:
See if this is more like what you were thinking...

Checking thread id first avoids the overhead of having to call kosi_get_current_process_address on every call to fill_thread.

Ben-Dumas (approved):
Can't see the former commit anymore, but I thought you were checking the thread ID at an unnecessary time. May have been mistaken, but either way, this latest commit looks good to me!

MarkMankins:
This isn't waiting on any more inputs from me is it?

PR #1185, opened 2022-03-08
Fix taint inconsistency

### LINKED ISSUE #711: taint2: inconsistent taint results
I used `taint2` and `tainted_branch` to conduct some taint experiments. During this process, I observed that `tainted_branch` outputs different results for the same replay, especially when the replay is a long one. Sometimes it have abnormal results: some branches that should be tainted were not included in the pandalog.

If my understanding is correct, here's one of the possible reason: according to the code, the taint operations is performed during the LLVM code generation, before the basic block being executed. Therefore, the taint operations were done even if the block is not executed.

To verify this, l patched `taint2`. At the second time the case I mentioned above happens, the emulation would stop:

```
ae9fa80768dc /panda/panda/plugins/taint2 # diff -Nuar taint2.cpp.bak taint2.cpp
--- taint2.cpp.bak	2020-07-20 12:40:01.643000000 +0000
+++ taint2.cpp	2020-07-24 10:21:34.175328705 +0000
@@ -57,6 +57,9 @@
 #include "taint_api.h"
 #include "taint2_hypercalls.h"

+//for outputting debug message
+#include <inttypes.h>
+
 #define CPU_OFF(member) (uint64_t)(&((CPUArchState *)0)->member)

 extern "C" {
@@ -68,6 +71,11 @@
 int after_block_translate(CPUState *cpu, TranslationBlock *tb);
 bool before_block_exec_invalidate_opt(CPUState *cpu, TranslationBlock *tb);

+void after_block_exec(CPUState *cpu, TranslationBlock *tb, uint8_t exitcode);
+bool taint_changed_in_last_bb = false;
+// when enabling taint(the first time we taint something), the exitcode would likely be 3, ignore this one.
+bool not_first_time = false;
+
 // for i386 condition code adjustments
 #if defined(TARGET_I386)
 void i386_after_cpu_exec_enter(CPUState *cpu);
@@ -114,6 +122,10 @@
 #endif

 bool asid_changed_callback(CPUState *env, target_ulong oldval, target_ulong newval);
+
+// get guest instr count
+#include "panda/rr/rr_log.h"
+
 }

 ShadowState *shadow = nullptr; // Global shadow memory
@@ -290,6 +302,9 @@

     panda_enable_precise_pc(); //before_block_exec requires precise_pc for panda_current_asid

+    pcb.after_block_exec = after_block_exec;
+    panda_register_callback(taint2_plugin, PANDA_CB_AFTER_BLOCK_EXEC, pcb);
+
     if (!execute_llvm){
         panda_enable_llvm();
     }
@@ -464,27 +479,43 @@
  */
 void taint_state_changed(Shad *shad, uint64_t shad_addr, uint64_t size)
 {
+    fprintf(stderr,"[debug] instr: %" PRIu64 " ", rr_get_guest_instr_count());
     Addr addr;
     if (shad == &shadow->llv) {
         addr = make_laddr(shad_addr / MAXREGSIZE, shad_addr % MAXREGSIZE);
+        fprintf(stderr,"taint llvm register, addr=%" PRIx64 ", size=%" PRIu64 "\n",shad_addr, size);
+        taint_changed_in_last_bb = true;
+
     } else if (shad == &shadow->ram) {
         addr = make_maddr(shad_addr);
+        fprintf(stderr,"taint ram, addr=%" PRIx64 ", size=%" PRIu64 "\n",shad_addr, size);
+        taint_changed_in_last_bb = true;
     } else if (shad == &shadow->grv) {
         addr = make_greg(shad_addr / sizeof(target_ulong), shad_addr % sizeof(target_ulong));
+        fprintf(stderr,"taint guest general purpose registers, addr=%" PRIx64 ", size=%" PRIu64 "\n",shad_addr, size);
+        taint_changed_in_last_bb = true;
     } else if (shad == &shadow->gsv) {
         addr.typ = GSPEC;
         addr.val.gs = shad_addr;
         addr.off = 0;
         addr.flag = (AddrFlag)0;
+        fprintf(stderr,"taint guest special values, addr=%" PRIx64 ", size=%" PRIu64 "\n",shad_addr,size);
+        taint_changed_in_last_bb = true;
     } else if (shad == &shadow->ret) {
         addr.typ = RET;
         addr.val.ret = 0;
         addr.off = shad_addr;
         addr.flag = (AddrFlag)0;
+        fprintf(stderr,"taint llvm return value, addr=%" PRIx64 ", size=%" PRIu64 "\n",shad_addr,size);
+        taint_changed_in_last_bb = true;
     } else if (shad == &shadow->hd) {
         addr = make_haddr(shad_addr);
+        fprintf(stderr,"taint harddrive, addr=%" PRIx64", size=%" PRIu64 "\n",shad_addr,size);
+        taint_changed_in_last_bb = true;
     } else if (shad == &shadow->io) {
         addr = make_iaddr(shad_addr);
+        fprintf(stderr,"taint io, addr=%" PRIx64", size=%" PRIu64 "\n",shad_addr,size);
+        taint_changed_in_last_bb = true;
         /*    } else if (shad == &shadow->ports) {
                 addr = make_paddr(shad_addr); */
     } else return;
@@ -493,12 +524,24 @@
 }

 bool before_block_exec_invalidate_opt(CPUState *cpu, TranslationBlock *tb) {
+    taint_changed_in_last_bb = false;
     if (taintEnabled) {
         return tb->llvm_tc_ptr ? false : true /* invalidate! */;
     }
     return false;
 }

+void after_block_exec(CPUState *cpu, TranslationBlock *tb, uint8_t exitcode) {
+    if (taint_changed_in_last_bb && (exitcode > 1)){
+        fprintf(stderr,"[error] taint state changed although last bb was not executed.\n");
+        if(not_first_time == false){
+            fprintf(stderr,"[debug] but this is the first time, let it pass...\n");
+            not_first_time = true;
+        }
+        else assert(0);
+    }
+    taint_changed_in_last_bb = false;
+}

 /**
  * @brief Basic initialization for `taint2` plugin.
```

I tested this with a replay about running `cat 1234` and tainted all `0xffffffff` function returns, on an `versatilepb` machine. It should be noted that the problem might not happen every time. What's more, I've updated capstone 4.

lacraig2:
This PR resolves a taint inconsistency present in the wrapper for pypanda. This should resolve #711.

Essentially, two problems existed:
- using `queue_main_loop` can significantly delay actions past the point at which a taint should have been applied
- The system allowed taint to be enabled in the same context as it was used. This can not be done. Taint must be enabled at least 1 block prior to being used.

FIXES: #711

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -220,8 +220,6 @@ class Panda():
 
         self._initialized_panda = False
         self.disabled_tb_chaining = False
-        self.taint_enabled = False
-        self.taint_sym_enabled = False
         self.named_hooks = {}
         self.hook_list = []
         self.hook_list2 = {}
@@ -655,6 +653,10 @@ class Panda():
         Load a C plugin with no arguments. Deprecated. Use load_plugin
         '''
         self.load_plugin(name, args={})
+    
+    def _plugin_loaded(self, name):
+        name_c = self.ffi.new("char[]", bytes(name, "utf-8"))
+        return self.libpanda.panda_get_plugin_by_name(name_c) != self.ffi.NULL
 
     def load_plugin(self, name, args={}):
         '''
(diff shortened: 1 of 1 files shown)


panda/python/core/pandare/panda.py
```diff
@@ -1868,71 +1870,65 @@ def unregister_pyperipheral(self, pyperiph):
...
+
+    def taint_label_reg(self, reg_num, label):
+        '''
+        Labels taint register reg_num with label.
+        '''
+        self._assert_taint_enabled()
         for i in range(self.register_size):
-            self.queue_main_loop_wait_fn(self.plugins['taint2'].taint2_label_reg, [reg_num, i, label])
-        self.queue_main_loop_wait_fn(self.libpanda.panda_cont, [])
+            self.plugins["taint2"].taint2_label_reg(reg_num, i, label)
```
AndrewFasano, line 1900: Now that the taint plugin will definitely be loaded prior to this is it okay to apply labels immediately here and on 1907 instead of waiting until the next main loop?

panda/python/core/pandare/panda.py
```diff
@@ -1868,71 +1870,65 @@ def unregister_pyperipheral(self, pyperiph):
...
 
-    # returns true if any bytes in this register have any taint labels
     def taint_check_reg(self, reg_num):
-        if not self.taint_enabled: return False
-#        if debug:
-#            progress("taint_check_reg %d" % (reg_num))
+        '''
+        Checks if register reg_num is tainted. Returns boolean.
+        '''
+        self._assert_taint_enabled()
```
AndrewFasano, line 1913: Thoughts on just returning False if taint isn't enabled? Both here and in taint_check_ram? I think that was the old behavior - but I'm not opposed to changing it.

panda/python/core/pandare/panda.py
```diff
@@ -1975,43 +1976,31 @@ def taint_get_laddr(self, addr, offset):
...
-    # XXX label must increment by panda.register_size after the call
+        self._assert_taint_sym_enabled()
+        self.plugins['taint2'].taint2_sym_label_ram(addr,label)
+
     def taint_sym_label_reg(self, reg_num, label):
+        # label all bytes in this register.
+        # or at least four of them
+        # XXX label must increment by panda.register_size after the call
+        self._assert_taint_sym_enabled()
         self.taint_sym_enable(cont=False)
```
AndrewFasano: This needs to be updated - cont is no longer an option to taint_sym_enable

AndrewFasano (changes requested):
It's great to see this issue finally being addressed! Just have a few small nits and one actual bug.

panda/python/core/pandare/panda.py
lacraig2: Good catch. Fixed. Thanks.

panda/python/core/pandare/panda.py
lacraig2: I don't think that returning False is appropriate here. I think if you ask for taint information without it being enabled that's an error of which you should generally be aware. In C, this would result in either a segfault or an assertion thrown.

panda/python/core/pandare/panda.py
lacraig2: Yes. I believe so. As long as the assertions are met I believe we should be able to handle this behavior. The key difference here is that these are done deterministically.

PR #1189, opened 2022-03-16
callbacks: normalize plugin name path

lacraig2:
This PR normalizes the panda path to plugins such that the realpath is provided back. This resolves an issue whereby plugins could be loaded from different sources with different visible paths that resolve to the same path. Because our comparison for the plugin loaded is done by the abolute path this can lead to a situation where the same plugin is loaded multiple times.

### DIFF
--- a/panda/src/callbacks.c
+++ b/panda/src/callbacks.c
@@ -99,19 +99,18 @@ bool panda_add_arg(const char *plugin_name, const char *plugin_arg) {
 // Forward declaration
 static void panda_args_set_help_wanted(const char *);
 
-static char* attempt_normalize_path(char* path){
+char* attempt_normalize_path(static char* path){
     char* new_path = g_malloc(PATH_MAX); 
     if (realpath(path, new_path) == NULL) {
-        g_free(new_path);
-        return path;
-    }else{
-        return new_path;
+        strncpy(new_path, path, PATH_MAX);
     }
+    g_free(path);
+    return new_path;
 }
 
 bool panda_load_external_plugin(const char *filename, const char *plugin_name, void *plugin_uuid, void *init_fn_ptr) {
     // don't load the same plugin twice
-    const char* rfilename = attempt_normalize_path((char*)filename);
+    char* rfilename = attempt_normalize_path(strdup(filename));
     uint32_t i;
     for (i=0; i<nb_panda_plugins_loaded; i++) {
         if (0 == (strcmp(rfilename, panda_plugins_loaded[i]))) {
@@ -134,7 +133,7 @@ bool panda_load_external_plugin(const char *filename, const char *plugin_name, v
     if (plugin_name) {
         strncpy(panda_plugins[nb_panda_plugins].name, plugin_name, 256);
     } else {
-        char *pn = g_path_get_basename((char *) rfilename);
+        char *pn = g_path_get_basename(rfilename);
         *g_strrstr(pn, HOST_DSOSUF) = '\0';
         strncpy(panda_plugins[nb_panda_plugins].name, pn, 256);
         g_free(pn);
--- a/panda/src/callbacks.c
+++ b/panda/src/callbacks.c
@@ -99,7 +99,7 @@ bool panda_add_arg(const char *plugin_name, const char *plugin_arg) {
 // Forward declaration
 static void panda_args_set_help_wanted(const char *);
 
-char* attempt_normalize_path(static char* path){
+static char* attempt_normalize_path(static char* path){
     char* new_path = g_malloc(PATH_MAX); 
     if (realpath(path, new_path) == NULL) {
         strncpy(new_path, path, PATH_MAX);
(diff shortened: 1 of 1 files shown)


panda/src/callbacks.c
```diff
@@ -99,18 +99,29 @@ bool panda_add_arg(const char *plugin_name, const char *plugin_arg) {
 // Forward declaration
 static void panda_args_set_help_wanted(const char *);
 
+static char* attempt_normalize_path(char* path){
```
jamcleod, line 105: The lifetime of the path returned by `attempt_normalize_path` is ambiguous.

Either:
1. normalization succeeds and the caller is responsible for freeing the path, so the caller should free the output and (if already responsible for it) the input.
2. normalization fails and the caller should not separately free both the returned path and the input

So the options are either:
1. have the caller responsible for handling failures of `realpath`
2. memcpy path to new_path on error and return new_path unconditionally
3. use an out parameter to indicate whether the lifetime of the return value is owned or borrows from the input (if you're familiar--this is basically `Cow<'a, str>` in Rust)
4. leak memory if `realpath` succeeds (eg never free the return value of `attempt_normalize_path`)

Currently I believe you're choosing option 4. Easiest option would likely be #2, as its only difference is that on error you make a copy (you'd already be allocating/freeing so it's a pretty minimal cost), and then it'd just have the caller be responsible for freeing.

panda/src/callbacks.c
```diff
@@ -99,18 +99,29 @@ bool panda_add_arg(const char *plugin_name, const char *plugin_arg) {
...
+        g_free(new_path);
+        return path;
+    }else{
+        return new_path;
+    }
+}
+
 bool panda_load_external_plugin(const char *filename, const char *plugin_name, void *plugin_uuid, void *init_fn_ptr) {
     // don't load the same plugin twice
+    const char* rfilename = attempt_normalize_path((char*)filename);
```
jamcleod: `rfilename` should not be const as it is responsible for freeing its contents

panda/src/callbacks.c
```diff
@@ -123,7 +134,7 @@ bool panda_load_external_plugin(const char *filename, const char *plugin_name, v
     if (plugin_name) {
         strncpy(panda_plugins[nb_panda_plugins].name, plugin_name, 256);
     } else {
-        char *pn = g_path_get_basename((char *) filename);
+        char *pn = g_path_get_basename((char *) rfilename);
```
jamcleod: this cast is casting away a const (unnecessarily, even, `g_path_get_basename` takes a `const gchar*`, which is a type alias for `const char*`)

panda/src/callbacks.c
```diff
@@ -289,9 +301,9 @@ char* resolve_file_from_plugin_directory(const char* file_name_fmt, const char*
     g_free(plugin_path);
 
     // Finally, try relative to the installation path.
-    plugin_path =
+    plugin_path = attempt_normalize_path(
```
jamcleod, line 306: this duplicates the path twice without freeing either copy

jamcleod (changes requested):
Minor issues regarding path lifetimes and const-correctness, once those are fixed, looks good to me.

panda/src/callbacks.c
lacraig2: yes. removed cast.

panda/src/callbacks.c
lacraig2: Fair. #2 sounds like a reasonable option.

panda/src/callbacks.c
lacraig2: should be resolved with the changes to `attempt_normalize_path`

panda/src/callbacks.c
lacraig2: yes.

panda/src/callbacks.c
```diff
@@ -99,18 +99,28 @@ bool panda_add_arg(const char *plugin_name, const char *plugin_arg) {
 // Forward declaration
 static void panda_args_set_help_wanted(const char *);
 
+static char* attempt_normalize_path(const char* path){
+    char* new_path = g_malloc(PATH_MAX); 
+    if (realpath(path, new_path) == NULL) {
+        strncpy(new_path, path, PATH_MAX);
+    }
+    g_free((char*)path);
```
jamcleod, line 110: If `path` is assumed to be owned (and thus `attempt_normalize_path` is responsible for freeing it) then `path` shouldn't be a `const` pointer. I'd say just change `path` to a `char*` and add a comment above the function stating that `attempt_normalize_path` takes ownership of the input path. 

And while imo not worth the effort, if you really want to take it even further it'd reduce allocator churn in the error case to return `new_path` on success and `path` on error (which, unlike before, is fine assuming that `attempt_normalize_path` takes ownership of its input). I wouldn't bother though.

jamcleod (approved):
One more minor issue about const correctness, but looks good to me.

panda/src/callbacks.c
lacraig2: Seems reasonable. Yes that should be `char*` and not `const char*`.

AndrewFasano:
Closing and reopening this to get CI to run

Issue #1087, opened 2021-09-03
Integrate new recording format


rjwhelan:
We have a new RR format we'd like to integrate. This issue will track development.

- Format is known as RR2
- Consists of a compressed file that contains:
    - Snapshot
    - Nondet log
    - Capture command
    - Magic file
    - SHA1 hash
    - Metadata file

To accompany this, there should probably be a command line option to use the old format vs. the new format.  It may also be helpful to include utilities/documentation for converting between formats.

Additionally, we propose the next major release of PANDA only supports this version.

Related to #339.

AndrewFasano:
Supersedes #625.

+1 for saving this until the next major version release - I think that could come out as early as this winter according to our rules.

Is the capture command just human-readable text? It would be nice to have a machine-readable field for the QEMU/panda machine type (the `-M` flag) which mostly matters for non-x86 systems.

rjwhelan:
The capture command is a human-readable copy of the command.  The metadata file could be used to store anything useful in JSON format.  We currently use it to store the memory size, OS profile, the image the recording was taken on, among other things.

Ben-Dumas:
This is currently being worked on by zacogen and is in progress.

lacraig2:
Finished with merge by #1194!

PR #1194, opened 2022-04-14
RR2 file compressed format

zacogen:
Add code for using rr2 file compressed format in record and replay.
More details on format in https://github.com/panda-re/panda/issues/1087


panda/src/rr/rr_log.c
```diff
@@ -286,58 +289,58 @@ static inline void rr_write_item(RR_log_entry item)
             break;
         case RR_INPUT_2:
             RR_WRITE_ITEM(item.variant.input_2);
-            break;
+	    break;
```
lacraig2: formatting looks a bit broken.

configure
```diff
@@ -3254,6 +3254,31 @@ if ! compile_prog "$glib_cflags -Werror" "$glib_libs" ; then
     fi
 fi
 
+##########################################
```
lacraig2, line 3257: Dependencies added here need to be added to Docker config files.

scripts/rr_pack.py
```diff
@@ -0,0 +1,142 @@
+import os
```
lacraig2: This script should be moved to /panda/scripts. /scripts if for QEMU scripts.

panda/include/panda/rr/rr_log.h
```diff
@@ -141,7 +142,11 @@ typedef struct RR_log_t {
     RR_prog_point last_prog_point; // to report progress
 
     char* name; // file name
-    FILE* fp;   // file pointer for log
+    union {
```
lacraig2, line 146: The union here makes me think that this supports both the old system and the new. Is that the case?

lacraig2 (commented):
Seems like a good start. I'd like some more clarification on the plan for old recordings.

panda/include/panda/rr/rr_log.h
zacogen: Originally, I had the code only supporting the new rr2 format, however this commit (https://github.com/panda-re/panda/pull/1194/commits/75a752a9520c3dbe1dfa8031f59bfb2d8949d38c) allows for the support of both rr2 and the older rr format for replay and `rr_print`.

Documentation on record and replay changes can be found here: https://github.com/panda-re/panda/pull/1194/files#diff-9b6b498aa794f6943da36f22f611f6fd99b868b5b98265db52ae1372b9170248

vl.c
```diff
@@ -3145,7 +3145,12 @@ int main_aux(int argc, char **argv, char **envp, PandaMainMode pmm)
         = QSIMPLEQ_HEAD_INITIALIZER(bdo_queue);
 
     // PANDA stuff
-    gargv = argv;
+    gargv = malloc(argc * sizeof(gargv));
```
lacraig2, line 3144: why are we doing this?

panda/python/core/pandare/panda_expect.py
```diff
@@ -334,7 +334,7 @@ def _dump(lines_out, cur_line, line_pos):
         if len(lines_out[-1].strip()):
             self.current_line.append(lines_out[-1])
 
-    def expect(self, expectation=None, timeout=30):
```
lacraig2: Why the change here?

lacraig2 (dismissed):
@zacogen It doesn't look like recordings requested from PyPANDA or the monitor actually create RR2 format recordings.

panda/python/core/pandare/panda_expect.py
zacogen: I found that the setting up the compressed file took slightly longer than the current method. I changed the default time to make sure the tests did not fail because of this.

vl.c
zacogen: The `gargv` is used in `rr_log.c` to save the command line into the compressed folder. I malloced memory, so that the values in `gargv` aren’t overwritten. https://github.com/panda-re/panda/blob/98e2d2d55264fc170f63e880b9ea3d4f942f45f2/panda/src/rr/rr_log.c#L1533-L1545

panda/docs/manual.md
```diff
@@ -429,8 +429,9 @@ end_record:
     so be careful to choose a unique name.
 
     The recording log consists of two parts: the snapshot, which is
-    named `<name>-rr-snp`, and the recording log, which is named
-    `<name>-rr-nondet.log`.
+    named `<name>-rr-snp`, the recording log, which is named
+    `<name>-rr-nondet.log`, and the command line command used to initiate 
+    PANDA, which is named `<name>-rr.cmd`.
```
AndrewFasano: Since it says "two parts" before, should t his maybe say something more like "Also the command line used to run PANDA is stored in `<name>-rr.cmd`

zacogen: Good catch, I need to change to "three parts". Though could also do separate sentence to emphasize two "core" parts, which are necessary for replay. Then another extra file, which can assist in setting up a replay correctly, by reminding you which arguments you used, and also can be used with retroactively creating an `rr2` file.

PR #1205, opened 2022-07-14
fixed arm osi bug

off-by-1-error:
Removed use of `per_cpu_offset_0_addr` in computation of current task_struct pointer for arm kernels


lacraig2:
@off-by-1-error have you checked that this doesn't break other osi_linux profiles?

off-by-1-error:
I tested it using the [osi_syscalls.py](https://github.com/panda-re/panda/blob/dev/panda/python/examples/osi_syscalls.py) script for the i386, x86_64, arm, aarch64, mips, and mipsel default images. As far as I can tell, they behave identically before and after the fix is applied

PR #1208, opened 2022-07-27
Explicitly enable taint for python unicorn examples

mariusmue:
It seems pypanda now requires that tainting is explicitly enabled, at least for the unicorn samples - this PR adds the according lines of code, to prevent error messages like:

```
[...]
    raise Exception("taint2 must be loaded before tainting values")
Exception: taint2 must be loaded before tainting values
```


AndrewFasano:
Thanks Marius!

PR #1211, opened 2022-09-01
panda_*_memory_write now raises an exception in PyPANDA

zestrada:
PyPANDA documentation states that {virtual,physical}_memory_write return a bool, but those functions pass the value returned by the libpanda memory_write_external functions back to the caller. The memory_write_external functions return err < 0 on failure and 0 on success, however. This PR changes the PyPANDA _memory_write function to instead check the return value from the libpanda memory_write_external function and raise an exception on failure, similar to _memory_read.

Tried to do my due diligence by searching for existing code that expects the behavior prior to this PR. Only example I could find of a return value getting checked in the panda code itself was when set_arg attempts a write to the stack in arch.py (corrected in the same commit here).  On failure, set_arg raises a ValueError exception so the proposed change preserves behavior.

These plugins would technically need to be updated:
https://github.com/panda-re/pypanda-plugins/blob/main/src/pandarepyplugins/FileHallucinator.py
https://github.com/panda-re/pypanda-plugins/blob/main/src/pandarepyplugins/rwreplace.py

### DIFF
--- a/panda/python/core/pandare/arch.py
+++ b/panda/python/core/pandare/arch.py
@@ -199,8 +199,7 @@ class PandaArch():
         stack_idx = int(argloc.split("stack_")[1])
         stack_base = self.get_reg(cpu, self.reg_sp)
         offset = reg_sz * (stack_idx+1)
-        if self.panda.virtual_memory_write(cpu, stack_base + offset, val):
-            raise ValueError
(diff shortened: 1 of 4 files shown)


AndrewFasano:
This is a good idea - looks like CI got messed up so I'll close and reopen so it actually runs

PR #1223, opened 2022-09-28
MIPS Hypercalls

lacraig2:
Adds hypercalls to MIPS.

hypercalls are defined for two sets of instructions:

Conditional movs whose destination is register 0. For example:
```
movz $0, $1, $2
```

This instruction works in MIPS ISA V4 onward in all privilege levels. 

We also support mfc0, but this only works in some privilege levels.
```
mfc0 $0, $#
```

This is an attempt to move from any coprocessor register into register zero. In QEMU this is represented as a NOP (with no side effects) because the zero register ($0) is hardcoded to always have the value 0.

This makes the operation highly effective as a mechanism to indicate from the guest that a message is available for the host.

### DIFF
--- a/panda/include/panda/callbacks/cb-helper-defs.h
+++ b/panda/include/panda/callbacks/cb-helper-defs.h
@@ -14,6 +14,6 @@ PANDAENDCOMMENT */
 DEF_HELPER_1(panda_insn_exec, void, tl)
 DEF_HELPER_1(panda_after_insn_exec, void, tl)
 
-#if defined(TARGET_ARM)
+#if defined(TARGET_ARM) || defined(TARGET_MIPS)
 DEF_HELPER_1(panda_guest_hypercall, void, env)
 #endif
--- a/panda/include/panda/callbacks/cb-helper-impl.h
+++ b/panda/include/panda/callbacks/cb-helper-impl.h
@@ -35,7 +35,7 @@ void HELPER(panda_after_insn_exec)(target_ulong pc) {
     }
 }
 
-#if defined(TARGET_ARM)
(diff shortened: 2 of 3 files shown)


AndrewFasano:
When you have a chance can you also add this info to the manual in the PANDA_CB_GUEST_HYPERCALL callback documentation?

AndrewFasano:
@lacraig2 would this syntax work for mips64? I was thinking the mips64 c code to issue a hypercall would be something like:

```C
#ifdef CONFIG_64BIT
typedef uint64_t arg_t;
#else
typedef uint32_t arg_t;
#endif

static inline void my_hypercall(arg_t num, arg_t a1) {
#ifdef CONFIG_MIPS
  #ifdef CONFIG_64BIT
    asm volatile(
      "dmovz $0, %0, %1"  /* use 64-bit instruction and registers */
      :
      : "r" (num), "r" (a1)
      );
  #else
// ...
#endif
```

but perhaps a movz would just work (does that only work on 32-bit args?).

lacraig2:
@AndrewFasano I can't seem to find any documentation for `dmovz` in the MIPS64 ISA docs.

The implementation we have for mips64 hypercalls is here: https://github.com/panda-re/hypernvram/blob/1671595ed9231cbf869e4580d53a8eb4a0a676fa/include/hypercall.h#L96-L127

I can confirm that the `movz $0, $0, $0` instruction triggers the same hypercall functionality on MIPS64.

It also appears that the `move` instructions in that code carry with them 64-bit registers.

In reality the `move` instruction is a pseudo-instruction in both ISAs. In 32-bit MIPS it generates a `lui` and `li` pair.

PR #1227, opened 2022-10-14
Set replay_name in rr_do_begin_replay

zacogen:
replay_name is set in vl.c, but could be used in other places, such as in plugins. By setting it in `rr_do_begin_replay` we make it available to use for the CLI and PyPANDA.

### DIFF
--- a/panda/src/rr/rr_log.c
+++ b/panda/src/rr/rr_log.c
@@ -63,6 +63,8 @@
 /******************************************************************************************/
 /* GLOBALS */
 /******************************************************************************************/
+extern const char *replay_name;
+
 // record/replay state
 rr_control_t rr_control = {.mode = RR_OFF, .next = RR_NOCHANGE};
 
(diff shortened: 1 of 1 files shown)


lacraig2:
@zacogen Why is `rr_control.name` insufficient? https://github.com/zacogen/panda/blob/e59ee230fa65b240f54fed3886ad01f5ce166833/panda/include/panda/rr/rr_types.h#L61-L67

Separately, it seems like we're not using `set_replay_name` internally. https://github.com/zacogen/panda/blob/e59ee230fa65b240f54fed3886ad01f5ce166833/vl.c#L3096-L3098

I think we should make changes here, but I'd like to see some consistency out of it.

AndrewFasano:
Sorry @zacogen I might've given you bad advice now that I see Luke's feedback. It looks like the current replay name is already stored in `rr_control.name` which you can access by including [panda/rr/rr_types.h](https://github.com/panda-re/panda/blob/dev/panda/include/panda/rr/rr_types.h#L69).

When we talked offline I thought the `replay_name` variable was used `extern` in existing plugins, not just in the ones you were working in.

zacogen:
@AndrewFasano no worries!

Closing, solved by: https://github.com/panda-re/panda/pull/1229

PR #1231, opened 2022-10-17
Fix compiler error for NULL va_list

domenukk:
Compiling on aarch64 doesn't accept NULL as va_list, so I added a wrapper function.


lacraig2:
@domenukk Cool! I tried to do this as well quite recently. I have a [branch](https://github.com/panda-re/panda/compare/arm64_host) you're welcome to look at if you're looking at fixing stuff up for arm64. It's pretty basic, but this fix was in there.

PR #1251, opened 2022-11-10
compress ida taint2 labels in script

LauraLMann:
The main thing was to modify the ida_taint2.py IDA script that imports ida_taint2 plugin output to compress the label list that is added to the comments.  For example the list [2, 3, 4, 5, 6, 20, 22, 30, 31, 32] would instead be shown as [2-6, 20, 22, 30-32].  It uses the same algorithm as the ida_taint2_plugin.py IDA plugin.
Also corrected a typo in the ida_taint2 README.md, and corrected several other plugin README.md files that no longer agreed with the actual behavior of the plugins they documented.  These were mostly related to the libosi work deleting various Windows OS specific OSI plugins, and taint reports being at the instruction level instead of the block level.


AndrewFasano:
Thanks for the updates!

PR #1255, opened 2022-11-29
Update docker rr2 dependencies

zacogen:


AndrewFasano:
So is the problem that these are runtime dependencies, not build-time dependencies?

zacogen:
> So is the problem that these are runtime dependencies, not build-time dependencies?

Yes, the dependencies are needed at runtime to package up the RR2 recordings.

PR #1267, opened 2023-01-11
RR Upgrade

off-by-1-error:
Because replays do not depend on the state of peripheral devices, the snapshot generated by a PANDA recording will now only write the contents of the CPU and RAM to the snapshot file. 

The process of loading a replay is unchanged, so old recordings will work the same way they did before.

Tested on x86, x64, arm, mips, and mipsel.

### DIFF
--- a/include/migration/savevm.h
+++ b/include/migration/savevm.h
@@ -1,1 +1,4 @@
 extern int panda_is_in_record;
+
+void set_rr_snapshot(void);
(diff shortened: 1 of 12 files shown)


AndrewFasano:
We had the PR open for a while and solicited feedback, nobody complained and it passed all our tests, so I'm assuming everything is great with this PR. If I'm wrong someone will complain soon and it will be my fault.

PR #1269, opened 2023-01-13
On Taint Propagation, Convert Network Transfer Taint RAM Addresses to Offsets

LauraLMann:
This pull request is mainly about fixing a problem with propagating taint in the taint2 plugin that occurs sometimes when the guest has more than 4G of RAM.  We have a 64-bit Windows 7 guest with 4G RAM that does not have contiguous RAM, which results in the raw RAM addresses needing to be converted to RAM offsets before they can be used to propagate taint on network transfers.
These changes also enhance the taint2 debug output, as that was invaluable in tracking down this issue.
Also, now that someone added OSI support for 64-bit Windows 7, there is no reason for the callstack_instr plugin to forbid the use of the threaded stack type for recordings from such a guest.  As it was useful in testing these changes, that update was made too in this pull request.


jamcleod (approved):

LauraLMann:
Thanks for turning the tabs into spaces @jamcleod - I didn't INTEND to check in tabs :-(.

PR #1271, opened 2023-01-19
OSI Linux: bugfix for big-endian pcomp lengths

lacraig2:
PR #1253 introduces a subtle bug on big endian guests. It removed a `fixupendian` in `read_dentry_name`. https://github.com/panda-re/panda/commit/a5ac10a8bca7df2df835e923ed51038b59f400ec#diff-9659c43f5ac2d777134e55906094d1affea2b7db78d5e29c2cb5d8064f792d31L443

The nature of this bug made it such that small buffers turned into massive buffers, but kept the program running (e.g. 3 bytes is now 0x3000000 bytes).

I noticed this bug by observing very slow runtimes of follow-on plugins in mips big endian targets.

I was able to determine that #1253 was the culprit using a `git bisect run` with the following script which rebuilds PANDA on revision and times out after 30 seconds.

```xonsh
#!/bin/xonsh
cd build
nproc = ![nproc].out.strip()
make -j@(nproc) clean
../build.sh
cd ..
cd panda/python/core
python3 setup.py develop
cd ../../..


import glob, os, sys

rm -rf logo.png

ret =!(timeout 30s python hook_symbol.py mips)

if ret.returncode == 124:
    echo "timeout"
    sys.exit(2)
elif ret.returncode == 2:
    echo "didn't run"
    sys.exit(1)
else:
    echo "runs fine"
    sys.exit(0)

```

### DIFF
--- a/panda/plugins/osi_linux/osi_linux.h
+++ b/panda/plugins/osi_linux/osi_linux.h
@@ -470,7 +470,11 @@ static inline char *read_dentry_name(CPUState *env, target_ptr_t dentry) {
               printf("Warning: OSI_linux Unhandled pcomp value, ignoring\n");
               break;
         }
-        assert(pcomp_length <= PATH_MAX);
+
+        if (pcomp_length > PATH_MAX){
+            printf("Error: OSI_linux pcomp length %d exceeds PATH_MAX. Check endianness.\n", pcomp_length);
+            assert(pcomp_length <= PATH_MAX);
+        }
         pcomp_length += 1; // space for string terminator
 
         if (pcomp_capacity < pcomp_length) {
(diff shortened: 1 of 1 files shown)


AndrewFasano (approved):

panda/plugins/osi_linux/osi_linux.h
```diff
@@ -469,6 +470,7 @@ static inline char *read_dentry_name(CPUState *env, target_ptr_t dentry) {
               printf("Warning: OSI_linux Unhandled pcomp value, ignoring\n");
               break;
         }
+        assert(pcomp_length <= PATH_MAX);
```
AndrewFasano: Maybe worth adding a debug message to go with this in case it's ever triggered? It could be good to distinguish between an unexpectedly long path and a flipped endian length. Or this is probably fine as is.

AndrewFasano (approved):

panda/plugins/osi_linux/osi_linux.h
lacraig2: Added new if statement and debug message.

PR #1277, opened 2023-02-01
Addings support for mips32 to asidstory

elishevaz:
Modifying asidstory to only error out with MIPS64, instead of all MIPS as discussed on the slack.


AndrewFasano:
Thanks for fixing this! The change looks great, I'll just let the CI finish before merging.

PR #1284, opened 2023-02-27
Remove insecure evals in pypanda

jamcleod:
Switch from using `eval` for dynamic accesses to using `getattr`


AndrewFasano:
Thank you so much

AndrewFasano (approved):

PR #1286, opened 2023-03-02
added py3 bytestring fix to rrpack scripts

whoismissing:
Tested on Python 3.8.10 in Ubuntu 20.04 WSL2.

The `rr*pack.py` scripts required some small fixes changing strings -> byte-strings in order for the scripts to run using `python3`.

```sh
$ python3 rrpack.py poc
Packing RR log poc with 1048580 instructions...
  File "rrpack.py", line 39, in <module>
    outf.write(RRPACK_MAGIC)
TypeError: a bytes-like object is required, not 'str'
```

The lack of byte-string prefix makes the check `if magic != RRPACK_MAGIC:` fail because `b"PANDA_RR" != "PANDA_RR"`

```sh
$ python3 rrunpack.py poc.rr
poc.rr is not in PANDA Record/Replay format
```

### DIFF
--- a/panda/scripts/rrpack.py
+++ b/panda/scripts/rrpack.py
@@ -5,7 +5,7 @@ import subprocess
 import struct
 import hashlib
 
-RRPACK_MAGIC = "PANDA_RR"
+RRPACK_MAGIC = b"PANDA_RR"
 
 # PANDA Packed RR file format (all integers are little-endian):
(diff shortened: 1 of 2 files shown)


AndrewFasano:
Thanks for the contribution! I'll try to take a look at those issues you opened/commented on soon :)

PR #1289, opened 2023-03-13
Add syscalls2 support for 64-bit Windows 7

LauraLMann:
Added 64-bit Windows 7 profile to syscalls2 plugin.  Also fixed a bug constructing the dso filename for 32-bit Windows 7.
Noticed that the syscalls2 framework was already treating 16 bit system call arguments as 32-bit (with a comment in the code asking if that was correct), so continued that pattern for the 16-bit and 8-bit arguments in the 64-bit Windows 7 profile rather than enhancing the framework.  (None of the current plugins use system calls with 16-bit or 8-bit arguments, so that seemed safe.)
Only provided dummy implementations for the 64-bit Windows 7 get_return_<type> functions specified in the profile, as they are not used anywhere, and are not listed as API, and based on the history of the syscalls2 plugin appear to be dead code leftover from some changes made in 2016.  As I couldn't get a definitive answer from more knowledgeable parties, I didn't want to rip them out of the framework for all the profiles.
Updated file_taint plugin so it can be used with 64-bit Windows 7.
Updated query_file_writes plugin to explicitly reject all Windows OSes, instead of letting syscalls2 reject 64-bit Windows for it.


LauraLMann:
Is there any way to restart the checks?  The only one that's failing is test_installer, but I don't think the syscalls2 changes impacted that.  Sometimes there have been hiccups in the checks, so it could be informative to try again.

AndrewFasano:
Just set them to restart, I agree that that failure should be unrelated - will be happy to merge this shortly, once that last one (presumably) passes.

PR #1290, opened 2023-03-21
cosi: profile path fix, mips support via hw_proc_id, arm support via current_ksp

zestrada:
This PR has two changes for adding MIPS support to COSI:

1. Removes the need for specifying an `-os` parameter to PANDA when specifying a profile with an argument
2. Uses hw_proc_id to get the `current_task` for mips (same as OSI Linux)

Might need to do [something similar for arm](https://github.com/panda-re/panda/blob/dev/panda/plugins/osi_linux/default_profile.cpp#L14-L55)

### DIFF
--- a/panda/plugins/cosi/src/lib.rs
+++ b/panda/plugins/cosi/src/lib.rs
@@ -39,9 +39,9 @@ static ARGS: Lazy<Args> = Lazy::new(Args::from_panda_args);
 #[allow(deprecated)]
 fn symbol_table() -> &'static VolatilityJson {
     SYMBOL_TABLE.get_or_init(|| {
-        let name = get_symtab_name();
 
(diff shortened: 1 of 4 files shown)


panda/plugins/cosi/src/lib.rs
```diff
@@ -39,9 +39,9 @@ static ARGS: Lazy<Args> = Lazy::new(Args::from_panda_args);
 #[allow(deprecated)]
 fn symbol_table() -> &'static VolatilityJson {
     SYMBOL_TABLE.get_or_init(|| {
-        let name = get_symtab_name();
```
jamcleod, line 42: probably worth running `cargo fmt` so there aren't any formatting issues

jamcleod (approved):
looks great!

PR #1292, opened 2023-03-23
dynamic_symbols: improve reliability around symbol hooks

lacraig2:
These changes provide significant upgrades to the way we handle memory regions under dynamic symbols


AndrewFasano:
Kicking CI

PR #1301, opened 2023-04-04
pypanda: added search for osi mapping by addr, other minor tweaks

zestrada:
This PR includes a few minor changes to PyPANDA:
1. Adds `get_current_process` so that users don't have to do `panda.plugins['osi']`.

2. Adds a `get_mapping_by_addr` to search for an OSI mapping by a virtual address.

3. Fixes a docstring typo on `hook_symbol`.

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -12,7 +12,6 @@ import socket
 import select
 import threading
 
-
 import readline
 # XXX: readline is unused but necessary. We need to load python
 # readline, to preempt PANDA loading another version. PANDA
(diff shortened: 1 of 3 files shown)


panda/python/core/pandare/panda.py
```diff
@@ -1687,6 +1699,30 @@ def get_mappings(self, cpu):
...
+            cpu: CPUState struct
+            addr: int
+
+        Returns:
+            OsiModule: the matching OsiModule structure
+            None: on failure
+        '''
+        mappings = self.get_mappings(cpu)
+        for m in mappings:
+            if m.name != self.ffi.NULL:
```
AndrewFasano: Can M itself be null? I know some of these C structures need to be checked at multiple pointers, but I'm not sure about this one.

zestrada: Good point, looks like you can.

Did a quick test with a bbe callback on boot of the x86_64 generic image and you do eventually get a segfault without this check:
```python
if m == self.ffi.NULL:
    continue
```
Will add that to the next commit along with a test against /proc/self/maps

Issue #901, opened 2021-03-16
PyPANDA syscalls2 load-info fails to load shared libraries without LD_LIBRARY_PATH


AndrewFasano:
From the `pandare/panda:latest` docker container, the following script aborts:

```py
from pandare import Panda
panda = Panda(generic="i386")

panda.load_plugin("syscalls2", {"load-info": True})
panda.run()
```

with error
```
PANDA[syscalls2]:E:syscalls2_info.c(load_syscall_info)> syscalls2_dso_info_linux_x64.so: cannot open shared object file: No such file or directory
PANDA[core]:FAILED to load required plugin syscalls2 from /usr/local/lib/python3.8/dist-packages/pandare/data//x86_64-softmmu/panda/plugins//panda_syscalls2.so
```

It appears the pypanda installer has copied the various `syscalls2_dso_info_linux_*.so` libraries into `/usr/local/lib/python3.8/dist-packages/pandare/data/[ARCH]-softmmu/panda/plugins/` but these aren't loaded successfully (even though regular plugins `.so`s like panda_syscalls2.so are in that directory and presumably loaded from there).

A workaround is to set the environment variable directly:
```
LD_LIBRARY_PATH "/usr/local/lib/python3.8/dist-packages/pandare/data/[ARCH]-softmmu/panda/plugins/"

AndrewFasano:
Looks to be caused by syscalls2 directly using dlopen [here](https://github.com/panda-re/panda/blob/master/panda/plugins/syscalls2/syscalls2_info.c#L45) while regular plugin loading does [something more intelligent](https://github.com/panda-re/panda/blob/6f735eaf254974eab73ac9827c63f36c52f5024a/panda/src/callbacks.c#L189-L197) to construct the library path. We should probably make some sort of `panda_find_library` API function to do this the same way in both places.

AndrewFasano:
Still a bug!

AndrewFasano:
Turns out there's still a bug in here. Using the command line with our container, if you set load-info to True, it will be unable to find the dso shared object.

PR #1303, opened 2023-05-09
copy syscalls2 dso_info files to installation packaging area

LauraLMann:
The missing dso information files prevent the syscalls2 load-info option from working when using an installed build of PANDA instead of a development build.


AndrewFasano:
Oh awesome, thanks for tracking this down. I wonder if this could be the root cause of #901 as well!

PR #1310, opened 2023-06-04
Add Python F.L.I.R.T. signatures example

joren485:
### Description
This is the code for a simple project I wrote to get acquainted with PANDA.

I used this code for demonstration purposes in [a blog post about PANDA](https://blog.nietaanraken.nl/posts/pandare-flirt/).

I am not sure what the best place for this code would be (the `python/examples` directory, the pypanda-plugins repo or a separate repo under my own name). Please let me know if you find another location more appropriate.

### DIFF
--- a/panda/python/examples/flirt-signatures/.gitignore
+++ b/panda/python/examples/flirt-signatures/.gitignore
@@ -1,0 +1,5 @@
+binaries/
+images/
+recordings/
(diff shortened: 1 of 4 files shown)


AndrewFasano:
This is awesome work! Thanks for sharing the post and code, it means a lot to us.

PR #1312, opened 2023-06-17
Add YARA rule Python example

joren485:
### Description

This PR adds Python example code that dumps buffers in memory to disk, if they match a YARA rule.
This is another small example I made to learn more about PANDA. I thought it might be nice to share.

This PR also adds a docstring to my previous example code (#1310). If you want, I can move that to a separate PR.


AndrewFasano:
This is awesome, thank you again for the wonderful contributions!

In general, I'm fine with including unrelated things in a single PR so long as they have separate commits.  - I'll merge this once CI finishes.

PR #1315, opened 2023-06-30
Fix RR2-related regression in checkpoint restore

lacraig2:
In #1194 many record/replay related file functions were replaced with equivalent functions supporting the RR2 format with care to ensure that RR1 support remained.

In `checkpoint.c` our changes do not maintain support for RR1 resulting in a segmentation fault: https://github.com/panda-re/panda/pull/1194/files#diff-b559907e401ac3b65e17bad5551e724edc5c3740cc46de90b16271bbac513a81R199

This PR seeks to add back support for RR1 functionality in the same manner that `fseek` is handled in `rr_print.c`:
https://github.com/panda-re/panda/pull/1194/files#diff-08cc650562b2d2414fec98c0fcb5669d59b75b54a5d83cb26e449e3c0a04e734R178


zacogen (approved):
Looks good!

PR #1317, opened 2023-07-10
Fix debug print in osi_linux

lacraig2:
osi_linux has a printf that should be a debug print. This fixes that.

### DIFF
--- a/panda/plugins/osi_linux/osi_linux.h
+++ b/panda/plugins/osi_linux/osi_linux.h
@@ -472,7 +472,7 @@ static inline char *read_dentry_name(CPUState *env, target_ptr_t dentry) {
         }
 
         if (pcomp_length > PATH_MAX){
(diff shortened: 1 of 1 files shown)


AndrewFasano:
I think you wrote that printf so you're the expert, but at the time we merged it, I thought this was a Very Bad Serious Error so we always wanted to print it in case it ever happened. Is it actually triggering sometimes (and not a terrible thing)?

Fine by me to merge either way, I'm just curious.

lacraig2:
Yes. It seems to trigger sometimes for unclear reasons on otherwise functional systems.

When making many requests for lookups these logs seem to happen a lot.

AndrewFasano:
Huh, weird, thanks for explaining. I just reset the CI (seems like it's failing more than usual lately, this time there was some JSON parsing error internal to the github runners? Never saw that before) but it seems like this should be good to merge

PR #1320, opened 2023-07-20
proc_start_linux: 32-bit app fix

MarkMankins:
Update proc_start_linux to support 32-bit applications running on a 64-bit (x64) host.


AndrewFasano:
Thanks for the fix!

PR #1323, opened 2023-08-09
get_stackid is expensive

MarkMankins:
Reduce number of times this function is called.


AndrewFasano:
This is great, thanks!

PR #1324, opened 2023-08-09
fix arm thumb mode return address

Zanatticus:
masked ARM return address


AndrewFasano:
Maybe it would be better to do an `& ~0x1`? but this seems good to me! Will merge after CI passes

PR #1330, opened 2023-08-16
do not remove stopped blocks from stack twice

LauraLMann:
In cpu-exec.c, in the before-block-exec callback for a block, the callstack_instr plugin removes the block from the stack if it is the block that a return instruction returns to.  The problem is if the flag to say the block is interrupted is set between this point and when the block prologue is executed, the block will be Stopped, and then when the block is run for real after dealing with the interruption, then the block is removed from the callstack again (if it is there).  This doesn't happen a lot (our test case had to run about 200 times before we saw it, and of course it happened once during regression testing which prompted a search for a problem).
This problem and a related problem were fixed once before, by two different groups, and the merge together caused a problem which it appears was resolved (in 2020) by removing too much of one of the original solution.


AndrewFasano:
Thanks for the detailed explanation and fix!

PR #1348, opened 2023-09-01
pc_search plugin

FlyingRagnar:
new plugin for seaching for pc values in a replay and outputting guest instruction values.

### DIFF
--- a/panda/plugins/pc_search/pc_search.c
+++ b/panda/plugins/pc_search/pc_search.c
@@ -1,6 +1,6 @@
 /* PANDABEGINCOMMENT
  * 
- * Authors:
+ * Authors: FlyingRagnar (Jim Knapp)
  * 
  * This work is licensed under the terms of the GNU GPL, version 2. 
  * See the COPYING file in the top-level directory. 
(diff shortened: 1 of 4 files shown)


panda/plugins/pc_search/pc_search.c
```diff
@@ -0,0 +1,157 @@
+/* PANDABEGINCOMMENT
+ * 
+ * Authors:
+ *
```
lacraig2, line 4: Please add your name to the authors field.

panda/plugins/pc_search/pc_search.c
```diff
@@ -0,0 +1,157 @@
...
+        counterslog = fopen(out_filename, "w");
+        
+        // print out input parameters to console
+        fprintf(stderr, "%sfirst last instructions only %s\n", PANDA_MSG, PANDA_FLAG_STATUS(first_last_only));
+        for (int l=0; l < num_of_input_pcs; l++) {
+            fprintf(stderr, "%sinput pc %i = 0x%" PRIx64 "\n", PANDA_MSG, l, input_pcs[l]);
+        }
+    }
+
+    panda_enable_precise_pc();
```
lacraig2: `panda_enable_precise_pc` is unnecessary here because the PC value is used in two places: in TCG translation and as a TCG constant argument.

`enable_precise_pc` is generally necessary if you depend on the result of `panda_current_pc`.

lacraig2 (commented):
Thanks for the contribution! I've noted a couple of small things, but overall this looks great.

AndrewFasano:
This is cool, thanks for the PR! It would also be nice if you supported a PC range, but only if you feel like adding that - otherwise it looks good to me once you address Luke's suggestions.

FlyingRagnar:
Pushed an update to add support for searching by a range of pcs.  Also addressed other comments.  Please let me know if there is any other feedback.

AndrewFasano:
Thanks for making those updates! Just merged

PR #1351, opened 2023-09-11
Fix MIPS64BE Syscalls2 (O32) and OSI support

lacraig2:
This PR fixes an issue where MIPS64 did not support the O32 ABI (which it seems to have programs use a lot).

Further, it fixes OSI by mostly using `fixupendian2` in many places and adding a check for pointer size of `target_ulong` in `struct_get`.


panda/plugins/syscalls2/generated/syscalls_numbers.h
```diff
@@ -5174,1081 +5174,1225 @@ namespace syscalls2 {
 	namespace linux {
 		namespace mips64 {
```
AndrewFasano, line 6167: These changes are fine by me as is, but do you think it would be worth trying to encode os -> arch -> ABI for this? Right now we pretend there's a single syscall number for a given syscall on an os+arch but really there can be multiple depending on the ABI. Or perhaps we could change these to arrays? Not sure what would be best. But also happy to leave it as is, I'm not sure if this info is ever really used and it was just as wrong before.

AndrewFasano (approved):

PR #1352, opened 2023-09-11
PyPANDA: directory overrides for PANDA constructor

MarkMankins:
PANDA and PyPANDA make some assumptions about where files are installed that don't currently align with how things are done in my environment.  I updated both to allow for more fine-grained configuration to be specified.  I can rework if there is a better way.  Is there anything else I need to do - since this logic is tightly coupled between PyPANDA and PANDA I wasn't sure if anything else needed tweaked (version numbers?)

Allow caller to specify location of libpanda shared object, pc-bios files, and directory where plugins are installed.  If these arguments are left as their default values PyPANDA will use the previous algorithm for finding these files.

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -61,7 +61,9 @@ class Panda():
             raw_monitor=False, # When set, don't specify a -monitor. arg Allows for use of -nographic in args with ctrl-A+C for interactive qemu prompt.
             extra_args=None,
             catch_exceptions=True, # Should we catch and end_analysis() when python code raises an exception?
-            libpanda_path=None):
+            libpanda_path=None,
+            biospath=None,
+            plugin_path=None):
         '''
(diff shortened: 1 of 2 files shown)


AndrewFasano:
These changes look good to me but I'd love for either @lacraig2 or @jamcleod to double check that the `PANDA_LIB` environment variable changes wont' break anything with panda-rs.

MarkMankins:
The output from pypanda's test generic_tests.py doesn't look like it's what is expected - I don't think that originated with this PR.

AndrewFasano:
Oh wow yeah, that looks very wrong. Thanks for catching it. Since the test is still marked as passing despite it going off the rails, I'm going to assume that has been broken for a long time and not related to this PR. I'll open an issue.

MarkMankins:
If there is a chance that PANDA_LIB might cause a problem I could change that to something else.  PYPANDA_PANDA_LIB perhaps?

AndrewFasano:
Nah, I think PANDA_LIB is fine - nobody complained so I'll merge this now. Thanks!

PR #1358, opened 2023-09-13
Fix #1345: Errors in pypanda shutdown with multiple SIGINTs

### LINKED ISSUE #1345: Errors in pypanda shutdown with multiple SIGINTs
If we go to terminate PyPANDA using ctrl-c (sending SIGINT), we'll get the following error. Should probably add a lock somewhere or a check to avoid starting our shutdown logic twice.

```
Exception ignored from cffi callback <function Panda._setup_internal_signal_handler.<locals>.SigHandler at 0x7efe6efb79d0>:                                                                                                        
  File "/usr/local/lib/python3.8/dist-packages/pandare/panda.py", line 446, in SigHandler                                                                                                                                          
    self.end_analysis()                                                                                                                                                                                                            
  File "/usr/local/lib/python3.8/dist-packages/pandare/panda.py", line 535, in end_analysis                                                                                                                                        
    self.unload_plugins()                                                                                                                                                                                                          
  File "/usr/local/lib/python3.8/dist-packages/pandare/panda.py", line 733, in unload_plugins                                                                                                                                      
    self.pyplugins.unload_all()                                                                                                                                                                                                    
  File "/usr/local/lib/python3.8/dist-packages/pandare/pypluginmanager.py", line 264, in unload_all                                                                                                                                
    for name in self.plugins.keys():                                                                                                                                                                                               
RuntimeError: dictionary changed size during iteration      
```

lacraig2:
This PR primarily fixes #1345 which indicated that the pypluginmanager `plugins` size may change during iteration due to multiple threads.

This can be fixed by changing to a loop which pops items while there remain items to pop rather than forming an iterator.

https://github.com/panda-re/panda/compare/fix_1345?expand=1#diff-a4f554d6018061dd19f397a13397cc37399a0cc4a602f566ce50f7e462dc9075R260-R261

There are a handful of other code cleanup items as well.

fixes #1345

### DIFF
--- a/panda/python/core/pandare/pypluginmanager.py
+++ b/panda/python/core/pandare/pypluginmanager.py
@@ -21,7 +21,7 @@ class _DotGetter(object):
         self.data[k] = v
 
     def __str__(self):
-        return str([x for x in self.data.keys()])
+        return str(list(self.data.keys()))
 
     def __getattr__(self, name):
         raise NotImplementedError("Subclass must implement this virtual method")
@@ -215,7 +215,7 @@ class PyPluginManager:
 
                 bp = self.blueprint(name, __name__, template_folder=template_dir)
                 self.plugins[name].webserver_init(bp)
-                self.app.register_blueprint(bp, url_prefix="/" + name)
+                self.app.register_blueprint(bp, url_prefix=f"/{name}")
 
     def load_all(self, plugin_file, args=None, template_dir=None):
         '''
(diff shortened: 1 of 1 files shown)


AndrewFasano (approved):
I was thinking we'd just use a lock, but this might be cleaner. Other changes look good too!

AndrewFasano:
This introduced a bug where `unload_all` pops items from `self.plugins` and then calls the `unload` function which tries to access the just-popped item from self.plugins.

I'll work on a fix as part of #1357 which I'll expand to include some pypanda fixes.

PR #1362, opened 2023-09-25
Wintrospection: add option to use asid change heuristic

lacraig2:
This PR adds an option to wintrospection to enable/disable the `asid_change_heuristic`. It remains `false` when unspecified/default.

Likely of interest to @MarkMankins

### DIFF
--- a/panda/plugins/wintrospection/wintrospection.cpp
+++ b/panda/plugins/wintrospection/wintrospection.cpp
@@ -777,8 +777,6 @@ void initialize_introspection(CPUState *cpu) {
     return;
   }
 
-  using_asid_change_heuristic = false;
-
   GArray *mods = get_modules(cpu);
   for (int i = 0; i < mods->len; i++) {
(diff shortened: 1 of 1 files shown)


MarkMankins:
It might be worth adjusting the comments on line 766 and 774.  Previously using_asid_change_heuristic would be set to true in those cases.  Now the user would need to know to set the new command line parameter.  Maybe even a warning would be helpful in those cases to alert the user that the command line parameter is now necessary.

MarkMankins:
Thinking about this a little more, is there any reason not to set using_asid_change_heuristic to true for the windows 2000 case and the case where swapcontext isn't found?  That would make this backward compatible with the current functionality.

lacraig2:
@MarkMankins I've added code to set automatically turn on the asid change heuristic for Windows 2000 and print a warning if that option hasn't been set. Let me know if this fits your description.

MarkMankins:
I ran this branch against our regression tests and there were no behavior changes reported.

This one looks good - thanks for the nice update!

PR #1364, opened 2023-09-27
Add targetcmp/callwitharg/findcall plugins

AndrewFasano:
New set of plugins for dynamically analyzing what happens to a known string. We can find values it's compared to and functions that it's passed as an argument to.

(Also includes an unrelated minor pypanda bugfix as a separate commit)


AndrewFasano:
I might expand this to be more PPP-based to provide an "on_call_with_target_arg" callback

PR #1365, opened 2023-09-29
OSI-based Pagewalk

lacraig2:
Based on @AndrewFasano's #1353 this PR allows for ARM32 and MIPS(EL/BE)32 targets with OSI support and 2-level page tables to bypass TLB misses by reading the page table directly.

This PR provides wrappers for 3 standard PANDA function with the additional functionality described:
- `panda_virtual_memory_[read/write]` -> `osi_linux_virtual_memory_[read/write]`
- `panda_virt_to_phys` -> `osi_linux_virt_to_phys`

For each function the OSI wrapper first calls the original PANDA function and only attempts a page table walk if the original function fails.

This PR also adds a Python example of the functionality.

Lastly, this PR changes PyPANDA to use these functions in place of the PANDA functionality when `osi_linux` has been loaded.

### DIFF
--- a/panda/plugins/osi_linux/osi_linux.cpp
+++ b/panda/plugins/osi_linux/osi_linux.cpp
@@ -76,6 +76,7 @@ struct KernelProfile const *kernel_profile;
 
 extern const char *qemu_file;
 static bool osi_initialized;
+static bool pagewalk_enabled;
 static bool first_osi_check = true;
 
 /* ******************************************************************
@@ -817,6 +818,17 @@ target_ulong read_target_ulong(CPUState *cpu, target_ulong addr) {
 }
 
 
+/**
+ * Our current implementation of walk_page_table is based on 2-level page table
+ * for MIPS and ARM. 
+*/
+
+/**
+ * An effort has been made to keep down the repeated code. In this case we 
+ * simply need different methods for pte_offset and pte_valid for MIPS and ARM.
+ * 
+ * This will be somewhat different for 3 and 4-level page tables.
+*/
 #if (defined(TARGET_MIPS) && !defined(TARGET_MIPS64))
 
 #define PGDIR_SHIFT 0x16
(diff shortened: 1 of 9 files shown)


panda/plugins/osi_linux/osi_linux.cpp
```diff
@@ -799,6 +803,168 @@ void on_get_process_ppid(CPUState *cpu, const OsiProcHandle *h, target_pid_t *pp
...
+    return 1;
+}
+
+bool pud_valid(CPUState *cpu, target_ulong pud){
+    return 1;
+}
+
+bool pmd_valid(CPUState *cpu, target_ulong pmd){
+    return 1;
+}
```
AndrewFasano, line 898: Are these NYI methods that you'd want to support later? Not sure what the value of them is right now, some comments might help.

panda/python/examples/osi_pagewalk.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python3
+'''
+This is an example of the functionality that OSI provides for reading 2-level
+page tables.
+
+This does not work with the generic ARM image because it does not have the same
+type of page table. It does, however, work for some ARM images and the MIPS
+default image.
```
AndrewFasano, line 8: Should we make these options configurable and/or make the default work with our generic arm image?

panda/plugins/osi_linux/osi_linux.cpp
lacraig2: Those are the correct implementation for the 2-level page tables we've implemented. I keep them in there for two reasons:
- It leaves us with the infrastructure to handle additional page table levels in the future
- It closely matches the linux kernel reference and makes the implementation easier in the future.

panda/python/examples/osi_pagewalk.py
lacraig2: I have added the `pagewalk` option.

Our generic arm image is so old that it's difficult to find out why it's not working. It's possible that it's a 3-level page table. I don't particularly want to add support for that model at the moment.

panda/plugins/osi_linux/osi_linux.cpp
AndrewFasano: Cool, want to add comments saying that so next time we're hacking on this file we know why they're there?

lacraig2: I did! I guess it just didn't update here. https://github.com/panda-re/panda/pull/1365/files#diff-2cbdbd9fde9ca83251f09a5c05263f1edbb07069e56916884ece0fdcf6fc2180R876

PR #1371, opened 2023-10-11
Panda load plugin enhancements

MarkMankins:
The major new functionality here is in do_check_export_symbols().

I have a rather large C++ panda plugin that I'm trying to split up into separate plugins.  I would like to have a base C++ class in plugin plugin_x.  I would like to have plugins plugin_y and plugin_z extend my C++ class defined in plugin_x to implement custom functionality.

This doesn't work with the current logic for loading plugins, as each plugin loaded doesn't provide symbols to subsequently loaded plugins.

do_check_export_symbols changes that behavior, but only for plugins that opt-in to the new behavior.

In plugin_x, by defining a magic symbol:

```
uint32_t PANDA_EXPORT_SYMBOLS_plugin_x = 1;
```

panda_plugin_x.so will be opened (via dlopen) with the RTLD_GLOBAL flag.  This makes symbols defined in plugin_x available to subsequently loaded plugins.

The majority of the commits in this PR aren't related to this new function.  I tried to clean up the various functions that deal with plugin loading and unloading, mostly in callbacks.c.  It's possible I broke something here that someone was using - I can rework things if we find I did break something.

Major changes (including potential breaking changes in obscure cases):

- Consolidated all plugin information in the panda_plugin struct.  Removed panda_plugins_to_unload array and panda_plugins_loaded array.
- Moved pypanda specific logic into their own functions.
- Removed panda_load_external_plugin.  If anyone is using this please let me know and I'll implement a replacement.
- Replaced most printfs with calls to the appropriate logging api.
- Plugin help output now always goes to stdout (previously it was a mix of stderr and stdout.)
- plugin_name is now required in _panda_load_plugin (previously in some cases it was allowed to be null)
- Previously plugins were prevented from being loaded twice by examining the list of filenames already loaded.  This was changed to examine the list of plugin names instead.

### DIFF
--- a/panda/include/panda/plugin.h
+++ b/panda/include/panda/plugin.h
@@ -67,7 +67,7 @@ void panda_disable_plugin(void *plugin);
 
 // Structure to store metadata about a plugin
 typedef struct panda_plugin {
-    char name[256];        // Currently basename(filename)
+    char *name;            // Plugin name: basename(filename)
     void *plugin;          // Handle to the plugin (for use with dlsym())
     bool unload;           // When true, unload plugin when safe
     bool exported_symbols; // True if plugin dlopened with RTLD_GLOBAL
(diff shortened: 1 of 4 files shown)


MarkMankins:
Rebased to latest dev.

AndrewFasano:
Thanks, these seem like some great changes - I had been hoping to get more eyes on the PR before merging, but I'm sure someone will let us know if this breaks anything.

PR #1374, opened 2023-10-19
ARC migrate

coolkingcole:


.github/workflows/parallel_tests.yml
```diff
@@ -73,6 +73,10 @@ jobs:
 
     steps:
     # Given a container with PANDA installed at /panda, run the taint tests
+    - name: Update
+      run: apt-get update -y
+    - name: Install ssl
+      run: apt-get install -y wget
```
AndrewFasano: I think wget should already be installed in our PANDA containers since it's in our base dependency list (i.e., the runtime dependencies) [here](https://github.com/panda-re/panda/blob/dev/panda/dependencies/ubuntu%3A22.04_base.txt).

coolkingcole: This particular step is ran on the base runner, so I had to add this. If there is a docker image I should be using for this step lmk, I can try that.

PR #1375, opened 2023-10-25
Remove rust artifacts from panda docker container

lacraig2:
This commit removes the artifacts of the rust plugin build process from the docker container.

This one change decreases the size of our container from (uncompressed) 15.155GB to 2.837GB (18% original size).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,7 +80,13 @@ WORKDIR /panda/
 
 #### Install PANDA + pypanda from builder - Stage 4
 FROM builder as installer
-RUN  make -C /panda/build install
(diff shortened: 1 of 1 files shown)


AndrewFasano:
Wow, nice find. Looks good to me but I'd want @jamcleod to also approve in case this breaks something important?

jamcleod:
> Wow, nice find. Looks good to me but I'd want @jamcleod to also approve in case this breaks something important?

I instructed Luke on what to delete and where. Slight change in syntax over the command I gave him but looks good to me.

jamcleod (approved):

PR #1385, opened 2023-11-20
Fix variety of problems in coverage plugin

LauraLMann:
Added fetches for the FS and DS registers, or the plugin would crash when trying to instrument an instruction using one of those segment registers.
Added back in the response to plugin_cmd help (which mysteriously disappeared a few years ago).
Fixed pointer arithmetic problems when fetching register values, which resulted in garbage values being used when trying to read any register except the first one in the regs list.
Fixed calculations of jump destination addresses when the 'base' register is RIP.
Made some minor fixes in the README file.

### DIFF
--- a/panda/plugins/coverage/EdgeInstrumentationDelegate.cpp
+++ b/panda/plugins/coverage/EdgeInstrumentationDelegate.cpp
@@ -317,6 +317,35 @@ static void jmp_mem_indirect(EdgeState* edge_state, CPUState *cpu,
     });
 }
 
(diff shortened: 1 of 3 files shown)


LauraLMann:
I really don't see why the test_installer CI test should fail due to the changes for this PR, but it's done so 2X so far.  I suspect this is one of the intermittent technical difficulties.

AndrewFasano:
Yeah, I don't think that CI failure has anything to do with you, seems broken in all the latest PRs. I'll merge this as is. Thanks!

PR #1387, opened 2023-11-20
Arc resource sharing

coolkingcole:
testing out different ways to speed up builds.


coolkingcole:
This branch should be good to merge as long as the tests pass. Losing a lot of time having to pull the container for each test. Could be sped up if the tests could be lumped into fewer jobs or if the jobs can share a docker cache on the server.

PR #1396, opened 2023-12-15
added plugin track_proc_hc and associated PPP callbacks

off-by-1-error:
Adds the `track_proc_hc` plugin for tracking process changes, new process execs, and VMAs, as reported by the linux kernel via hypercalls. 

Intended for use with [https://github.com/panda-re/linux/tree/4.10_hc](https://github.com/panda-re/linux/tree/4.10_hc)


AndrewFasano:
What are your thoughts on keeping this plugin out of tree? I don't think it's generally useful for panda users since it's tied to a specific kernel

PR #1399, opened 2023-12-16
Support building debian packages for PANDA and pip packages for PyPANDA

AndrewFasano:
This PR aims to give us a sane way to distribute PANDA and PyPANDA.

# Debian packaging
Add support for creating debian packages for PANDA using our existing docker containers. 

Running `./panda/debian/setup.sh Ubuntu [VERSION]` with VERSION set to 18.04/20.04/22.04 will create a pandare.deb package that can be installed with `apt install pandare.deb`. This PR does not integrate this with CI but I think that's something we should explore in the future.

# PyPANDA packaging
This PR refactors the PyPANDA install to remove the panda binaries, libraries and plugins  and limits it to just python scripts, the qcows.json, and the autogenerated headers. As a result the package is about 50x smaller and no longer gets out of sync with the host (non-python) version of PANDA.

Instead of including these binaries in the package, this now uses the binaries at the expected paths from the debian package package. If a user installs with `python3 setup.py develop` it should continue to behave as normal and pull libpanda libraries and compiled plugins from the panda build path.

To create a source dist in panda/panda/python/core you run:
```
python3 setup.py install
python3 setup.py sdist
# distribution created in ./dist/pandare-[version].tar.gz
```

# Known issues
* The PyPANDA package used to copy llvm .bc files which might be required for taint. Maybe those should be installed by the debian packages? I'm not sure where they're used.

# Example
The following dockerfile installs panda and pypanda in Ubuntu 20.04 from the panda .deb and the pypanda sdist:
```
FROM ubuntu:20.04

ENV DEBIAN_FRONTEND=noninteractive

# Install panda deb
COPY pandare.deb /host/
RUN apt update && apt install -q -y /host/pandare.deb

# Install pypanda
COPY pandare-0.1.2.0.tar.gz /host/
RUN pip install /host/pandare-0.1.2.0.tar.gz

# Download a qcow and test PyPANDA
COPY syscalls.py /host
RUN python3 -m pandare.qcows i386
RUN python3 /host/syscalls.py i386
```


AndrewFasano:
Seems like this needs more work to resolve plugin paths:
> PANDA[core]:E:callbacks.c(try_open_libpanda)> PANDA[core]:Failed to load libpanda: /usr/local/bin//arm-softmmu/libpanda-arm.so: cannot open shared object file: No such file or directory from /usr/local/bin//arm-softmmu/libpanda-arm.so
> PANDA[core]:E:callbacks.c(_panda_require)> PANDA[core]:FAILED to load required plugin syscalls2 from /usr/local/lib/panda/arm/panda_syscalls2.so

AndrewFasano:
The CI failure [here](https://github.com/panda-re/panda/actions/runs/7241647856/job/19726288914) found an unrelated bug introduced in #1327 when PLogReader is unavailable. I added a fix to this PR

PR #1400, opened 2023-12-18
Tainted instruction plugin optimizations.

MarkMankins:
- Made globals static.
- Added "suppress_redundant_taint_reports" option. -- When enabled, won't output the same PC multiple times consecutively. -- Only used when pandalog is not enabled.
- In the for loop that computes num_tainted in taint_change: -- Exit early if not pandalogging.
-- Further looping in this case isn't necessary as output doesn't change.
- Removed redundant ifs.
- Fix memory leak (ti->taint_query allocated via malloc was not freed.)
- Save "pc = 0x..." generated string in case it can be reused.
- Moved assignments to last_asid and last_pc to avoid unnecessary assignments.

### DIFF
--- a/panda/plugins/tainted_instr/README.md
+++ b/panda/plugins/tainted_instr/README.md
@@ -1,4 +1,4 @@
-Plugin: tainted_instr
+Plugin: tainted\_instr
 ===========
 
 Summary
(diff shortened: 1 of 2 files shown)


AndrewFasano:
CI is a bit broken right now, but this looks good, thanks!

AndrewFasano:
@coolkingcole I don't think we've seen this CI issue before - It was trying to push a new container on the PR and getting permission denied. The fact that it didn't have permissions seems good, but it probably shouldn't have tried or reported a failure about it.

coolkingcole:
I believe this is a user permissions error related to to role in the org. We should add additional branch protection to avoid malicious workflows. For now we have the setting [Fork pull request workflows from outside collaborators](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) set to the most restrictive setting which restricts Outside Collaborators, which @MarkMankins is.

AndrewFasano:
@coolkingcole makes sense that we'd restrict roles - but if someone with more permissions opened a PR, we wouldn't want the PR itself to publish a new release. We'd only want that after the PR is merged, right?

coolkingcole:
If we want that, we could change it to something like this
```
on:
  push:
    branches:
      - dev
      - stable
  pull_request:
    types:
      - closed
```

Issue #922, opened 2021-03-27
Makefile:687: *** new RPATH too long - cannot adjust .so files for installation. Stop.


username7778:
hello
here is log: https://gist.github.com/username7778/232accfaa6159b9e0d4f6420ae435c25
how to fix?

thx

lacraig2:
There was a PR about this not too long ago #533. Maybe @LauraLMann has some advice?

LauraLMann:
It is possible to write PANDA plugins which use the RPATH in an .so file to load other PANDA plugins.  When PANDA is built, the RPATH in each .so file is set to the location in which the plugin was built.  But if the "make install" option is used, the plugins will be installed somewhere else and thus not be able to be found via the RPATH.  So #533 changed the "make install" option to use the chrpath utility to modify the RPATH in the .so files to correspond to the location where the plugins will be installed.  The problem is, this cannot be done if the desired new RPATH would be longer than the original RPATH.
So, you can change where you build PANDA so the original RPATH is longer than the path to the install location.
Or, you can change where you install PANDA so the new RPATH will be shorter than the path specified by the old RPATH.

AndrewFasano:
I've seen a PANDA fork that does this a little differently, and perhaps it's something we'd want to incorporate- they modified Makefile.target (Around line 233) file to add some LDFLAGS for the rpath:

```
$(PANDA_PROG): LDFLAGS+=-shared -Wl,-rpath '-Wl,$$ORIGIN:$$ORIGIN/../lib'
```

This then resolves the variable $ORIGIN at runtime (instead of compile time) which contains the path to the executable being run. [This post](https://stackoverflow.com/a/42347206) has more details and looks like it would work well here and remove the need for `chrpath`

PR #1402, opened 2023-12-19
Add flake.nix

be32826:
Add a Nix flake for PANDA.

After installing Nix with flakes support (e.g. with https://github.com/DeterminateSystems/nix-installer, the official installer doesn't enable flake support by default),

- `nix build` (in directory) or `nix build github:panda-re/panda` anywhere - Build and install PANDA into the Nix store and create a `result/` symlink to the installation.
- `nix shell` (in directory) or `nix shell github:panda-re/panda` anywhere - Enter a temporary development shell with the PANDA binaries in `PATH`.
- `nix develop` - Enter a development shell with all the build dependencies of PANDA (for debugging the build process)
- `nix flake update` - Update pinned dependencies in `flake.lock` lockfile.
- `nix bundle --bundler github:NixOS/bundlers#toDEB github:panda-re/panda` - Generate a .deb package for PANDA (currently puts a broken symlink in /usr/bin/panda, might need some tweaks)
- `nix bundle --bundler github:NixOS/bundlers#toRPM github:panda-re/panda` - Generate an RPM package
- `nix bundle --bundler github:NixOS/bundlers#toDockerImage` - Generate a minimal Docker image with PANDA

This doesn't have to replace the existing Docker container. It could just be an alternative install option like the install scripts in [panda/scripts](https://github.com/panda-re/panda/tree/dev/panda/scripts).


AndrewFasano:
This looks interesting! Why does it delete the cargo locks, are those something we want to track in git? Might be of interest to @jamcleod

be32826:
> This looks interesting! Why does it delete the cargo locks, are those something we want to track in git? Might be of interest to @jamcleod

It puts all the Rust plugins in a cargo workspace, so they share one lockfile instead of having their own separate ones. This makes it much easier to package in Nix, but it should also reduce build times/size because it makes the Rust plugins share a build artifact directory.

be32826:
Also, I'm not sure why the RPATH length check exists or if removing it causes problems.

MarkMankins:
My guess is that the rpath check is there to prevent one from going through an entire build only to have chrpath fail during the "make install" phase.  From the BUGS section in "man chrpath":

>       This  program  cannot create an RPATH tag if the ELF does not have one,
>       and it can only replace an RPATH with one of equal or  shorter  length.
>       (Moving  ELF sections following a lengthened string table would be dif‐
>       ficult and error-prone at best, and is  sometimes  outright  impossible
>       due to issues like limited ranges in jump instructions.)

AndrewFasano:
The RPATH stuff was added in #533 to fix an issue with `make install` but we've talked about a better way to resolve plugin paths with $ORIGIN [here](https://github.com/panda-re/panda/issues/922#issuecomment-809442676). I never figured out how to get that working though.

Now that we've merged #1399 perhaps we could revisit that and come up with something better that would work for development environments, `make install` environments and our .deb package.

MarkMankins:
Oh!  I didn't know about $ORIGIN.  Good write-up here about how that can be used:

https://www.baeldung.com/linux/rpath-change-in-binary  (Section 6)

be32826:
I moved the Makefile patch into the `flake.nix` for now so it won't change the behavior of normal builds.

jamcleod:
Thank you, sharing a lockfile/build directory has been something I've been interested in for the Rust plugins for a while.

AndrewFasano:
Since this doesn't break the rust plugins, I don't see any reason not to merge (though I'll admit I still haven't figured out how to use flakes)

PR #1404, opened 2023-12-20
automatic pypanda ppp header setup

off-by-1-error:
Removes the need for manual edits to `generate panda_datatypes.py` when setting up pypanda-compatible PPP callbacks.


lacraig2 (approved):
I've made some small changes and checked that this produces identical outputs as the original. I've also rebased it.

Should be ready to merge.

PR #1408, opened 2024-01-05
Add PANDA_PLUGIN_PATH

be32826:
Add a colon-separated `PANDA_PLUGIN_PATH` environment variable for specifying plugin directories similar to `PATH`, `PYTHONPATH`, and `LD_LIBRARY_PATH`.

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -106,7 +106,10 @@ class Panda():
         self.qlog = QEMU_Log_Manager(self)
         self.build_dir = None
         prev_path = environ.get("PANDA_PLUGIN_PATH")
-        environ["PANDA_PLUGIN_PATH"] = plugin_path + (":" + prev_path if prev_path else "")
+        if plugin_path and prev_path:
+            environ["PANDA_PLUGIN_PATH"] = f"{plugin_path}:{prev_path}"
+        elif plugin_path:
+            environ["PANDA_PLUGIN_PATH"] = plugin_path
 
         self.serial_unconsumed_data = b''
 
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -105,7 +105,7 @@ class Panda():
         self.catch_exceptions=catch_exceptions
         self.qlog = QEMU_Log_Manager(self)
         self.build_dir = None
-        prev_path = environ["PANDA_PLUGIN_PATH"]
+        prev_path = environ.get("PANDA_PLUGIN_PATH")
         environ["PANDA_PLUGIN_PATH"] = plugin_path + (":" + prev_path if prev_path else "")
 
         self.serial_unconsumed_data = b''
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -105,7 +105,8 @@ class Panda():
         self.catch_exceptions=catch_exceptions
         self.qlog = QEMU_Log_Manager(self)
         self.build_dir = None
-        environ["PANDA_PLUGIN_PATH"] = f"{plugin_path}:{environ['PANDA_PLUGIN_PATH']}"
+        prev_path = environ["PANDA_PLUGIN_PATH"]
+        environ["PANDA_PLUGIN_PATH"] = plugin_path + (":" + prev_path if prev_path else "")
 
         self.serial_unconsumed_data = b''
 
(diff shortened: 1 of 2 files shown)


panda/python/core/pandare/panda.py
```diff
@@ -259,19 +263,12 @@ def __init__(self, arch="i386", mem="128M",
...
-            elif isdir(rel_dir):
-                self.plugin_path = rel_dir
-            else:
-                raise ValueError(f"Could not find plugin path. Build dir={build_dir}")
-        return self.plugin_path
+    def get_plugin_path(self, plugin_name):
+        plugin_name_ffi = self.ffi.new("char[]", bytes(plugin_name, "utf-8"))
+        plugin_path_ffi = self.libpanda.panda_plugin_path(plugin_name_ffi)
+        plugin_path = self.ffi.string(plugin_path_ffi).decode("utf-8")
+        # TODO: self.libpanda.g_free(plugin_path_ffi)
```
jamcleod, line 270: leftover TODO comment

panda/python/core/pandare/panda.py
```diff
@@ -259,19 +263,12 @@ def __init__(self, arch="i386", mem="128M",
...
-
-            if build_dir == "/usr/local/bin/":
-                # Installed - use /usr/local/lib/panda/plugins
-                self.plugin_path = f"/usr/local/lib/panda/{self.arch_name}"
-            elif isdir(rel_dir):
-                self.plugin_path = rel_dir
-            else:
-                raise ValueError(f"Could not find plugin path. Build dir={build_dir}")
-        return self.plugin_path
+    def get_plugin_path(self, plugin_name):
```
jamcleod, line 266: this doesn't seem to cover all the cases from before, was the `/usr/local/bin` special-casing redundant? it also doesn't appear to perform error propagation/checking, so this is a breaking change (as catching ValueError to handle missing plugins isn't possible)

panda/src/callbacks.c
```diff
@@ -366,6 +367,20 @@ char* resolve_file_from_plugin_directory(const char* file_name_fmt, const char*
     }
     g_free(plugin_path);
 
+    // Finally, try the directories in PANDA_PLUGIN_PATH
+    gchar **panda_plugin_path = g_strsplit(g_getenv("PANDA_PLUGIN_PATH"), ":", -1);
+    for (gchar **dir = panda_plugin_path; *dir; dir++) {
+        gchar *plugin_path = attempt_normalize_path(g_strdup_printf(
+                                    "%s/%s/%s", *dir,
+                                    TARGET_NAME, name_formatted));
+        if (TRUE == g_file_test(plugin_path, G_FILE_TEST_EXISTS)) {
```
jamcleod, line 376: minor nit but don't think you need to actually check against glibc TRUE here. just checking for non-zero return is a superset of checking for glib TRUE, so there's no real reason to write in this style.

jamcleod (commented):
Overall design question: using some combination of `PANDA_PATH` and `PANDA_PLUGIN_DIR` depending on needs has been the way of handling this usecase in the past. Is there a reason we're making a new variable with overlapping usecases rather than fixing the previous implementation?

The main differences I see here are:
- supports providing multiple paths (beneficial, but I don't think requires adding a new environment variable)
- supports absolute paths (the current code does support this, albeit inconsistently and definitely has bugs for usecases I haven't personally considered. however to me this would be a reason to fix the current code rather than)

This is mostly coming from a place of wanting to avoid having 5 different ways of doing this, each with their own little bugs, with documentation being unclear which one to use. It's also possible that incrementally deprecating the old system is the correct move, all up for consideration.

not saying these are reasons not to merge, just questions worth considering before merging. also be sure to notify me when the design has been ironed out so I can ensure panda-rs also receives a corresponding update.

panda/python/core/pandare/panda.py
be32826: I'm pretty sure this is a memory leak without the free, but CFFI doesn't seem to find the `g_free()` function, which is weird because other parts of the code find `g_malloc0()`. I don't know what's going on here.

panda/src/callbacks.c
be32826: I can change it but all the other tests in the function use the same style so it would feel inconsistent.

be32826:
> Overall design question: using some combination of `PANDA_PATH` and `PANDA_PLUGIN_DIR` depending on needs has been the way of handling this usecase in the past. Is there a reason we're making a new variable with overlapping usecases rather than fixing the previous implementation?

Mostly because I'm afraid of breaking backwards compatibility and I don't know how all the different cases apply to how PANDA is used in practice. Do you think it would be better to make `PANDA_PLUGIN_DIR` support colon-separated multiple directories instead?

PR #1410, opened 2024-01-08
Add error message when volatility profile is empty

jamcleod:
Previously this gave an unhelpful error message, leading to users of MIPS64.


lacraig2:
This looks good.

PR #1412, opened 2024-01-10
Fixed bug in use of string search file

tleek:
There was a bug that meant if you used a search strings file with other than exactly one search string (two, three, etc) then the relevant mem callbacks would never get registered meaning nothing would get searched for. Basically, the `num_strings==1` test was outside a loop (the one that increments num_strings) when it should have been inside it.

### DIFF
--- a/panda/plugins/stringsearch/stringsearch.cpp
+++ b/panda/plugins/stringsearch/stringsearch.cpp
@@ -290,18 +290,28 @@ bool init_plugin(void *self) {
                 strlens[num_strings] = i;
             }
 
-            printf("stringsearch: added string of length %d to search set\n", strlens[num_strings]);
+             printf("stringsearch: added string %d of length %d to search set\n", num_strings,  strlens[num_strings]);
+            printf("[");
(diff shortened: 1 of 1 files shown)


AndrewFasano:
The CI failure there wasn't related to your changes, not sure what was going on:
```
2024-01-10T18:50:56.3659428Z 2980800K                                                       100% 0.00 =16s
timeout: failed to run command ‘panda-system-x86_64’: No such file or directory
2024-01-10T18:50:56.3659567Z /bin/sh: 1: nc: not found
2024-01-10T18:50:56.3659730Z Traceback (most recent call last):
2024-01-10T18:50:56.3659963Z   File "./generic_tests.py", line 138, in <module>
2024-01-10T18:50:56.3660087Z     runner(argv[1])
2024-01-10T18:50:56.3660284Z   File "./generic_tests.py", line 73, in runner
2024-01-10T18:50:56.3660511Z     raise RuntimeError("Failed to run CLI panda")
2024-01-10T18:50:56.3660681Z RuntimeError: Failed to run CLI panda
```

PR #1418, opened 2024-01-11
Conditionally load callstack_instr

MarkMankins:


MarkMankins:
I'd rather see the taint2 hypercalls be disabled by default - but that would be a breaking change.  If everyone can live with that I can update this PR.

MarkMankins:
Just to clarify - there are two options for this PR.

1) Accept is as-is.  If someone wants to take advantage of the optimization, they would need to add no_hypercalls=true as an argument to the taint2 plugin.  This keeps taint2 hypercalls working the way they always have.

2) I could rework this so that taint2 hypercalls are disabled by default.  This enables the optimization by default, but breaks things for anyone using taint2 hypercalls (which I suspect isn't many people but I can't be sure.)  I could add some warning messages that would be printed if a hypercall that appears to be a taint2 hypercall is encountered with instructions to add a command line parameter to enable hypercalls.

AndrewFasano:
Where are taint2 hypercalls used today? Are they in in-tree plugins? I'm not opposed to breaking changes so long as there's a way for users to turn them back on.

MarkMankins:
taint2_hypercalls.cpp contains the callbacks for the taint2 hypercalls.  guest_hypercall_callback is registered in init_plugin() in taint2.c.  I don't believe any in-tree plugins deliver taint2 hypercalls.  There is some test code that shows how they might be used (see the two panda_mark.h files - the windowsIPC flavor seems to be using the wrong constant values for LABEL_BUFFER & QUERY_BUFFER.)

AndrewFasano:
If it's just used in [LAVA](https://github.com/panda-re/lava) I think it would be fine to have the hypercalls disabled by default so long as there's an option to enable it when loading the plugin. We could fix it in LAVA, or just ignore it for now since I think they might be using an old version of PANDA anyway.

MarkMankins:
Sounds good - I'll update this PR.

PR #1423, opened 2024-01-19
osi_linux/kernelinfo_gdb: Ignore print_offset() errors

### LINKED ISSUE #1422: OSI Linux broken since `thread_group` removed
Linux removed `thread_group` in https://github.com/torvalds/linux/commit/8e1f385104ac044f1552686ad6e1cbc71cc05a30, and this breaks `osi_linux`, which expects that field to exist.

```
Reading symbols from /kernels/6.7/vmlinux.armel...
Python Exception <class 'gdb.error'>: There is no member named thread_group.
Error occurred in Python: There is no member named thread_group.
```

be32826:
Fixes part of #1422. If a struct field is deleted, kernelinfo_gdb will do the best it can instead of erroring and outputting nothing. Higher-level plugins that need to get a list of threads and need to be compatible with newer kernels can implement their own handling on top of this.

### DIFF
--- a/panda/plugins/osi_linux/utils/kernelinfo_gdb/extract_kernelinfo.py
+++ b/panda/plugins/osi_linux/utils/kernelinfo_gdb/extract_kernelinfo.py
@@ -10,8 +10,11 @@ def print_size(structv, cfgmemb, cfgname):
     print(f"{cfgname}.{cfgmemb} = {size}",file=file_out)
 
 def print_offset(structp, memb, cfgname):
-    offset = gdb.execute(f'printf "%d",(int)&(({structp}*)0)->{memb}',to_string=True)
-    print(f"{cfgname}.{memb}_offset = {offset}",file=file_out)
+    try:
+        offset = gdb.execute(f'printf "%d",(int)&(({structp}*)0)->{memb}',to_string=True)
+        print(f"{cfgname}.{memb}_offset = {offset}",file=file_out)
+    except gdb.error:
+        pass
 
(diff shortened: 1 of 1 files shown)


lacraig2:
I don't think this is the right approach.

The parser here provides an error because `osi_linux` requires that field as part of its profile. If you simply bypass the error you'll generate invalid profiles (i.e. `osi_linux` can't start up with them).

A better approach would be to see how this field contributes to `osi_linux` logic, ensure it handles the case that it is not present, and then mark this field optional (e.g. as with `start_time` [example](https://github.com/panda-re/panda/blob/8b04a8e2f2b4c522a2c66e89f1b1c43bd25bb8e0/panda/plugins/osi_linux/osi_linux.h#L246-L249)) and change the corresponding script to handle that.

AndrewFasano:
Perhaps there's some middle ground? It might be nice to have the config-generator support creating partial configs when fields are unavailable instead of having to patch it every time we want to make something optional? Perhaps a warning being generated along the lines of `Field X is missing - this may cause an error with OSI Linux`. Or maybe a flag like `--skip-missing` that turns these errors into a warning?

Then, to fix #1422 we'll need to update OSI Linux itself to work without this information.

be32826:
I have a patch for fixing `osi_linux` with newer kernels and I will make a PR soon, but this change is one part of the fix. If we aren't going to merge this, should I put everything in one PR, or is there a better way to have optional fields with kernelinfo_gdb?

PR #1424, opened 2024-01-25
syscalls2: Add sys_statx() for ARM

be32826:


lacraig2:
There was some really cool work @jamcleod did to build kernels and auto-generate these lists from their syscall tables.

This PR is fine, but a better one might be able to permanently fix these discrepancies.

be32826:
I originally tried running the scripts to automatically generate the syscall lists but it made too many changes that looked like breaking changes.

PR #1425, opened 2024-01-25
PyPANDA: Fix syscall args list size

be32826:
Before this change, accessing non-first syscall args in PyPANDA (such as in the below hook) gives incorrect bytes.

```python3
@panda.ppp("syscalls2", "on_all_sys_return2")
def all_sysret(cpu, pc, call, rp):
    print(list(rp.args[1]))
```

### DIFF
--- a/panda/python/core/create_panda_datatypes.py
+++ b/panda/python/core/create_panda_datatypes.py
@@ -37,7 +37,7 @@ INCLUDE_DIR_PYP = os.path.abspath(os.path.join(*[os.path.dirname(__file__), "pan
 INCLUDE_DIR_PAN = os.path.abspath(os.path.join(*[os.path.dirname(__file__), "..", "..", "include", "panda"])) # panda-git/panda/include/panda
 INCLUDE_DIR_CORE = os.path.abspath(os.path.join(*[os.path.dirname(__file__), "..", "..", "..", "include"]))   # panda-git/include
 
-GLOBAL_MAX_SYSCALL_ARG_SIZE = 64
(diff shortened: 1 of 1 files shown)


lacraig2:
This needs a fix, but this isn't quite the right one.

The code you're changing in `create_panda_datatypes.py` ([here](https://github.com/panda-re/panda/blame/ef40a12a2d6737af6f8d5cfc1202514a127373ed/panda/python/core/create_panda_datatypes.py#L291)) is reflecting code in [syscalls](https://github.com/panda-re/panda/blame/a5420fc766dba5d4f8debcc9abfc019431ce3c46/panda/plugins/syscalls2/generated/syscalls_ext_typedefs.h#L29).

The problem as I see it is that `create_panda_datatypes.py` defines the structure as:

```C
            uint8_t args[{GLOBAL_MAX_SYSCALL_ARG_SIZE}]
                 [{GLOBAL_MAX_SYSCALL_ARG_SIZE}]; /**< arguments */
```
when in fact the structure should be:
```C
    uint8_t args[GLOBAL_MAX_SYSCALL_ARGS]
                [GLOBAL_MAX_SYSCALL_ARG_SIZE]; /**< arguments */
``` 

The definition of `GLOBAL_MAX_SYSCALL_ARG_SIZE` of 8 makes sense given that it's defined in C as `sizeof(uint64_t)`. Realistically, this second dimension should change to `GLOBAL_MAX_SYSCALL_ARG_SIZE` to correctly reflect the C code.

be32826:
Fixed, thanks!

lacraig2 (approved):
Looks good.

PR #1427, opened 2024-01-26
Fix off-by-1 error in stringsearch.

MarkMankins:
Memory callbacks were not being enabled when using a stringsearch file that contained exactly one string. Enable callbacks when the first string is read from the file, not the second string.

### DIFF
--- a/panda/plugins/stringsearch/stringsearch.cpp
+++ b/panda/plugins/stringsearch/stringsearch.cpp
@@ -300,7 +300,7 @@ bool init_plugin(void *self) {
             }
             printf("]\n");
 
-            if (num_strings == 1) {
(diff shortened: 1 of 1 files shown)


lacraig2:
Nice catch. Will merge once the CI accepts it.

MarkMankins:
CI always errors out on my PRs - if there is something I can do on my side to fix that let me know.

AndrewFasano:
The `build_container` CI job (at least) has been broken since #1387 where it tries pushing a container to dockerhub on PRs which (fortunately) fails due to a permissions issue. Not sure what happened with the publish deb failure here, but it's clearly unrelated to your changes.

PR #1428, opened 2024-01-27
Fix split_taps.py to work in python3

Phasip:
split_taps.py does not work in python3, it mixes up byte arrays and strings and uses functionality not existing in python3.

This results in exceptions when trying to use it both for plain txt and for .gz files.

```
# ./split_taps.py write_tap_buffers.txt.gz out
  File "/usr/local/share/panda/scripts/split_taps.py", line 37, in <module>
    main(args.logfile, args.prefix, num_callers=args.callers)
  File "/usr/local/share/panda/scripts/split_taps.py", line 20, in main
    callers, pc, stack_kind, sid_first, sid_second, addr, n, val = line.strip().rsplit(" ", 7)
TypeError: a bytes-like object is required, not 'str'
```
And for non-gz:
```
./split_taps.py read_tap_buffers.txt out
  File "/usr/local/share/panda/scripts/split_taps.py", line 37, in <module>
    main(args.logfile, args.prefix, num_callers=args.callers)
  File "/usr/local/share/panda/scripts/split_taps.py", line 26, in main
    val = val.decode('hex')
AttributeError: 'str' object has no attribute 'decode'
```

### DIFF
--- a/panda/scripts/split_taps.py
+++ b/panda/scripts/split_taps.py
@@ -5,9 +5,9 @@ import time
 
 def main(logfile, prefix, num_callers=1):
     if logfile.endswith('.gz'):
-        f = gzip.GzipFile(logfile)
+        f = gzip.open(logfile, mode='rt')
     else:
-        f = open(logfile)
+        f = open(logfile,'r')
 
     filemap = {}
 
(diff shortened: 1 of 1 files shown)


lacraig2:
Looks good. Will merge after CI finishes.

PR #1435, opened 2024-02-04
Creating a pypanda wheel file

AndrewQuijano:
This patch will also generate the `pandare-0.1.2.0-py3-none-any.whl` file, which can successfully install pypanda via pip.


AndrewFasano:
Thanks for this! We could integrate it with the CI job [here](https://github.com/panda-re/panda/blob/dev/.github/workflows/publish_deb.yml) that publishes the deb and make that publish a wheel too if you're interested in giving it a try.

PR #1436, opened 2024-02-06
hooks: change block invalidator callback to PANDA_CB_BEFORE_BLOCK_EXEC_INVALIDATE_OPT

be32826:
Previously, the block invalidation function was called on `PANDA_CB_BEFORE_BLOCK_TRANSLATE`, which is only triggered when both TCG caches miss (i.e. when the block is already invalidated), so I think it is a no-op. It also seems to only invalidate the `tb_jmp_cache` and not `tb_ctx.htable`. I think both of these issues exist and that this PR solves them, but I'm not 100% sure.

### DIFF
--- a/panda/plugins/hooks/hooks.cpp
+++ b/panda/plugins/hooks/hooks.cpp
@@ -69,7 +69,7 @@ SUPPORT_CALLBACK_TYPE(after_block_exec)
 SUPPORT_CALLBACK_TYPE(start_block_exec)
 SUPPORT_CALLBACK_TYPE(end_block_exec)
 
-panda_cb before_block_translate_block_invalidator_callback;
+panda_cb block_invalidator_callback;
 
 // Handle to self
 void* self = NULL;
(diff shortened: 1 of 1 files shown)


AndrewFasano (approved):
I think these changes make sense, but I'll defer to @lacraig2 since he wrote it!

lacraig2:
> Previously, the block invalidation function was called on PANDA_CB_BEFORE_BLOCK_TRANSLATE, which is only triggered when both TCG caches miss (i.e. when the block is already invalidated), so I think it is a no-op.

The missing piece here is the context that we're enabling this context so that we can acquire the `tb_lock` so that we can efficiently handle dropping our block from the tb cache if we fail by other means. The event that generates the block translate is most likely unrelated to the block in question.

> It also seems to only invalidate the tb_jmp_cache and not tb_ctx.htable.

Those are hash tables utilized in distinct contexts: translation and runtime. I don't see an issue invalidating the `htable`, but I'm not convinced it's necessary.

lacraig2:
FWIW if we're looking for a better implementation of hooks have a look at this PR: https://github.com/panda-re/panda/pull/1187

PR #1439, opened 2024-02-14
New Dwarf2 plugin that supports x64 binary

HighW4y2H3ll:
The idea of this patch is to add x64 support for taint analysis of DWARF symbols. The `pri_dwarf` plugin is rewritten into `dwarf2` plugin to support the extension of new architectures. Now the `pri_dwarf` plugin is split into two parts: the first part is a Python script that reformats the outputs of `dwarfdump` command into a few JSON files, and the second part is the `dwarf2` plugin which takes in the pre-processed DWARF info.

In conclusion, this patch includes:  
- A Python script that dumps Dwarf info into JSON files. ([dwarfdump.py](https://github.com/panda-re/panda/compare/dev...HighW4y2H3ll:panda:newdwarf?expand=1#diff-82166e850c1f36ef2c46772d45f3e29de0fb2322df2a4ed0a3b267df8e807071))
- `dwarf2` PANDA plugin that replaces the old `pri_dwarf` plugin.
- Fixups in `pri_taint` PANDA plugin to use new `dwarf2` plugin.
- Other minor fixes in `file_taint` and `stringsearch` plugins

### DIFF
--- a/panda/plugins/dwarf2/dwarf2.cpp
+++ b/panda/plugins/dwarf2/dwarf2.cpp
@@ -203,7 +203,7 @@ struct VarInfo {
     VarInfo(target_ulong cu, target_ulong var_type, std::string var_name,
             target_ulong lowpc, target_ulong highpc, Json::Value ops, target_ulong dl, std::string fn) :
         cu(cu), var_type(var_type), var_name(var_name),
-        lowpc(lowpc), highpc(highpc), loc_ops(ops), dec_line(dl), fname(fn) {}
+        lowpc(lowpc), highpc(highpc), dec_line(dl), loc_ops(ops), fname(fn) {}
     VarInfo() {}
 };
 
@@ -502,8 +502,7 @@ LocType execute_stack_op(CPUState *cpu, target_ulong pc, Json::Value ops,
     //process_dwarf_locs(loc_list, loc_cnt);
     //printf("} = \n");
     target_ulong stack[64];	/* ??? Assume this is enough.  */
-    int stack_elt, loc_idx, i;
-    unsigned int next_offset;
+    int stack_elt, loc_idx;
     target_ulong result;
     bool inReg = false;
     //stack[0] = initial;
@@ -1318,7 +1317,6 @@ uint64_t elf_get_baseaddr(const char *fname, const char *basename, target_ulong 
 }
 
 int die_get_type_size (DwarfTypeInfo *ty){
-    int rc;
 
     // initialize tag to DW_TAG_pointer_type to enter the while loop
     while (ty->type == SugarType)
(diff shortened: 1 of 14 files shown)


panda/plugins/dwarf2/dwarf2.cpp
```diff
@@ -0,0 +1,2650 @@
+/* PANDABEGINCOMMENT
+ *
+ * Authors:
+ *  Tim Leek               tleek@ll.mit.edu
```
lacraig2: I'm aware it is the template, but this should only include folks that contributed to this.

panda/plugins/file_taint/file_taint.cpp
```diff
@@ -128,6 +128,7 @@ void read_return(uint64_t file_id, uint64_t bytes_read,
     ReadKey key;
     std::unique_ptr<OsiProc, decltype(free_osiproc)*> proc { get_current_process(first_cpu), free_osiproc };
     std::unique_ptr<OsiThread, decltype(free_osithread)*> thr { get_current_thread(first_cpu), free_osithread };
+    if (!thr)   return;
```
lacraig2, line 131: Will this work with the unique_ptr?

panda/plugins/stringsearch/stringsearch.cpp
```diff
@@ -24,6 +24,7 @@ PANDAENDCOMMENT */
 #include <sstream>
 #include <string>
 #include <iostream>
+#include <array>
```
lacraig2, line 27: Import added, but is it used anywhere?

lacraig2 (commented):
@HighW4y2H3ll you should rebase this on `dev` to get the CI to pass. (e.g. `git rebase -i origin/dev`)

This is due to no fault of your own, but it is necessary.

panda/plugins/file_taint/file_taint.cpp
HighW4y2H3ll: For the nullptr check, yup, it worked in my branch which was based on an older version of PANDA. unique_ptr overloads `operator bool()`, so I believe it should also work here.

panda/plugins/stringsearch/stringsearch.cpp
HighW4y2H3ll: yup, I got a compiler error at these 2 lines: https://github.com/panda-re/panda/blob/0afc7c38ff454a215a475d91e2ce625f99bdee68/panda/plugins/stringsearch/stringsearch.cpp#L60-L61

panda/plugins/dwarf2/dwarf2.cpp
HighW4y2H3ll: Updated!

lacraig2:
This PR isn't passing tests for reasons unrelated to it.

It does pass `test_installer` which indicates that it all compiles.

I will be merging this without tests if no one objects within a day or so.

PR #1445, opened 2024-02-17
CallstackInstr: switch to SBE/EBE callbacks

### LINKED ISSUE #1443: Callstack instr affected by panda_precise_pc & TB chaining
If TB chaining is enabled, callstack_instr finds > 2x calls with a linux-based system that runs a very simple bash-based init script. If panda_precise_pc is enabled, callstack_isntr finds 5% more calls.

Some of this is probably related to performance (maybe a slower guest runs more background timer-based things?). But I'm pretty sure at least some events are getting dropped without disabling tb chaining.

I don't think this is supposed to happen. If so, it should be documented.

AndrewFasano:
The old callstack instr logic would miss many calls if tb_chaining was enabled (which is is by default). Since this plugin did not disable tb_chaining, many calls would be missed if a user didn't disable chaining or load another plugin that disabled chaining.

This commit updates the plugin to use start_block_exec and end_block_exec which work even with tb_chaining enabled.

Thanks to @lacraig2 for suggesting this fix.

Hopefully fixes #1443.

One slight concern about this change is how the EBE callback works with interrupts - I know ABE is triggered when a block is interrupted and callbacks are infored of this through the `exitCode` arg which callstack_instr was checking. If EBE is just not called when this happens, this updated code will be correct. Otherwise, if it's still called on these weird "end" blocks, I'm not sure these changes will be correct.

### DIFF
--- a/panda/plugins/callstack_instr/callstack_instr.cpp
+++ b/panda/plugins/callstack_instr/callstack_instr.cpp
@@ -55,10 +55,8 @@ extern "C" {
 #include "panda/plog.h"
 #include "callstack_instr_int_fns.h"
 
-bool translate_callback(CPUState* cpu, target_ulong pc);
-int exec_callback(CPUState* cpu, target_ulong pc);
-void before_block_exec(CPUState* cpu, TranslationBlock *tb);
-void after_block_exec(CPUState* cpu, TranslationBlock *tb, uint8_t exitCode);
+void start_block_exec(CPUState* cpu, TranslationBlock *tb);
+void end_block_exec(CPUState* cpu, TranslationBlock *tb);
 void after_block_translate(CPUState* cpu, TranslationBlock *tb);
 
 bool init_plugin(void *);
@@ -322,7 +320,7 @@ void after_block_translate(CPUState *cpu, TranslationBlock *tb) {
     return;
 }
 
-void before_block_exec(CPUState *cpu, TranslationBlock *tb) {
+void start_block_exec(CPUState *cpu, TranslationBlock *tb) {
   // if the block a call returns to was interrupted before it completed, this
   // function will be called twice - only want to remove the return value from
   // the stack once
@@ -366,7 +364,7 @@ void before_block_exec(CPUState *cpu, TranslationBlock *tb) {
   }
 }
 
-void after_block_exec(CPUState* cpu, TranslationBlock *tb, uint8_t exitCode) {
+void end_block_exec(CPUState* cpu, TranslationBlock *tb) {
     target_ulong pc = 0x0;
     target_ulong cs_base = 0x0;
     uint32_t flags = 0x0;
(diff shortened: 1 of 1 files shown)


panda/plugins/callstack_instr/callstack_instr.cpp
```diff
@@ -611,9 +593,9 @@ bool init_plugin(void *self) {
 
     pcb.after_block_translate = after_block_translate;
     panda_register_callback(self, PANDA_CB_AFTER_BLOCK_TRANSLATE, pcb);
-    pcb.after_block_exec = after_block_exec;
+    pcb.end_block_exec = end_block_exec;
     panda_register_callback(self, PANDA_CB_AFTER_BLOCK_EXEC, pcb);
```
MarkMankins, line 597: This is still AFTER_BLOCK_EXEC?  Is that what you intended?  Same with BEFORE_BLOCK_EXEC below.  @LauraLMann spotted this.

AndrewFasano: I don't think so - looks like before_block_exec -> start_block_exec and after_block_exec -> start_block_exec. They're conceptually similar, but the implementation is different (most relevant here is that SBE and EBE trigger for each block even with TB chaining enabled).

AndrewFasano: I might be missing something super obvious though

MarkMankins: The second arg to panda_register_callback didn't change.  pcb is just a big union - the field names in that structure can be interchanged - changing just the field name in pcb is essentially a no-op.  Changing the field name implied that you wanted to change which callback was being invoked, but you are still registering the same callback.

AndrewFasano: Ah yes, I am missing something super obvious! Thanks, I'l fix

panda/plugins/callstack_instr/callstack_instr.cpp
```diff
@@ -377,35 +375,19 @@ void after_block_exec(CPUState* cpu, TranslationBlock *tb, uint8_t exitCode) {
...
-        }
+    if (tb_type == INSTR_CALL) {
+        stack_entry se = {tb->pc + tb->size, tb_type};
+        callstacks[curStackid].push_back(se);
+
+        // Also track the function that gets called
+        // This retrieves the pc in an architecture-neutral way
         cpu_get_tb_cpu_state(env, &pc, &cs_base, &flags);
-        // erase nicely does nothing if key DNE
-        stoppedInfo.erase(curStackid);
```
MarkMankins, line 407: I think I have a case where the removal of this logic is causing a regression.  Basically running the same recording can result in callstack_instr generating different output.  I think that's what this logic was preventing from happening.  Working on verifying this is indeed the case.

AndrewFasano: If callstack_instr is firing _more often_ after this change (including the correct callback types in #1447), I'd expect that - with the new callbacks it should be able to detect more calls.

If it's firing less, then that's an issue. I was thinking the end-block-exec callback would only trigger at the very end of a TB, and not if the block is jumped out of prematurely. But if the end-block-exec callbacks fires even if the block is interrupted, that would cause issues (particularly because that callback doesn't provide an "exitCode" arg to check for this behavior against). Hey @lacraig2 - do you know which of these behaviors we should expect from EBE?

AndrewFasano: (I'll move this discussion over to the new PR)

PR #1448, opened 2024-02-21
Add taint compute number output to ida_taint2 output

LauraLMann:
Include the minimum taint compute number encountered on each tainted instruction reported on by the ida_taint2 plugin.
Enhance the ida_taint2 IDAPython script and plugins to let the user pick the maximum compute number to report on.  This way, you don't have to rerun PANDA with a different max compute number to find out how soon after the tainted data makes it into your binary that it is processed by whatever part of the code you are interested in.  You can just select a new maximum to display in IDA.
The IDAPython script and plugins will still process old ida_taint2 output without complaint - you just won't get a chance to select a maximum taint compute number.
Note that the common parts of the IDAPython files were moved into a separate file used by the other 3.  There are directions in the README and in the code on where to put the common file so the other files can find the common code.


LauraLMann:
The error from the make_check check is "unable to find image <ugly path> locally".  Doesn't sound like something that changing a plugin and some IDAPython scripts could have caused.  But I am surely no expert in the checks so hopefully someone else knows better than me what is going on here.

AndrewFasano:
I'm not sure what's going on with that CI failure but it's certainly not your fault

PR #1452, opened 2024-02-23
Update publish to include Pypanda

AndrewQuijano:
Every time a Debian package release is created, the pypanda release should also be released as well. This would be important as when updating LAVA, I need both pypanda and the panda debian package to correctly install LAVA.

NOTE:
I have not been able to fully test this locally if the release correctly publishes the wheel.

Also, it is IMPORTANT that the whl file name not change!

Right now if you change the file name 'pandare-0.1.2.0-py3-none-any.whl' to anything else, pip will not correctly install pypanda.
This is reflected on $(echo "$(ls panda/debian/pandare-*.whl)" | sed 's/^.*pandare-//') to just get the '0.1.2.0-py3-none-any.whl' to append to the 'pandare-' package.

P. S.
Might also not be a bad idea to change 'publish_debian' to just 'publish'

### DIFF
--- a/.github/workflows/publish_deb.yml
+++ b/.github/workflows/publish_deb.yml
@@ -51,7 +51,7 @@ jobs:
             panda/debian/pandare-*.whl
           asset_name: |
             pandare_${{ matrix.ubuntu_version }}.deb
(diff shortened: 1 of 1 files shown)


AndrewFasano:
From what I understand, the only way to test changes to github CI actions is to merge them and see what happens. No worries if this takes a few tries to get right, it looks good to me - let's see what happens.

PR #1467, opened 2024-02-25
Attempt 2 on publishing wheel

AndrewQuijano:
I did a bit more investigating on publishing wheel files and found this link:
https://github.com/pypa/gh-action-pypi-publish/discussions/15

Based on the information here, I updated the YAML to repeat what they did for releasing py panda
https://github.com/ansible/pylibssh/blob/17e8683/.github/workflows/build-test-n-publish.yml#L353-L357
https://github.com/ansible/pylibssh/releases

### DIFF
--- a/.github/workflows/publish_docker.yml
+++ b/.github/workflows/publish_docker.yml
@@ -154,6 +154,28 @@ jobs:
              git commit -m "Documentation update for PANDA commit ${{ github.sha  }} branch dev" &&
              git push || true
 
+  publish-to-pypi:
+    name: >-
+      Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    needs:
+    - build_release_assets
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/pandare  # Replace <package-name> with your PyPI project name
+    permissions:
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
+
+    steps:
+    - name: Publish distribution 📦 to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: pandare
+        password: ${{ secrets.PYPI_API_TOKEN }}
+        packages-dir: panda/debian/
+        verbose: true
+        
   build_stable:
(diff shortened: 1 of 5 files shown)


lacraig2:
I tried getting this workflow working several times and on each of the runs the whl file was missing.

https://github.com/panda-re/panda/actions/runs/8035190848/job/21947562763#step:5:11

This version looks like we've changed to release assets and now uploading the the whole folder as release assets.

AndrewQuijano:
Hm, so technically the file is specifically called 'pandare-0.1.2.0-py3-none-any.whl' after I finish running setup. I do try to be flexible with the naming in case any changes occur to setup.py.

I'm a bit surprised as when I run this locally, I do get the wheel file

lacraig2:
I also get the whl file locally.

I'm not quite sure what the issue is.

AndrewQuijano:
Actually, I just noticed there is a whole pypackage. Could we not use GitHub actions to automate updating the pip package?

https://pypi.org/project/pandare/#description

AndrewQuijano:
https://github.com/marketplace/actions/pypi-publish

lacraig2:
@AndrewQuijano now that the whl is small I think that would be fine.

AndrewQuijano:
After a quick chat with Brendan, apparently * is not allowed, so I used this comment for a work around. I will also work on pypi upload too on this PR.

https://github.com/actions/upload-release-asset/issues/47#issuecomment-668543873

AndrewQuijano:
I made a first attempt at writing the job to publish to pypi, I will likely need help from MITLL to get the secret and all to get this working too.

Also, I should note, currently for python, protobuf 3.20.0 is the latest version that works with the installed version of protobuf. Not sure if you suggest I update the pypanda requirements for this?

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

AndrewQuijano:
For transparency, I am also working on updating how python package dependencies are managed. I will run the test debian package set up later tonight to confirm this works. 

We should avoid having 3 separate places where we install a python package.

lacraig2:
> We should avoid having 3 separate places where we install a python package.

❤️

AndrewQuijano:
packages_made
pip_worked

lacraig2:
I think this isn't passing tests for unrelated reasons.

@AndrewQuijano I will merge this if you think it's ready.

PR #1468, opened 2024-02-28
Make errors on return type much harder to achieve in PyPANDA

lacraig2:
This is a draft for code that would need to be propagated to other parts of PyPANDA to make it much harder for callbacks to error on the incorrect type returned from a callback.

In this case we move the type conversion that cffi would do upon return of our outer callback wrapper to the forefront and check it for errors. If there is an error we make an explicit offering of a type cast from zero.

This should be handled in the other _run_and_catch functions. It should also have a better policy on errors. Printing every time is annoying, but some feedback is useful.


AndrewFasano (approved):
Looks reasonable to me, assuming you've run it and it worked.

PR #1471, opened 2024-03-11
Fix start_block_exec nondeterminism

lacraig2:
#1447 determined that switching callstack_instr to start_block_exec caused nondeterministic behavior.

This PR attempts to resolve that issue.

### DIFF
--- a/panda/src/cb-support.c
+++ b/panda/src/cb-support.c
@@ -83,6 +83,9 @@ MAKE_CALLBACK(void, END_BLOCK_EXEC, end_block_exec,
 
 // Non-macroized version for SBE - if panda_please_retranslate is set, we'll break
 void PCB(start_block_exec)(CPUState *cpu, TranslationBlock *tb) {
+    if (unlikely(panda_exit_loop)){
+        return;
+    }
     panda_cb_list *plist;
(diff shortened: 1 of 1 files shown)


AndrewFasano:
Prior to this, if `panda_exit_loop` was true was the sequence of events something like:
1) An interrupt/exception/etc occurs which should redirect control flow from block B (about to be executed) to block C (some other place)
2) SBE callback triggers for block B
3) SBE callback triggers for block C
4) Code in block C runs (probably multiple blocks with SBE/EBE callbacks firing as exepected)
5) Control flow returns to block B
6) SBE callback triggers again for block B
7) Block B actually runs

And the issue was that this double-SBE callback for block B should have only happened once (and was non-deterministic in when it happened twice)

AndrewFasano:
Even if this doesn't fix the issue in #1447, is this still worth merging? Or do you want to try tracking it down further and seeing if you find the fix?

lacraig2:
I'm not sure it really adds anything in the scenario that it doesn't fix #1447.

Or rather, I expect this _and_ something else will completely resolve #1447, but I haven't had a chance to figure out what that is.

PR #1481, opened 2024-03-22
Update pypanda README based on poking around with pypanda wheel

AndrewQuijano:
Long story short, I tested running plog reader from a Panda Log Zhenghao made for toy. I do want to report this potential issue just in case to MITLL and panda users. At least for LAVA, I may potentially set this environment variable to minimize chances of crashing now that panda submodule will be removed from LAVA

https://github.com/panda-re/panda/blob/dev/panda/python/core/pandare/plog_reader.py


.github/workflows/publish_docker.yml
```diff
@@ -25,14 +25,6 @@ jobs:
         with:
           release_branch: dev
           use_api: true
-      - name: Create release
```
lacraig2, line 28: Why are we getting rid of create_release?

panda/debian/setup.sh
```diff
@@ -53,6 +53,7 @@ docker build -t packager .
 
 # Copy deb file out of container to host
 docker run --rm -v $(pwd):/out packager bash -c "cp /pandare.deb /out"
+mv pandare.deb pandare_${version}.deb
```
lacraig2, line 56: Can we not do this with CI? I think the output of this could quite reasonably be pandare.deb for every user that wants to build a deb.

.github/workflows/publish_docker.yml
lacraig2: I see what you are doing here.

panda/debian/setup.sh
lacraig2: Nevermind. This is fine.

PR #1483, opened 2024-03-27
Replace gnutls-dev in ubuntu_20.04_build.txt to use exact package

saccharide:
`gnutls-dev` is actually a virtual package, and could resolve to multiple package such as `libgnutls-dev` and `libgnutls28-dev`, which could lead to installation abortion and user would need to select the exact `gnutls` package in order to proceed.

This PR seeks to replace this virtual package to use the exact package `libgnutls28-dev`
image

### DIFF
--- a/panda/dependencies/ubuntu_18.04_build.txt
+++ b/panda/dependencies/ubuntu_18.04_build.txt
@@ -36,7 +36,7 @@ nasm
 
 # Qemu build deps
 device-tree-compiler
-gnutls-dev
+libgnutls28-dev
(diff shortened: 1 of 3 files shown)


AndrewFasano:
Happy to merge this, but would you mind also updating the other ubuntu_*_build.txt files so we can keep things consistent?

saccharide:
> Happy to merge this, but would you mind also updating the other ubuntu_*_build.txt files so we can keep things consistent?

sure thing and thanks!

PR #1490, opened 2024-04-03
Add back .git folder to dockerfile build stage

lacraig2:
Required for some context and version information.


AndrewFasano:
I'm not opposed to merging this, but would it be possible to avoid keeping the full git history around and still preserve the version number. At least on my machine the .git directory is nearly 900MB which seems like a lot to just get a version number.

lacraig2:
@AndrewFasano this change does not actually impact the size of the final panda container.

It may not even impact the size of intermediate stages given how docker stages function.

lacraig2:
Just to clarify: we're not distributing the .git folder in the final image. It's just there as we build.

Issue #1380, opened 2023-11-03
clone error


cctv130:
Cloning into 'panda'...
remote: Enumerating objects: 742156, done.
remote: Counting objects: 100% (742135/742135), done.
remote: Compressing objects: 100% (134176/134176), done.
remote: Total 742156 (delta 605732), reused 741245 (delta 605290), pack-reused 21Receiving objects: 100% (742156/742156)Receiving objects: 100% (742156/742156), 391.09 MiB | 10.95 MiB/s, done.

Resolving deltas: 100% (605747/605747), done.
error: invalid path 'panda/dependencies/ubuntu:18.04_base.txt'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'

AndrewFasano:
Sounds like something wrong with your system or your connection to github, not an issue with our code. Perhaps you can delete your panda directory and re-clone or try the command git suggested in your error.

MarkMankins:
I've seen this if you try to clone the repo on windows.

The panda repo contains files that have characters that are illegal in windows file names.

AndrewFasano:
🤯 wow that's unexpected - perhaps this is a bug we should be tracking then!

We should be able to rename the dependency list filenames to drop the `:` and make Windows happy

PR #1505, opened 2024-06-24
Remove colons from filename.

MarkMankins:
This allows the panda repo to be cloned on windows.


AndrewFasano:
Oh no, it's the same as [this issue](https://github.com/panda-re/panda/issues/1380) just in a different place. Sorry about that - thanks for the fix.

The CI failure here is unrelated to your changes, but caused by the fact that outside contributors can't push container to dockerhub and we use dockerhub as some temporary storage during CI testing - @AndrewQuijano is working on some alternative designs to fix that.

MarkMankins:
Thanks for the quick merge!

PR #1507, opened 2024-07-02
Updates to GHA, Install Script and documentation

AndrewQuijano:
I squashed the commits from this old PR, not sure why rebasing didn't work.
https://github.com/panda-re/panda/pull/1504

Here is the summary:

1- Update Action version to the latest version

2- Update to use Org secret (I will require pandare/panda_test repo to be made to be the cache for containers used for testing)

3- Slightly update README about debian packages/pypanda distribution

4- Slightly optimize setup so we only build one panda image, not two

5- Deleting .travis.yaml and publish_debian.yaml

6- install_ubuntu.sh should be downloading the package not building libosi, install capstone v5 and not use sudo to install pypanda

Most important
7- Pushing containers used in parallel tests to pandare/panda_test repository as the cache. This is because external users can't push to GHCR, but we should be able to push to DockerHub

This should allow testing to fully run even for those not part of the PANDA org


AndrewFasano:
Great. I updated the `ALL_PANDARE_DOCKERHUB` and made the `panda_test` project on dockerhub. Let's see if this works!

panda/scripts/install_ubuntu.sh
```diff
@@ -25,13 +25,14 @@ lsb_release --help &>/dev/null || $SUDO apt-get update -qq && $SUDO apt-get -qq
 git --help &>/dev/null || $SUDO apt-get -qq update && $SUDO apt-get -qq install -y --no-install-recommends git
 
 # some globals
+LIBOSI_VERSION="0.1.3"
```
AndrewFasano: Do you want this to be 0.1.7? That's the version with your latest PR merged

PR #1510, opened 2024-07-09
Added new functionality to filereadmon plugin

Anteater2019:
- Added 64 bit support for windows
- Added Linux support for 32 and 64 bit
- Added pread system calls for Linux OS


AndrewFasano:
Thanks for the PR!

Ignore that one CI failure, it's an issue with permissions and not your fault. I'm hoping the other CI jobs will still run though, if they pass I think we can merge this

PR #1514, opened 2024-07-30
Update to PR #1509

AndrewQuijano:
This PR contains the squashed commits from #1509, it seems caching test containers on DockerHub works, figure once this is in, let's see if Workflow dispatch does allow for running parallel tests on external PRs.

I do want to get this in as Zhenghao (one of Brendan's PhD students) just graduated, and does want to refactor IRQ Debloat to use the Panda Debian package

https://github.com/messlabnyu/irqdebloat

PR #1515, opened 2024-07-30
Update CI versions & cleanup setup scripts

AndrewFasano:
Supersedes @AndrewQuijano's #1514. I want to test if CI behaves differently when issued from a member of the panda-re org.

I dropped the CI changes around org-wide secrets and cache images on dockerhub instead of github infrastructure because it didn't seem to make a difference when external folks open PRs. So we may as well stick with what has been working.


AndrewQuijano:
Nice! It worked! I guess once the workflow_dispatch is there, we can see if clicking a different button might allow running parallel tests for external works.

https://stackoverflow.com/questions/69499645/how-to-securely-allow-github-actions-to-check-pr-and-post-results-in-comment

PR #1516, opened 2024-08-01
FIX: make hypercall return none always

lacraig2:
Hypercaller always returns void so this change ensures that callback always returns void.


AndrewQuijano:
@lacraig2 

Quick thing, could you also please update 'workflow_run' to 'workflow_dispatch'
https://github.com/panda-re/panda/blob/dev/.github/workflows/publish_docker.yml#L4

Just to fix the publish_debian too while you get your fix in too

PR #1518, opened 2024-08-07
Fix the time-trevel feature (panda/docs/time-travel.md)

artmetla:
By cleaning panda/scripts/gdbinit from redundant commands fix time-traveling feature and ensure that it works according to the docs: panda/docs/time-travel.md.


artmetla:
I checked the error and it's 403. Looks like it has nothing to do with the change proposed.

lacraig2:
There is nothing in this PR that should impact any test.

Will merge.

PR #1524, opened 2024-08-19
Correct documentation.

MarkMankins:
panda_virtual_memory_* and panda_physical_memory_* return non-zero values on error.  These can be positive - in fact in the case of panda_physical_memory_* there are only two possible values currently, 0 and 1.


lacraig2:
Good catch @MarkMankins.

This doesn't need the tests.

PR #1528, opened 2024-09-02
add check for nulled list

### LINKED ISSUE #1527: SEGFAULT in some plugin deletion contexts
When deleting a callback under very specific conditions we can get ugly errors.

In particular, when deleting a callback that is the last of its kind from within that callback we see a segfault in the following code:

https://github.com/panda-re/panda/blob/d0618d33c0a577f7e7281b29d3f04b31517dfc34/panda/include/panda/callbacks/cb-macros.h#L126-L138

A minimal example of this behavior for reproducing:

```python
#!/usr/bin/env python3
from pandare import Panda

panda = Panda(generic="i386")

@panda.queue_blocking
def run_cmd():
    panda.revert_sync("root")
    print(panda.run_serial_cmd("uname -a"))
    print(panda.run_serial_cmd("uname -a"))
    print(panda.run_serial_cmd("uname -a"))
    panda.end_analysis()

@panda.cb_main_loop_wait
def asidchange():
    panda.delete_callbacks()

panda.run()
```

lacraig2:
This PR fixes #1527 by adding checks to see if the list is empty on each iteration.

FIXES: #1527

### DIFF
--- a/panda/include/panda/callbacks/cb-helper-impl.h
+++ b/panda/include/panda/callbacks/cb-helper-impl.h
@@ -18,7 +18,9 @@ PANDAENDCOMMENT */
 void HELPER(panda_insn_exec)(target_ulong pc) {
     // PANDA instrumentation: before basic block
     panda_cb_list *plist;
-    for(plist = panda_cbs[PANDA_CB_INSN_EXEC]; plist != NULL; plist = panda_cb_list_next(plist)) {
+    for(plist = panda_cbs[PANDA_CB_INSN_EXEC]; 
+           (plist != NULL && panda_cbs[PANDA_CB_INSN_EXEC] != NULL) ;
+        plist = panda_cb_list_next(plist)) {
         if (plist->enabled) {
             plist->entry.insn_exec(plist->context, first_cpu, pc);
         }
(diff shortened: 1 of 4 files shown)


panda/include/panda/callbacks/cb-helper-impl.h
```diff
@@ -18,7 +18,9 @@ PANDAENDCOMMENT */
 void HELPER(panda_insn_exec)(target_ulong pc) {
     // PANDA instrumentation: before basic block
     panda_cb_list *plist;
-    for(plist = panda_cbs[PANDA_CB_INSN_EXEC]; plist != NULL; plist = panda_cb_list_next(plist)) {
+    for(plist = panda_cbs[PANDA_CB_INSN_EXEC]; 
+           (plist != NULL && panda_cbs[PANDA_CB_INSN_EXEC] != NULL) ;
```
AndrewFasano, line 22: From my reading of this, it seems like some of these changes might not be meaningful (though I'm sure there's something in this PR that's fixing the underlying bug)

Since `plist` is set to `panda_cbs[PANDA_CB_INSN_EXEC]`  I don't see how `plist` could ever be NULL while the `panda_cbs[PANDA_CB_INSN_EXEC]` object is non-NULL - since they're the same.

Though, if there are multiple threads causing race conditions, then that could mean timing is changed by the redundant check (if it's not optimized out) and perhaps improve things - though at that point I'd think the solution would be to do better thread safety instead of duplicating checks.

AndrewFasano: Hang on - staring at this closer and seeing how they could differ - after the initial iteration of the loop the `panda_cb_next` updates `plist` and perhaps after the first iteration the entire list could become NULL?

Seems like an easier fix might be to ensure the various callback lists only ever become empty, not null?

PR #1532, opened 2024-09-10
small pypanda  interface fixes

zestrada:
A few minor usability things

### DIFF
--- a/panda/python/core/pandare/panda.py
+++ b/panda/python/core/pandare/panda.py
@@ -1707,7 +1707,7 @@ class Panda():
         family_num = self.libpanda.panda_os_familyno
         family_name = self.ffi.string(self.ffi.cast("PandaOsFamily", family_num))
         return family_name
-    
+
     def get_file_name(self, cpu, fd):
         '''
         Get the name of a file from a file descriptor.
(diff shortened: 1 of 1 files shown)


AndrewFasano:
I know it's a draft but all these changes look good to me - will hold off on merging until you mark it as ready for review.

panda/python/core/pandare/panda.py
```diff
@@ -1725,7 +1725,7 @@ def get_file_name(self, cpu, fd):
             return None
         if fname_ptr == self.ffi.NULL:
             return None
-        return self.ffi.string(fname_ptr)
+        return self.ffi.string(fname_ptr).decode('utf8', 'ignore')
```
lacraig2, line 1728: As much as I like this change it will break all the code that uses this currently.

Also: in Linux I'd expect Latin-1.

AndrewFasano: +1 for Latin-1

PyPanda is versioned, are we not willing to make updates that break backwards compatibility with new versions? Or are you just suggesting we should update the in-tree uses of `get_file_name` to account for this change?

zestrada: We use utf-8 everywhere else in these interfaces, so maybe we should change those too if we're wanting to use Latin-1?

PR #1535, opened 2024-10-02
Reduce memory usage when LLVM is enabled.

MarkMankins:
Function pass manager objects were being created, but never destroyed.

When a function pass manager object is destroyed, it destroys all pass objects that have been registered.  Previously pass objects were being reused, which is why the pass managers were never being freed, but this results in a large memory leak.  This commit shuffles things around so the pass managers can be freed.

a891


AndrewFasano:
Thanks Mark!

PR #1536, opened 2024-10-02
DO NOT MERGE: Ubuntu 24 - LLVM 14 preview

MarkMankins:
I managed to get PANDA built under Ubuntu 24, including an update to LLVM 14.
This isn't production ready quite yet, but it's a start.
I don't have the bandwidth to get this all the way to Ubuntu 24, but hoping with this initial work others can chip away at the problems that are left.
There are a lot of little issues that are nicely segregated, that would take me a while to research and fix all of them.  I suspect many of these have an easy solution, so I'm looking for help to move this along.

I'm not convinced yet the update to LLVM 14 is 100% correct - but I have a lots of test cases and most agree with the output from LLVM 11.  I have a few tests that are tightly coupled to the IR that gets generated, these will require some time to investigate whether or not the output I'm seeing in LLVM 14 is correct.  Luckily I know where the author of these tests works (@LauraLMann) so I know where to look for help.

I didn't even try to fire up PyPANDA - so I'm not sure what state it is in.

Known issues:
- When LLVM is enabled many warnings about unsupported intrinsics appear on the console.
- The libosi build takes a long time.
- LLVM 14 uses significantly more memory mapped regions than LLVM 11.
-- "wc -l /proc/\<panda pid\>/maps" to see what I mean.
-- Long running recordings with LLVM enabled may fail with out of memory errors.
-- You can increase the number of permitted memory mapped regions for a process by writing to /proc/sys/vm/max_map_count.  Doing this resolves the issue.
- configure must be run with --disable-werror.
- When installing required Ubuntu packages:
-- The Ubuntu 20 qemu package was renamed to qemu-system in Ubuntu 24.
-- I also installed llvm-14, llvm-14-doc, clang-14 & clang-tools-14.  (Not sure the clang packages are necessary - things might be building with clang-18 which was already installed in my Ubuntu 24 environment.)
- cmake is complaining loudly - but no fatal issues.
- There are warnings when compiling PANDA.
-- Many of these are about invalid escape sequences - at first glance it looks like these might be coming from python.
- Network plugin won't build.
-- Wireshark api changed, again.

### DIFF
--- a/configure
+++ b/configure
@@ -269,8 +269,8 @@ stack_protector=""
 git_submodules=""
 protobuf="yes"
 llvm="no"
-llvmdir="/usr/lib/llvm-11" # Default to try if unspecified
-llvmver="11"
+llvmdir="/usr/lib/llvm-14" # Default to try if unspecified
+llvmver="14"
 
 # Don't accept a target_list environment variable.
 unset target_list
--- a/panda/include/panda/tcg-llvm.h
+++ b/panda/include/panda/tcg-llvm.h
@@ -280,6 +280,9 @@ class TCGLLVMTranslator {
 
     void jitPendingModule();
 
+    llvm::LoadInst *createLoad(llvm::Value *Ptr, const llvm::Twine &Name = "",
+        bool isVolatile = false);
+
     public:
     TCGLLVMTranslator();
     ~TCGLLVMTranslator();
--- a/panda/llvm/tcg-llvm.cpp
+++ b/panda/llvm/tcg-llvm.cpp
@@ -94,7 +94,7 @@ using namespace llvm;
  * generated code (aka the section size) for use in the associated
  * TranslationBlock.
  */
-void getLLVMAssemblySize(orc::VModuleKey,
+void getLLVMAssemblySize(orc::MaterializationResponsibility&,
         const object::ObjectFile &obj,
         const RuntimeDyld::LoadedObjectInfo &objInfo) {
     if (!need_section_size) {
@@ -331,6 +331,12 @@ unsigned TCGLLVMTranslator::getValueBits(int idx)
     return 0;
 }
 
+LoadInst *TCGLLVMTranslator::createLoad(Value *Ptr, const Twine &Name,
+        bool isVolatile) {
+    return m_builder.CreateLoad(Ptr->getType()->getPointerElementType(),
+        Ptr, isVolatile, Name);
+}
+
 Value *TCGLLVMTranslator::getValue(int idx)
 {
     assert(idx >= 0 && idx < TCG_MAX_TEMPS);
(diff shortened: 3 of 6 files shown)


AndrewQuijano:
Hello @MarkMankins, I wanted to ask how are things going with the LLVM-14 upgrade? I can try to help on my end trying to figure out how to enable testing from external PRs

MarkMankins:
Unfortunately I've had to set aside my panda work for the short term so I haven't been able to make any progress here.  Not sure when I'll be able to get back to it.

AndrewQuijano:
I see, @LauraLMann, would you have any time to help with the upgrade? 

With regard to libosi, I made a Ubuntu 24 Debian image, pending PR, hopefully this saves 40 minutes, really surprised it is this slow to compile...
https://github.com/panda-re/libosi/pull/17

Also about wireshark, seems like someone else got the fix
https://github.com/panda-re/panda/pull/1569

LauraLMann:
> I see, @LauraLMann, would you have any time to help with the upgrade?
> 
> With regard to libosi, I made a Ubuntu 24 Debian image, pending PR, hopefully this saves 40 minutes, really surprised it is this slow to compile... [panda-re/libosi#17](https://github.com/panda-re/libosi/pull/17)
> 
> Also about wireshark, seems like someone else got the fix #1569

Nope, I'm in the same boat as my coworker Mark Mankins.  Hopefully this summer we can return.

LauraLMann:
So the problem with "When LLVM is enabled many warnings about unsupported intrinsics appear on the console." is that the taint2 plugin does not have propagation rules for certain LLVM intrinsic functions, and so it spits out this warning to the console when it encounters one to say "we aren't even going to try to propagate taint for that function".
So far I've run into 11 such functions, 3 of which it appears to me existed back in LLVM 11 days - apparently the translation has changed since then so now they show up in one of our tests where they didn't show up before.  The other intrinsics it's complaining about are new.  Some of them (like llvm.abs.*) it should be easy to write propagation rules for.  The others will require more effort.
Regardless, this won't cause PANDA to roll over and die, or even do crazy things.  It's just another taint deficiency.  Hopefully I will have time to take care of these after more important issues are dealt with.

LauraLMann:
The reason why "configure must be run with --disable-werror" is that two functions the configure script is checking out, clock_adjtime and clock_gettime, have different definitions now.  Each has an annotation stating that the second argument must not be null, so the compiler will issue a warning if it is.  The little test programs the configure script uses to see if these functions are available was using 0 for the second argument.  That's easy enough to adjust, and I'll include it in an update to this pull request.

LauraLMann:
The libosi build goes a lot faster in the Ubuntu 24 container I am using if I tell it to build with g/g++-9 instead of the newer compiler that is the default in the container.  It only takes about 2 or 3 minutes as opposed to the 40-45 it takes with the default compiler.

LauraLMann:
Per coordination with @MarkMankins , this pull request is being closed in favor of https://github.com/panda-re/panda/pull/1584.

PR #1537, opened 2024-10-10
Fix AARCH64 syscalls2 and OSI

lacraig2:
This PR fixes assumptions in AARCH64 OSI and syscalls2.

### DIFF
--- a/panda/plugins/osi_linux/osi_linux.cpp
+++ b/panda/plugins/osi_linux/osi_linux.cpp
@@ -349,6 +349,19 @@ inline bool can_read_current(CPUState *cpu) {
 // won't check again until the first syscall.
 bool r28_set = false;
 inline void check_cache_r28(CPUState *cpu);
+#elif TARGET_AARCH64
+
+target_ulong spel0 = 0;
+bool aarch64_initialized = false;
+
+void aarch64_sbe(CPUState *cpu, TranslationBlock *tb);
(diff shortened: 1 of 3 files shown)


panda/plugins/osi_linux/osi_linux.cpp
```diff
@@ -349,6 +349,19 @@ inline bool can_read_current(CPUState *cpu) {
...
+#elif TARGET_AARCH64
+
+target_ulong spel0 = 0;
+bool aarch64_initialized = false;
+
+void aarch64_sbe(CPUState *cpu, TranslationBlock *tb);
+void aarch64_sbe(CPUState *cpu, TranslationBlock *tb) {
+    if (unlikely(panda_in_kernel_code_linux(cpu) && ((CPUARMState*) cpu->env_ptr)->sp_el[0] != 0)){
+        aarch64_initialized = true;
+        spel0 = ((CPUARMState*) cpu->env_ptr)->sp_el[0];
```
AndrewFasano, line 361: Would you be happier to have this logic in https://github.com/panda-re/panda/tree/dev/panda/plugins/hw_proc_id? We're doing something similar in there for mips - that way OSI itself isn't doing block-level analysis, just consuming the results of another analyses.

lacraig2: I thought about that, but the frustrating part is that it's only enabled for kernels later than ~4.6. Additionally, it's not actually producing a hwid value like r28 is used in MIPS.

Issue #1525, opened 2024-08-20
syscalls2 x64 and ARM64 do not support 32-bit syscalls


lacraig2:
This seems to have been a deliberate choice. I'd like to know if anyone has strong feelings about it.

If not, I'd propose we add support for it.

If people do care about it I think it'd be fine to add an argument to syscalls2 for one functionality or the other.

@LauraLMann it looks like git blame has you adding this so I want to make sure to get your input.

https://github.com/panda-re/panda/blob/1b4c51b32284bf894ad2fe941f166424edb60df0/panda/plugins/syscalls2/syscalls2.cpp#L1044-L1059

LauraLMann:
Yes @lacraig2 , you can blame me for that.  We didn't need it for what we were doing at the time, and I couldn't think of an easy way to implement it, so I put in those warnings instead.
As long as adding it doesn't break our regression tests, I can't imagine why we would have a problem with adding 32-bit system calls in syscalls2 x64.  What do you think, @MarkMankins ?

LauraLMann:
I have a vague recollection of thinking we'd have to support flipping back and forth between 64-bit syscalls and 32-bit syscalls in the same recording, but I may misunderstand how they really work.  (I also recall reading a lot of manuals to try to figure out what was going on.)
It sounds like you are instead thinking @lacraig2 that the use cases would only involve one or the other, not both at the same time, if you can set at plugin startup time which bitness of system calls you are interested in for a 64-bit guest. Do I understand you right?

MarkMankins:
This would be a nice feature to have.

AndrewQuijano:
Out of curiosity, was this the cause of LAVA crashing when injecting bugs in 32-bit binaries? Least on my end, I think the preference would be that LAVA can inject either 32-bit of 64-bit bugs. We will get LAVA to point to 64-bit bugs shortly, but I am already doing a lot with this upgrade.

lacraig2:
> It sounds like you are instead thinking @lacraig2 that the use cases would only involve one or the other, not both at the same time, if you can set at plugin startup time which bitness of system calls you are interested in for a 64-bit guest. Do I understand you right?

I'm interested in handling any syscall that is valid for the target architecture in question in any order they are presented. On a 64-bit guest that means handling the backwards compatible 32-bit versions.

The prototypical case is actually MIPS64 which handles 3 separate ABIs by switching on the syscall number.
[...]

PR #1541, opened 2024-10-21
syscalls2: add multi-abi support

lacraig2:
This PR addresses #1525 by adding the capacity for AARCH64 and X86_64 to call into the code paths for their 32-bit compatible code to parse and handle ABIs other than the original default ABI.

It does so by associating a specific profile that accompanies every syscall. That profile can change either at translate or runtime.

For example, at translate time we can know if an instruction on the X86_64 ISA is actually a backwards compatible `int 0x80` instruction and transfer from a PROFILE_LINUX_X64 to a PROFILE_LINUX_X86. In this case we actually call the x86 codepath instead of the x64 codepath.

Worth knowing:
- In the case that multiple ABIs define the same syscall (common) we take the function definition from the 64-bit version.
- I have not attempted to adjust the logic for changing profiles for anything other than Linux.


MarkMankins (approved):

panda/plugins/syscalls2/syscalls2.cpp
```diff
@@ -1324,7 +1378,8 @@ bool init_plugin(void *self) {
 
     // load system call info
     if (panda_parse_bool_opt(plugin_args, "load-info", "Load systemcall information for the selected os.")) {
-        if (load_syscall_info() < 0) return false;
+        printf("load_info\n");
```
AndrewFasano, line 1381: Can this be removed or at least wrapped in a call to the standard loggers?

PR #1557, opened 2024-12-06
Improve nix flake

lluchs:
This fixes the nix flake (it didn't work with current nixpkgs since the old libdwarf was removed and some Python trouble) and implements some improvements:

- The plugin configuration can be overwritten by flake consumers and the flake can reduce the dependencies accordingly.
- PyPANDA is built separately (mostly to avoid rebuilding all of PANDA with every change while figuring out Python packaging)
- I also updated Wireshark to the newest version supported by the network plugin

cc @be32826 who built the flake originally

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,47 +2,49 @@
   description = "PANDA: Platform for Architecture-Neutral Dynamic Analysis";
 
   inputs = {
+    # nixpkgs with a Wireshark version that works with the network plugin.
+    nixpkgs-wireshark = {
+      url = "github:NixOS/nixpkgs?rev=e93823409f9e6b8e878edf060b430a14353a28f9";
+    };
     libosi-src = {
       url = "github:panda-re/libosi";
       flake = false;
     };
+    nix-filter.url = "github:numtide/nix-filter";
   };
 
-  outputs = { self, nixpkgs, libosi-src }: {
-
-    packages.x86_64-linux.default = let
-
-      pkgs = import nixpkgs {
-        system = "x86_64-linux";
-        config.permittedInsecurePackages = [ "libdwarf-20210528" ];
-      };
-
+  outputs = { self, nixpkgs, nixpkgs-wireshark, libosi-src, nix-filter }:
+    let
+      system = "x86_64-linux";
+      pkgs = import nixpkgs { inherit system; };
+      lib = pkgs.lib;
       pyPkgs = pkgs.python3Packages;
+      filter = nix-filter.lib;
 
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -2,63 +2,83 @@
...
+    nixpkgs-wireshark = {
+      url = "github:NixOS/nixpkgs?rev=e93823409f9e6b8e878edf060b430a14353a28f9";
+    };
     libosi-src = {
       url = "github:panda-re/libosi";
       flake = false;
     };
+    # Override this input to change which plugins are built.
+    panda-config = {
+      url = "path:./panda/plugins/config.panda";
```
be32826: This will only work when running `nix build` in the cloned repo, but it won't work when used as a dependency. I get this error with `nix build github:lluchs/panda/flake`:

```
error: cannot fetch input 'path:./panda/plugins/config.panda' because it uses a relative path
```

Maybe plugin customization can be done at a different level from the flake inputs, such as with `.override`?

lluchs: Ah, annoying. Thanks for catching that. `.override` unfortunately doesn't easily work without making the whole thing a nixpkgs overlay.

I've seen people suggest an input "path:/dev/null" for optional external configuration. I will try that.

be32826: > `.override` unfortunately doesn't easily work without making the whole thing a nixpkgs overlay.

I think you might be able to just change the panda package definition from

```nix
panda = pkgs.stdenv.mkDerivation {
  ...
}
```

to something like

```nix
panda = pkgs.callPackage
  ({ plugins ? builtins.readFile ./panda/plugins/config.panda }:
    pkgs.stdenv.mkDerivation {
      ...
    }) {}
```

lluchs:
I implemented `file:/dev/null` as input now since it was simpler from what I already had. `nix build github:lluchs/panda/flake` works now, as well as local builds from the repository and overriding the configuration from another flake.

lluchs:
I implemented `file:/dev/null` as input now since it was simpler from what I already had. `nix build github:lluchs/panda/flake` works now, as well as local builds from the repository and overriding the configuration from another flake.

lluchs:
I updated the PR with a better approach now. The flake now exposes a function `pkgsWithConfig` that allows setting `configFile` (as before) and `targetList`, for example:

```nix
    pandaPkgs = panda.packages.${system}.pkgsWithConfig {
      configFile = ./config.panda;
      targetList = [ "x86_64-softmmu" ];
    };
```

I also updated the lockfile (again) for current nixpkgs and libosi.

PR #1558, opened 2024-12-06
[WIP] Use Capstone Debian package

AndrewQuijano:
Hello,
Based on work here, it seems like I got a fully functional version of libcapstone 5.0.4 working, I locally tested this, I can see libcapstone, I want to test this somehow.
https://github.com/capstone-engine/capstone/pull/2569

So two questions
1. I assume the integration test will make extensive use of callstack_instr which uses capstone right? I want to show the capstone team that if this passes the PANDA integration test, the Debian package works as expected

2. Once the confirmation is complete, I'll ask capstone to create a 5.0.4 release, with the Debian package, and I'd need to tweak this PR to match the capstone GitHub address.

Although, I think based on the work, it seems we should be very close to be able to put this on Debian apt repos directly too

P. S.
The Debian package already will run `ldconfig` for you.


AndrewQuijano:
Also, when updating the `cleanup` step, I noticed an error on copying over the libosi libraries, also had to update how libcapstone gets copied over.

image

AndrewQuijano:
Also, for libosi, wouldn't we want both the .so and .a files copied over?

PR #1559, opened 2024-12-06
[WIP] Allowing Testing on External Forks

AndrewQuijano:
Hello, (Update Feb 2, 2025)

So I have been testing running tests on external forks, and I think I got it working safely with this example
https://github.com/adwise-fiu/Level-Site-PPDT/pull/14

So for set up, I did the following
1- Setup branch protection rule
image

2- Update Action to allow external forks
image

3- See the updated workflow, so essentially what happens is, 
a) Someone creates a Pull Request
b) Before the option appears to even run the test cases, I need the changes approved, once I see that checkmark, then the option to run test cases appears, and I click OK and run the tests. 
image

The security comes from needing a manual approval. The GitHub Actions will prevent ANY code being run until Step A is completed.

There is one slight issue, the reviewer can't be the same person writing the PR, which is technically more secure, but does slightly make things a bit more annoying?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+<!-- This template is meant for GENERAL bug reports.
+
+Please be as descriptive as possible
+-->
+
+### Work environment
+
+<!-- Filling this table is mandatory -->
+
+| Questions                                | Answers
+|------------------------------------------|--------------------
+| System PANDA runs on OS/arch/bits        | panda-system-i386, panda-system-x86-64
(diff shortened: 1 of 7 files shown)


coolkingcole:
Will need to add protections provided by an environment that ensures approval from `panda-re` developers before running.

AndrewQuijano:
Btw @coolkingcole, since the test workflow stores containers in GHCR, but it seems this is linked to a user account, not sure how pull_request_target would work for that. At least secure use of pull_request_target means that you can securely share secrets.

You might need to switch to storing test containers in DockerHub too. Just something to note when testing this.
https://hub.docker.com/r/pandare/panda_test

AndrewQuijano:
Fixed by Luke

https://github.com/panda-re/panda/commit/3902702f0ccb9a17469ffc973e2f4e45db221103

PR #1560, opened 2024-12-07
Upgrade deploy

### LINKED ISSUE #1501: Version disconnects between deb and python packages

AndrewQuijano:
Hello,
This should be covering the last batch of chances to publishing GitHub action.

1. First, avoid publishing the same artifact twice, which caused an issue with the release that occurred last deployment
> Already fixed here: https://github.com/panda-re/panda/pull/1522

2. Changing the workflow to fix #1501 Essentially, having the first step be uploading to PyPi, this should enforce that the versions between releases and pip will be aligned permanently. If the PyPi upload fails, there will be no release created.

I should note, I believe you'd need to manually upload 1.8.28 and 1.8.35 to PyPi, not sure if you'd want to also delete 1.8.22.dev12 and 1.8.21.dev1 from PyPi, but this should hopefully make sure Pip is retroactively fixed to be in lock step after version 1.8.21

https://pypi.org/project/pandare/#history

3. Swap order of containers being built, so we can leverage caching for faster publishing of pandadev and panda containers.

4. I also removed duplicate code in regard to updating documentation used on both dev and stable branches.

To confirm this works, changing Python code causes a new commit created here in documentation step
https://github.com/panda-re/panda-re.github.io/commits/master/

Testing: 
I confirmed this approach of publish to pypi then release works, see code and action:

https://github.com/AndrewQuijano/Treespace_REU_2017/blob/main/.github/workflows/deploy_wheel.yml

https://github.com/AndrewQuijano/Treespace_REU_2017/actions/runs/14785728543/job/41513678740

As for the code de-duplication, I followed the steps done on the dev branch, the only difference instead of hard coding `dev` it points to `${{ github.ref_name }}`

### DIFF
--- a/.github/workflows/publish_docker.yml
+++ b/.github/workflows/publish_docker.yml
@@ -1,6 +1,7 @@
 name: Build and Publish Docker Container and Pypanda Docs # Only for main panda-re repo, not forks
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - dev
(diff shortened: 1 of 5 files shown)


AndrewQuijano:
Seems most of the work on the PR is already done in #1599

PR #1569, opened 2025-01-21
Support for Ubuntu 24.04 Noble

SneakyF0xy:
- Fix dependecies
- Fix network plugin for new wireshark headers
- Add building for LLVM [There is no apt version]
- Add building for libosi [No version for 24.04]
- Fixes small bugs in install script


AndrewQuijano:
Hello!
Good call out about Ubuntu 24, I just added this to my PR to update LibOSI CI/CD to also include building Ubuntu 24, I will work on adding this to LibOSI now

https://github.com/panda-re/libosi/pull/17
https://github.com/AndrewQuijano/libosi/releases/tag/v0.0.5

A few suggestions:

1- Rather than remove llvm-11 from dependencies, could you do a similar trick like on Ubuntu 18 on install script to have LLVM-11 for Ubuntu 24? Same with the docker image, let's try to avoid changing the installation files as much as possible

2- Seems container build and install_script failed
https://github.com/panda-re/panda/actions/runs/12891265305/job/35944130600?pr=1569

3- Not sure if you use libosi, but if you do, can you check the Debian package made works? Also undo the install libosi building step, since now I have a PR to get libosi to work with Ubuntu 24

4- I am going to get .eggs on gitignore file, so I think it is best to rebase your two commits to one. As the .eggs I think is generated when making PANDA

panda/scripts/build_libosi.sh
```diff
@@ -0,0 +1,18 @@
...
+else
+    echo "Directory 'libosi' already exists. Skipping clone."
+fi
+
+pushd libosi
+mkdir -p build && cd build
+cmake -GNinja ..
+ninja
+ninja package
+sudo dpkg -i libosi-*.deb
```
stevenjoezhang, line 16: I found a small issue here: the file name is actually like `libosi_*.deb` (changing `-` to `_`)

AndrewQuijano: yeah, just stick with `sudo dpkg -i libosi*.deb`, there was some mixup with how libosi releases were created, where on releases, it would be libosi_{ubuntu_version}.deb. I should have my PR to Libosi merged in shortly, although, I think getting help on creating the new Ubuntu 24 folder with build and runtime dependencies and proof that network plugin would work would be amazing to get!

https://github.com/panda-re/libosi/releases/tag/v0.1.9

AndrewQuijano:
@stevenjoezhang @SneakyF0xy I just updated LibOSI for it to have [Ubuntu 24 package](https://github.com/panda-re/libosi/releases/tag/v0.1.11)

@SneakyF0xy I have two requests:

1- Please rebase your Pull Request
2- Squash both commits, undo your changes to the installation script
3- Use this approach to install LLVM-11 for Ubuntu 20, 22, 24 in both install_ubuntu and in the Dockerfile
https://apt.llvm.org/
4- Please confirm that the network plugin works as intended for Ubuntu 24, ideally via screenshots or something.

panda/plugins/network/network.cpp
```diff
@@ -76,20 +77,19 @@ bool init_plugin(void *self) {
         fprintf(stderr, "Plugin 'network' needs argument: -panda-arg network:file=<file>\n");
         return false;
     }
-
+    int file_type = wtap_pcapng_file_type_subtype();
```
LauraLMann, line 80: I think you need to call this function AFTER calling wtap_init, or the file types will not yet be registered.  (At least that's what I'm seeing with my test case.)

AndrewQuijano:
Closing this as #1584 fixed everything listed here

PR #1571, opened 2025-01-30
mips64eb: don't assume size of endian flip

lacraig2:
When we fixup endianness in 64-bit big endian systems we assume that they are 64-bit.

This isn't the case for some targets. In particular, it's invalid for pid_t on 64-bit architectures.


lacraig2:
This does not quite work yet.

PR #1572, opened 2025-02-03
Prevent usage of uninitialized mem_buffer

emma592:
Check at line 186 (if (strlen(mem_buffer) > 0)) is not correct if the result of malloc is not zero, resulting in broken traces.


lacraig2:
Nice fix. Thanks.

PR #1580, opened 2025-04-29
Update callstack_instr to track binary names

MarkMankins:
Supercedes #1549 

Updates callstack_instr to track binary names along with program counters to make callstack dumps more useful.  Plugins that use callstack_instr to display callstacks need to opt-in to leverage the new functionality.  Enabling the new functionality does come with a performance hit.

Optimized the new API a bit and added some documentation to the README to illustrate how to use it.


LauraLMann:
Being closed as superseded by #1602.

PR #1587, opened 2025-08-04
Add process information option to tainted_branch plugin

LauraLMann:
Note that the tainted_branch plugin can still be used on guests that do not support OSI, as long as you don't use the new option to include the process information.


LauraLMann:
This PR has been rebased @lacraig2 and is ready for review again.  Thanks.

PR #1604, opened 2025-10-17
Create new Templates and DwarfDump documentation

AndrewQuijano:
I got these from the [Capstone project](https://github.com/capstone-engine/capstone/tree/next/.github). They are OK with me using it, and more importantly, I think we can leverage the labels to limit the scope of testing.

For good measure, I also noticed that panda.re has documentation, so I added some using AI for DwarfDump, which is used by LAVA.

Also, I think it's time to try reviving automatic documentation updates:
https://github.com/panda-re/panda-re.github.io

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,248 @@
+Build:
+  - Dockerfile
+  - ./panda/build.sh
+  - ./panda/build-llvm.sh
+  - ./panda/Makefile.panda
+  - ./panda/Makefile.panda.target
+  - Makefile
+  - Makefile.objs
+  - Makefile.target
+  - configure
+
+Documentation:
+  - '**/*.md'
+
+Github-files:
+  - .github/**
+
+PyPanda:
+  - ./panda/python/**/*.py
+  - ./panda/python/**/*.h
+
+Debian:
+  - ./panda/debian/**
+
+LLVM:
+  - ./panda/llvm/**
+  - ./panda/include/panda/tcg-llvm.h
+
+PANDA-Core:
+  - ./panda/src/**
+
+PANDA-Tests:
+  - ./panda/testing/**
+  - ./panda/plugins/taint2/tests/**
+
+# All PANDA plug-in tags, used to call specific plug-in tests
+asid_instr_count:
+  - ./panda/plugins/asid_instr_count/**
+
+asidstory:
+  - ./panda/plugins/asidstory/**
+
+callfunc:
+  - ./panda/plugins/callfunc/**
+
+callstack_instr:
+  - ./panda/plugins/callstack_instr/**
+
+callwitharg:
+  - ./panda/plugins/callwitharg/**
+
+checkpoint:
+  - ./panda/plugins/checkpoint/**
+
+collect_code:
+  - ./panda/plugins/collect_code/**
+
+correlatetaps:
+  - ./panda/plugins/correlatetaps/**
+
+cosi:
+  - ./panda/plugins/cosi/**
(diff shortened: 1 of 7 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,257 @@
...
+  - ./panda/plugins/cosi/**
+  - ./panda/python/core/extras/cosi.py
+
+cosi_strace:
+  - ./panda/plugins/cosi_strace/**
+
+coverage:
+  - ./panda/plugins/coverage/**
+
+cppskeleton:
```
LauraLMann: You don't need a label for this - it's just a template for writing a c++ plugin.  It's not even included in the config.panda file that says which plugins get built.

.github/labeler.yml
```diff
@@ -0,0 +1,257 @@
...
+  - ./panda/plugins/cppskeleton/**
+
+cskeleton:
```
LauraLMann: Ditto - the skeleton is not a real plugin.

.github/labeler.yml
```diff
@@ -0,0 +1,257 @@
...
+query_file_writes:
+  - ./panda/plugins/query_file_writes/**
+
+recctrl:
+  - ./panda/plugins/recctrl/**
+
+replaymovie:
+  - ./panda/plugins/replaymovie/**
+
+rust_skeleton:
```
LauraLMann: Although the rust skeleton is built, I do not understand why.  Not sure we really need a label for this - is it even sensible to have a test for this 'plugin'?

LauraLMann:
As for the questions asked via Slack:
If you have to have a label on the code for overall functionality, just use anything that says that to you like "other" or "general".  It would get applied to any code or build files, but not the documentation.  This means some code files would have multiple labels.  If possible, it might be easier to just assume every test labeled "other" or "general" (or whatever label you settle on) should be run for every non-documentation change.
the i386 and x86_64 architecture labels would be applied to the files in target/i386 and panda/disas/i386.c, and maybe some more.
the aarch64 and arm architecture labels would be applied to files in target/arm and panda/disas/arm-a64.cc and maybe some more.
the ppc architecture label would be applied to the files in target/ppc and panda/disas/ppc.c and maybe some more.
etc. for any architecture that PANDA builds (I think there are a bunch of architecture folders inherited from QEMU that PANDA doesn't actually support building).
You would need to search the code for references to the TARGET_<arch> constants to get an idea of what files and folders need to be labeled for that architecture.

LauraLMann (dismissed):

AndrewQuijano:
@LauraLMann The PR kinda got big, so I'll take care of tagging architectures in another PR, but which files or folders should I tag with general/other?

panda/python/core/pandare/extras/dwarfdump.py
```diff
@@ -514,7 +865,12 @@ def parse_dwarfdump(indat, prefix=""):
                 assert ('DW_AT_decl_file' in res)
                 decl_fn = res['DW_AT_decl_file']
                 decl_fn = decl_fn[v.decl_fn.find(' ')+1:]
-
+                # The check below may fail if v is not defined (first time through loop)
+                # It's better to use the current scope's file info if available, but
+                # given the original code's structure, we'll keep the assignment as is
+                # for minimal functional change, but note potential bug.
+                # TODO: Check if bottom line is correct in broader context.
+                # decl_fn = decl_fn[decl_fn.find(' ')+1:] # Safer version
```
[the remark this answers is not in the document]
AndrewQuijano: LAVA works as intended, I don't want to try anything just yet

PR #1607, opened 2025-11-10
Fix CI for multiple Debian Packages and Complete RR Snapshot

AndrewQuijano:
It appears that it can create both Debian packages and download them, but it can't find them for releases. Hopefully, updating the regex to recursively look around should fix this.

Also, I got LAVA to run locally to test that I can set the Complete RR variable to 1 via PyPanda. This is based on #1314


lacraig2:
In the future it'd be better to separate our these seemingly divergent issues into multiple PRs.

Issue #1605, opened 2025-10-17
Update Debian Package Install Path


AndrewQuijano:
Based on the documentation [here](https://www.debian.org/doc/debian-policy/ch-opersys.html), under 9.1.2, "As mandated by the FHS, packages must not place any files in /usr/local". I will write a PR to fix this.

PR #1617, opened 2025-11-30
Add PANDA header files to container/debian package

AndrewQuijano:
**Your checklist for this pull request**
- [X] I've documented or updated the documentation of every API function and struct in this PR.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

When installing PANDA, I also want to make sure the header files are included. This should make it easier for anyone who wants to write stuff in C++ with PANDA.

Additionally, regarding LAVA, I plan to relocate the hypercall struct to include it/ so PANDA can ensure that the struct is synchronized between LAVA and PANDA.
...

**Test plan**

Here is a screenshot confirming the Debian package now has the header files:
image

...

**Closing issues**

N/A
...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -180,6 +180,7 @@ COPY --from=base /tmp/build_dep.txt /tmp
 
 # Set up /package-root with files from panda we'll package
 COPY --from=p /usr/local/bin/panda* /usr/local/bin/libpanda* /usr/local/bin/qemu-img /package-root/usr/local/bin/
+COPY --from=p /usr/local/include/ /package-root/usr/include/
 COPY --from=p /usr/local/etc/panda /package-root/usr/local/etc/panda
 COPY --from=p /usr/local/lib/panda /package-root/usr/local/lib/panda
 COPY --from=p /usr/local/share/panda /package-root/usr/local/share/panda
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -180,6 +180,7 @@ COPY --from=base /tmp/build_dep.txt /tmp
 
 # Set up /package-root with files from panda we'll package
 COPY --from=p /usr/local/bin/panda* /usr/local/bin/libpanda* /usr/local/bin/qemu-img /package-root/usr/local/bin/
+COPY --from=p /usr/local/include/ /package-root/usr/include/
```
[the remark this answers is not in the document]
AndrewQuijano: This is intentional as this is a Debian package, so we should avoid installing to /usr/local, see #1605

lacraig2, line 183: How big is /usr/local/include? Does this add a ton of stuff?

Can we just add the things we care about here.

lacraig2 (dismissed):

Dockerfile
AndrewQuijano: It will only add the content of this folder

https://github.com/panda-re/panda/tree/dev/panda/include

This is about 300KB, which shouldn't make the package much larger, so you can now develop plugins in C++ with the Debian package.

image

lacraig2:
:+1: Rebase and I'll merge.

